PR #6, opened 2020-04-30
Add an rpm spec file

Gundersanne:
Includes a dockerfile and a `make container` target which installs the rpms in a f31 container.


distribution/osbuild-installer.socket
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=OSBuild Composer RCM API socket
```
teg, line 2: Copy-paste error ;)

osbuild-installer.spec
```diff
@@ -0,0 +1,71 @@
...
+# Upstream license specification: Apache-2.0
+License:	ASL 2.0
+URL:		%{gourl}
+Source0:        %{gosource}
+
+BuildRequires:	%{?go_compiler:compiler(go-compiler)}%{!?go_compiler:golang}
+BuildRequires:	systemd
+
+Requires: systemd
+Provides: osbuild-installer
```
teg, line 24: This is not needed. A package always provides itself.

Makefile
```diff
@@ -1,4 +1,4 @@
-PACKAGE_NAME = osbuild-installer-backend
```
teg: I like this change! Would it makes sense to rename the whole project to `osbuild-installer` and drop the `-backend` suffix?

Gundersanne: it does sound a bit silly, I copied it from the cloud rh projects but yea I'm in favour of dropping it :)

Gundersanne: Hm I don't have the rights to change the name of this project. Could you rename it to drop the `-backend` ?

PR #7, opened 2020-07-22
Buildimage endpoint

Gundersanne:
- [x] https://github.com/osbuild/osbuild-installer/pull/8


Gundersanne:
@teg yea I agree, currently the repos are passed all the way from the webpage. It would totally make sense if osbuild-installer would have a list of distributions/archs/repos it supports, instead of osbuilt-installer-frontend having that list.

+ the idea behind this api is also that we could bolt anything higher levely onto it.

PR #11, opened 2020-09-10
👂 Set a custom listen address if provided

major:
When running in OpenShift, installer needs to listen on a different
address (and possibly different port, too). Allow a user to specify a
custom listen address via environment variable.


larskarlitski (changes requested):
Agreed that it should not be hard-coded, but this should be in the [socket unit](https://github.com/osbuild/osbuild-installer/blob/master/distribution/osbuild-installer.socket) for the service.

major:
@larskarlitski I cannot put this into OpenShift if we want to use the same socket method as osbuild-composer. I need the daemon to listen on a port directly.

PR #15, opened 2020-09-11
👂 Set a custom listen address if provided

major:
When running in OpenShift, installer needs to listen on a different
address (and possibly different port, too). Allow a user to specify a
custom listen address via environment variable.

### DIFF
--- a/cmd/osbuild-installer/main.go
+++ b/cmd/osbuild-installer/main.go
@@ -1,9 +1,20 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


major:
This is a re-do of https://github.com/osbuild/image-builder/pull/11 which somehow broke with the project rename. :shrug:

Gundersanne (approved):
I find this ok. The truth is we need an executable as an entrypoint for the container.

PR #17, opened 2020-09-17
Tests

Gundersanne:
This just tests the parts that doesn't actually require osbuild-composer running, it's easiest if we wait with that until it's released.


larskarlitski (approved):
Thanks!

PR #19, opened 2020-09-23
🚄 Speed up container build

major:
Use the go-toolset ubi8 container to speed up the container build
process since it comes with golang right out of the box.

### DIFF
--- a/distribution/Dockerfile-ubi
+++ b/distribution/Dockerfile-ubi
@@ -1,15 +1,14 @@
 # Use a builder container to build the Go application (which we extract in
 # the second container).
-FROM registry.access.redhat.com/ubi8/ubi:latest AS builder
-RUN dnf -y install /usr/bin/go
-WORKDIR $GOPATH/src/github.com/osbuild/image-builder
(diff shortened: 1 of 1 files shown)


major:
Looks like about 40s including pulls in GitHub Actions. Not bad.

teg:
What's the difference between the golang container and the UBI? Is it just that it already contains the compiler, or does it have a different base too?

larskarlitski:
Is the speed worth it? Go 1.13 on debian might be differently configured from the one in RHEL.

larskarlitski:
> Debian?

:stuck_out_tongue: 

```
$ podman run -it --rm docker.io/library/golang:1.13 
root@24a9d897d838:/go# cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
[...]
```

major:
@teg @larskarlitski I discovered the `ubi8/go-toolset` container and used it instead. Just updated the PR.

major:
Still about 40-45 seconds to build including the pulls.

Gundersanne (approved):
lgtm :)

PR #20, opened 2020-09-30
server: Switch to echo

Gundersanne:


internal/server/server.go
```diff
@@ -13,161 +13,116 @@ import (
...
 
 type Handlers struct{}
 
-func (s *Handlers) GetVersion(w http.ResponseWriter, r *http.Request) {
+func (s *Handlers) GetVersion(ctx echo.Context) error {
 	spec, err := GetSwagger()
 	if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
+		return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
```
larskarlitski, line 24: Just returning `err` is enough here. Echo turns non`HTTPError`-errors into internal server errors automatically.

internal/server/server.go
```diff
@@ -13,161 +13,116 @@ import (
...
 	}
 
-	decoder := json.NewDecoder(r.Body)
+	decoder := json.NewDecoder(ctx.Request().Body)
 	var composeRequest cloudapi.ComposeJSONRequestBody
 	err := decoder.Decode(&composeRequest)
 	if err != nil {
-		http.Error(w, fmt.Sprintf("Failed decoding the compose request %s", err.Error()), http.StatusBadRequest)
-		return
+		return echo.NewHTTPError(http.StatusBadRequest, "Failed decoding the compose request %s", err.Error())
```
larskarlitski, line 85: `NewHTTPError()` does not accept format strings. (We are all making this mistake independently.)

Please use `fmt.Sprintf()`. Also, when using the `%v` directive, you don't need to call the `Error()` method:
```suggestion
		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Failed decoding the compose request %v", err))
```

larskarlitski (changes requested):
Thank you! Two small comments, which might need fixing in a few more places than the ones I pointed them to.

Otherwise this is perfect.

internal/server/server.go
Gundersanne: neat!

internal/server/server.go
Gundersanne: I've been using "" + err.Error(); "%v", error, "%s", err.Error(); fmt() or not intermittently, good to finally know how to do it :D

larskarlitski (approved):
Thanks for the quick turnaround!

PR #28, opened 2020-10-12
🗞 openshift: Use modern v1 API for deployments

major:
apps/v1 + Deployment is the old style.
v1 + DeploymentConfig is the new style.


larskarlitski (approved):
This seems correct to me, from looking at https://docs.openshift.com/container-platform/3.11/dev_guide/deployments/how_deployments_work.html

Thanks!

PR #29, opened 2020-10-12
Explicitly set listening port without variables

major:
Apparently dealing with integers in OpenShift template YAML is a pain.

### DIFF
--- a/templates/image-builder.yml
+++ b/templates/image-builder.yml
@@ -43,7 +43,7 @@ objects:
         - image: "${IMAGE_NAME}:${IMAGE_TAG}"
           name: image-builder
           ports:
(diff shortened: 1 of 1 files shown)


templates/image-builder.yml
```diff
@@ -43,7 +43,7 @@ objects:
         - image: "${IMAGE_NAME}:${IMAGE_TAG}"
           name: image-builder
           ports:
-          - containerPort: "${BACKEND_LISTENER_PORT}"
+          - containerPort: 8086
```
Gundersanne, line 46: Wait, without quotes?

major: @Gundersanne indeed. Couldn't get it to work and the docs said to do this. 🤷🏻‍♂️

PR #30, opened 2020-10-13
🔨 openshift: Fix deployment template

major:
Using github-mirror[0] as a guide:

  * Use double curly braces for integer values (will be left unquoted)
  * Change `DeploymentConfig` to `Deployment`
  * Change apiVersion for `Deployment` to `apps/v1`

[0] https://github.com/app-sre/github-mirror/blob/master/openshift/github-mirror.yaml


Gundersanne (approved):
I have a whole bag of these green checks!

PR #40, opened 2020-10-19
Drop fedora container and add distributions folder to the ubi container

Gundersanne:


major (approved):
LGTM. Sorry for the delay.

PR #41, opened 2020-10-19
Logging

Gundersanne:


larskarlitski (commented):
Wow, this seems like quite the ordeal to get logging going. It looks good to me, but why does the service need to care about where to send the logs? This seems like something that could be solved generically once, and hooked to the output of all services?

internal/server/logger.go
```diff
@@ -0,0 +1,123 @@
+package server
```
teg, line 1: Maybe make this a new package, with some high-level documentation what it does and what it expects from its environment, ideally with references to documentation? Is this specific to CloudWatch? cloud.redhat.com? OpenShift?

Makefile
```diff
@@ -36,11 +36,6 @@ rpm: $(RPM_SPECFILE) $(RPM_TARBALL)
 		--with tests \
 		$(RPM_SPECFILE)
 
-.PHONY: container
```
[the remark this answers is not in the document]
Gundersanne: Oof sorry, this is part of #40 , I need to rebase this. Adding a label.

internal/server/logger.go
```diff
@@ -0,0 +1,123 @@
...
+	b := &bytes.Buffer{}
+
+	now := time.Now()
+
+	hostname, err := os.Hostname()
+	if err == nil {
+		f.Hostname = hostname
+	}
+
+	data := map[string]interface{}{
```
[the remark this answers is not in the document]
Gundersanne: So most of this is shamelessly copied from https://github.com/RedHatInsights/insights-ingress-go/blob/master/logger/logger.go ; I think the format comes from the python example provided by insights which uses https://pypi.org/project/logstash_formatter/

internal/server/logger.go
```diff
@@ -0,0 +1,123 @@
...
+type Formatter struct {
+	Hostname string
+}
+
+// NewCloudwatchFormatter creates a new log formatter
+func NewCloudwatchFormatter() *Formatter {
+	f := &Formatter{}
+
+	var err error
+	if f.Hostname == "" {
```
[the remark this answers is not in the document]
Gundersanne: good point, I'll remove that

internal/server/logger.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	b.Write(j)
+	b.WriteRune('\n')
+
+	return b.Bytes(), nil
+}
+
+func ReadNameSpace() string {
+	namespace, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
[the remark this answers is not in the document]
Gundersanne: yes, and yes

Gundersanne:
> Wow, this seems like quite the ordeal to get logging going. It looks good to me, but why does the service need to care about where to send the logs? This seems like something that could be solved generically once, and hooked to the output of all services?

So its not the case that the stdout of all containers are sent to some central location. They use AWS cloudwatch for logging, and services have to send their logs there in a specific format. They have applications built on top of those logs which allow for filtering/analyzing etc...

internal/server/logger.go
Gundersanne: Do you want this documentation inline or rather in the README, not sure what conventions are in osbuild

larskarlitski:
> So its not the case that the stdout of all containers are sent to some central location. They use AWS cloudwatch for logging, and services have to send their logs there in a specific format. They have applications built on top of those logs which allow for filtering/analyzing etc...

Thanks for the explanation. Seems a bit weird, but I guess it is what it is 😄

internal/server/logger.go
teg: Both? ;) I mean, technical details in the package and enough info to set things up in the README? Ideally the code should be explained in isolation though (in case it is used in other places or implements features we don't yet/stop using).

internal/server/logger.go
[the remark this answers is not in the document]
teg: Maybe just add a link?

Gundersanne:
@larskarlitski @teg I moved configuration out to `cmd/image-builder/config.go`, and restructured the server a bit so it now contains an instance of the logger, and the logger is no longer a global thing. Also the structure is now a bit more similar to osbuild-composer.

cmd/image-builder/main.go
```diff
@@ -1,20 +1,63 @@
...
+			switch kind {
+			case reflect.Ptr:
+				fieldV.Set(reflect.ValueOf(&confV))
+			case reflect.String:
+				fieldV.SetString(confV)
+			default:
+				return fmt.Errorf("Unsupported type")
+			}
+		} else if kind == reflect.String {
+			return fmt.Errorf("Undefined non-pointer field without default value %v", fieldT)
```
Gundersanne, line 43: @larskarlitski what do you think enforcing this? Should I just drop the *string vs string altogether?

teg (approved):
This looks great!

cmd/image-builder/main.go
larskarlitski: Yeah, good point. I'd drop it and return an error when anything that is not a string is passed.

In fact, I think I'd drop the whole `default` tag. I don't think it adds anything over just setting the default values before passing it to your function:

```go
config = Config{
        ListenAddress: "localhost:8086",
}

err := LoadConfigFromEnv(&config)
```

Again, this is similar to how package `json` handles defaults (i.e., not).

larskarlitski: This also makes defaults more explicit – they seem quite hidden in the tags...

larskarlitski (commented):
I like it!

Replied to your comment about string pointers. I think defaults might not belong in the tags.

Otherwise, the commit message is a bit sparse and `LoadConfigFromEnv` could use a test.

I don't mind merging it as is and addressing the above in a follow up. Thanks! ❤️

PR #42, opened 2020-10-22
Error handling

Gundersanne:
- [x] blocked on #41 
- [x] needs adjustments in frontend (`composes/$id` rather than `compose/$id`)

---

This contains mostly little adjustments to be compliant with clouddot guidelines.

They also prefer using the content-Type `application/vnd.api+json` however echo doesn't really support this, and it would require overwriting headers, or encoding the response into json ourselves etc... so I'd just leave it unless it turns out to be a real problem.

### DIFF
--- a/internal/server/api.yaml
+++ b/internal/server/api.yaml
@@ -1,7 +1,7 @@
 ---
 openapi: 3.0.1
 info:
-  version: "1"
+  version: "1.0"
   title: Image-builder service
   description: Service that relays image build requests
   license:
@@ -59,7 +59,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Architectures'
-  /compose/{composeId}:
+  /composes/{composeId}:
     get:
       summary: get status of an image compose
       parameters:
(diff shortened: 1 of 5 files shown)


teg (dismissed):
Makes sense to me. A reference to the guidelines would be nice (if it is public).

internal/server/api.yaml
```diff
@@ -97,9 +97,31 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ComposeResponse'
-
+        '400':
+          description: the compose request is malformed
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPErrorList'
```
larskarlitski, line 105: Why is this a list of errors? Seems like you only ever put one into it, no?

Gundersanne: Yup, but it has to comply with https://jsonapi.org/format/#errors:

> Error objects MUST be returned as an array keyed by errors in the top level of a JSON:API document.

Gundersanne: Maybe I should rename the HTTPErrorList to just HTTPErrors? because it's an object container 'errors: []'

larskarlitski: Ah, interesting. Thanks for the link.

Both `ErrorList` and `Errors` work for me. This name is not exposed as API.

larskarlitski (approved):

PR #43, opened 2020-10-27
Client ssl osbuild

Gundersanne:


internal/server/api.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 openapi: 3.0.1
 info:
-  version: "1"
+  version: "1.0"
```
teg, line 4: Makes sense. Are the guidelines public? Could you link to them?

internal/server/server.go
```diff
@@ -21,6 +23,60 @@ import (
 
 type Handlers struct{}
 
+func OsbuildClient() (*cloudapi.Client, error) {
+	socket, ok := os.LookupEnv("OSBUILD_SERVICE")
```
teg, line 27: I think we can rely on this being set, and it being `https`, no need to support anything else, I think (not sure composer even does).

internal/server/server.go
```diff
@@ -21,6 +23,60 @@ import (
...
+			if err != nil {
+				return err
+			}
+			caCertPool := x509.NewCertPool()
+			caCertPool.AppendCertsFromPEM(caCert)
+			tlsConfig = &tls.Config{
+				Certificates: []tls.Certificate{cert},
+				RootCAs:      caCertPool,
+			}
+		} else {
```
teg, line 66: Don't think we need to / should support the case where the CA is not provided. @larskarlitski?

internal/server/server.go
```diff
@@ -21,6 +23,60 @@ import (
...
+		socket = "http://127.0.0.1:80"
+	}
+	Log.Infoln("Creating osbuildclient with socket %s", socket)
+	return cloudapi.NewClient(socket, ConfigureClient)
+}
+
+func ConfigureClient(client *cloudapi.Client) error {
+	// set up client certificate authentication
+	if (strings.HasPrefix(client.Server, "https")) {
+		certFile, ok := os.LookupEnv("OSBUILD_CERT_PATH")
```
teg, line 38: This is a general comment, and I don't really know enough about the environment we expect, so take it with a grain of salt: in composer we try to shy away from environment variables and rather use config files. Is that something we could/should do here too, or is the env vars more standard / makes stuff easier in some way?

Either way, I think it is better if we extract the config from env vars or from config files in `main.go` and pass it down from there, rather than doing it somewhere deep down in `internal/` (admittedly not so deep at the moment, but once things grow it might get unwieldy).

teg (changes requested):
Looking good, only minor nitpicks.

internal/server/server.go
Gundersanne: Since our main target is an openshift environment it's pretty flexible. Most services I've used as a reference use environment variables but there's nothing stopping us from using a file, as long as the variables are visible in openshift without having to open a shell, which we can easily do using a configmap which creates a file.

internal/server/server.go
Gundersanne: Isn't this useful for local development?

internal/server/server.go
Gundersanne: Ok I tried addressing this in #41:

https://github.com/osbuild/image-builder/blob/e041ef90579f1d29cf48a371f0bc1e881d44e809/cmd/image-builder/config.go

I think using the environment will still be easiest for the openshift environment, but if we want to ever fill the config in from a file that won't be too hard then.

teg: Ah, just read about config maps and secrets, very cool. I think having config as files would be best as it is easier to extend, and env vars are just iffy in general. I wouldn't mind keeping secrets as vars as their structure is much more limited, and i think it is more standard? i think they could be made into files too though, if you wanted to

Gundersanne: Hmm I'd make a config.go and a secrets.go then i guess, the latter reading from the environment (which will be mapped by openshift from Vault), and then the former reading from a file.

teg: After our call (and also reading all your comments, I somehow only saw the first), I take back my suggestion regarding config files. Let's change that if and when it becomes relevant and keep it simple for now.

internal/server/server.go
teg: In composer we just had a make target to generate certs, but this is really a nitpick and something we could change in the future, so if you are using it like this for development, then keep it as is for now.

internal/server/server.go
Gundersanne: i made it not mandatory to be https for now, that way I can figure out how to put all the certs/secrets in place on clouddot stage without fiddling on this PR.

teg (approved):
very nice, thanks!

schutzbot/build.sh
```diff
@@ -54,3 +54,9 @@ sudo mock -v -r $MOCK_CONFIG --resultdir $REPO_DIR --with=tests \
     rpmbuild/SRPMS/*.src.rpm
 
 sudo dnf localinstall -y $REPO_DIR/*x86_64.rpm
+
+# Build the container
+sudo dnf install -y podman
+sudo setenforce 0 # todo
```
Gundersanne, line 60: This was for building the container on rhel-8, which is disabled for now.

schutzbot/Jenkinsfile
```diff
@@ -33,29 +33,32 @@ pipeline {
...
-                        sh "schutzbot/ci_details.sh"
-                        retry(3) {
-                            sh "schutzbot/mockbuild.sh"
+                            sh "schutzbot/build.sh"
                         }
+                        sh "schutzbot/provision-composer.sh"
                         sh "schutzbot/run_tests.sh"
                     }
                 }
+                // TODO osbuild-composer-api isn't available in rhel8
```
Gundersanne, line 47: Any way to get the latest composer in rhel8?

PR #46, opened 2020-11-03
image-builder: Specify config default values when creating it

Gundersanne:
Instead of defining them in tags, define them when creating the object.


larskarlitski (approved):
Very lovely. Thank you!

PR #47, opened 2020-11-03
templates: Add database secrets to container environment

Gundersanne:
I want to see if the changes in app-interface worked, i somehow can't see the secrets directly, so I'm not actually sure if they're in place.

### DIFF
--- a/templates/image-builder.yml
+++ b/templates/image-builder.yml
@@ -77,11 +77,31 @@ objects:
               value: "${CW_LOG_GROUP}"
             - name: CW_AWS_REGION
               value: "${CW_AWS_REGION}"
-            - name: EXAMPLE
+            - name: PGHOST
               valueFrom:
-                configMapKeyRef:
-                  key: example
-                  name: image-builder-config
+                secretKeyRef:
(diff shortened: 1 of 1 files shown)


templates/image-builder.yml
```diff
@@ -77,11 +77,31 @@ objects:
...
               value: "${CW_AWS_REGION}"
-            - name: EXAMPLE
+            - name: PGHOST
               valueFrom:
-                configMapKeyRef:
-                  key: example
-                  name: image-builder-config
+                secretKeyRef:
+                  name: db
+                  key: host
```
major, line 84: So I think the `name` should be `image-builder-db` and the key should be `db.host`, `db.port`, etc.

The name of the secret itself is `image-builder-db`. See https://gitlab.cee.redhat.com/service/app-interface#manage-rds-databases-via-app-interface-openshiftnamespace-1yml

Gundersanne: Thanks! this makes more sense to me know.

major (approved):
Thanks for doing this. Let's give it a try.

PR #49, opened 2020-11-03
Add cloudwatch configuration and startup script

major:
Add the AWS Cloudwatch configuration and credentials to the deployment for logging. Also, use a startup script for image-builder that prints some diagnostic information about its database connection and where to find logs in AWS.


Gundersanne (approved):
Neat!

PR #54, opened 2020-11-04
🔍 Run gosec static analyzer

major:

### DIFF
--- a/.github/workflows/gosec.yml
+++ b/.github/workflows/gosec.yml
@@ -1,0 +1,28 @@
+name: gosec
+
+on:
+  pull_request:
+    branches:
+      - "*"
+  push:
+    branches:
+      - master
+  schedule:
+    - cron: '0 5 * * *' # Daily at 05:00 UTC
+
+jobs:
+  coverity:
+    name: "🔎 gosec check"
+    if: github.repository == 'osbuild/image-builder'
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+    steps:
+
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Run Gosec Security Scanner
+        uses: securego/gosec@v2.5.0
+        with:
+          args: ./...
--- a/internal/server/distribution.go
+++ b/internal/server/distribution.go
@@ -37,8 +37,8 @@ func ReadDistributions(distro string) ([]Distribution, error) {
 				return nil
 			}
 
-			f, err := os.Open(path)
-			defer f.Close()
+			f, err := os.Open(path) // #nosec G304
+			defer f.Close() // #nosec G307
 			if err != nil {
 				return err
 			}
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -48,7 +48,10 @@ func NewServer(logger *logrus.Logger) *Server {
 
 func (s *Server) Run(address string) {
 	s.logger.Infoln("🚀 Starting image-builder server on %s ...\n", address)
-	s.echo.Start(address)
+	err := s.echo.Start(address)
+	if err != nil {
+		s.logger.Errorln(fmt.Sprintf("Error starting echo server: %v", err))
+	}
 }
 
 func (h *Handlers) GetVersion(ctx echo.Context) error {


.github/workflows/gosec.yml
```diff
@@ -0,0 +1,24 @@
+name: gosec
+
+on:
+  pull_request: {}
+  push: {}
+  schedule:
+    - cron: '0 5 * * *' # Daily at 05:00 UTC
```
larskarlitski, line 11: Hm, why do we need to run this on a schedule in addition to when the code changes?

.github/workflows/gosec.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+    steps:
+
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Run Gosec Security Scanner
+        uses: securego/gosec@master
```
larskarlitski, line 26: I'd rather pin this to a version and update it separately. This way, we don't see unrelated PRs failing when `gosec` changes or becomes more strict.

internal/server/server.go
```diff
@@ -48,7 +48,10 @@ func NewServer(logger *logrus.Logger) *Server {
 
 func (s *Server) Run(address string) {
 	s.logger.Infoln("🚀 Starting image-builder server on %s ...\n", address)
-	s.echo.Start(address)
+	err := s.echo.Start(address)
+	if err != nil {
+		s.logger.Errorln(err)
```
larskarlitski, line 53: Please add some context to the error. Something like `"Error starting echo server: %v", err`

internal/server/distribution.go
```diff
@@ -37,8 +37,8 @@ func ReadDistributions(distro string) ([]Distribution, error) {
 				return nil
 			}
 
-			f, err := os.Open(path)
-			defer f.Close()
+			f, err := os.Open(path) // #nosec
+			defer f.Close() // #nosec
```
larskarlitski, line 41: Why does gosec need to be disabled here?

larskarlitski (changes requested):
Thanks for this!

Do you mind adding a quick explanation what gosec is and what kinds of errors it catches in the commit message introducing it?

.github/workflows/gosec.yml
major: @gicmo told me to

internal/server/distribution.go
major: It gets upset about the deferred close, but most people seem to disable the gosec checks here. I am not good enough with golang to know if there's a better pattern for this.

Do you have any ideas?

internal/server/server.go
major: Fixing

internal/server/distribution.go
larskarlitski: At the least, I'd say we need a comment stating which check is hidden here. Do you know if it is "G307: Deferring a method which returns an error"?

At least that looks like the most plausible to me. If it is, I think it's fine to ignore the error returned by `Close`, because we're catching any read errors from the file in `json.Decoder.Decode()`. Deferring in a loop is not the best idea anyway, though -- some linters complain about that already.

.github/workflows/gosec.yml
gicmo: Coverity has a limit of three scans per day, we can't run it on every code change sadly :/

internal/server/distribution.go
major: That's the one. It's `G307`.

.github/workflows/gosec.yml
larskarlitski: This is unrelated to coverty. But @gicmo just explained out of band that having this on a schedule can makes sense, because of the usual PR race. Let's leave it. Thanks!

internal/server/distribution.go
larskarlitski: Good to know.

1. Do you mind adding a comment/link mentioning that – so that the person touching this code next doesn't need to guess. Something like `// Ignore gosec's "G307: Deferring a method which returns an error"` should be enough.

2. Please remove it on the `Open` call. We do check the error there.

major: @larskarlitski I added some comments in there and re-pushed.

larskarlitski (approved):
Thanks for addressing my concerns, @major!

I filed https://github.com/osbuild/image-builder/issues/55 to remind us to follow-up on the G304 issue.

PR #56, opened 2020-11-05
db: Add database

Gundersanne:
- [ ] add test


cmd/image-builder/config_test.go
```diff
@@ -31,7 +31,7 @@ func TestEnv(t *testing.T) {
 
 	config := ImageBuilderConfig{
 		ListenAddress: "localhost",
-		LogLevel: "DEBUG",
+		LogLevel:      "DEBUG",
```
Gundersanne, line 34: this formatting and the similar ones are due to not running `tools/prepare-source.sh` in #46

internal/db/migrations/1_create_table_images.up.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TABLE IF NOT EXISTS Images(
+       Job UUID PRIMARY KEY
```
Gundersanne, line 2: obviously incomplete

internal/db/db.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+
+	"github.com/jackc/pgtype"
+	pgtypeuuid "github.com/jackc/pgtype/ext/gofrs-uuid"
+	"github.com/jackc/pgx/v4"
+	"github.com/jackc/pgx/v4/pgxpool"
+)
+
+type DB struct {
+	Pool *pgxpool.Pool
```
Gundersanne, line 14: maybe add a reference to the logger here too so we can use it in the db package?

.github/workflows/container.yml
```diff
@@ -16,20 +16,22 @@ jobs:
...
-      - name: Verify that the startup log line is in the logs
-        run: |
-          docker logs image-builder | grep "Starting image-builder server"
-
-      - name: Check that the API is responding
-        run: curl -sv http://127.0.0.1:8086/api/image-builder/v1/version
-
-      - name: Stop the container
-        run: docker stop --time 1 image-builder
+      # - name: Run the container
```
Gundersanne, line 19: We don't have docker here, and image-builder needs a psql instance. So i'm unsure what to do here.

Maybe we need to change it in such a way that image-builder fails to start when there's no DB there, but just marks itself as 'not ready'. We need to add a readiness check/endpoint anyway and this might be a good use case. We could start imagebuilder, ask for the version, and make sure it's not ready?

Makefile
```diff
@@ -3,6 +3,7 @@ PACKAGE_NAME = image-builder
 .PHONY: build
 build:
 	go build -o image-builder ./cmd/image-builder/
+	go build -o image-builder-migrate-db ./cmd/image-builder-migrate-db/
```
Gundersanne, line 6: We'd need to run this whenever there's a new version, but we don't want to run this within `image-builder` as then multiple instances would migrate the database all at once. Which I'm sure wouldn't break, or at least it shouldn't, but it's better to run it as a job, once, with every (re)deployment.

internal/config/config_test.go
```diff
@@ -12,7 +12,7 @@ func TestDefault(t *testing.T) {
 
 	config := ImageBuilderConfig{
 		ListenAddress: "localhost",
-		LogLevel: "DEBUG",
+		LogLevel:      "DEBUG",
```
Gundersanne, line 15: Formatting from prepare-source, so just ignore this diff.

Gundersanne:
@teg Would you mind having an initial look at this? I wrote down some questions/thoughts too and I'd be interested in hearing yours.

.github/workflows/container.yml
teg: Doesn't github have service containers just for this kind of test?

Gundersanne: I did not know about this! i'll look into it

ondrejbudai: This is what might interest you: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservices

Also, I don't know any details but I used `docker` and `podman` without any issues in GH Actions.

templates/image-builder.yml
```diff
@@ -34,6 +34,64 @@ objects:
     selector:
       matchLabels:
         name: image-builder
+    strategy:
+      type: Rolling
+      rollingParams:
+        pre:
```
Gundersanne, line 40: @major so k8s' Deployment doesn't have the pre-start lifecycle hook available under strategy, only ocp's DeploymentConfig does. Would you have any objections to switch to a DeploymentConfig?

templates/image-builder.yml
```diff
@@ -34,6 +34,64 @@ objects:
...
+          failurePolicy: Abort
+          execNewPod:
+            containerName: image-builder
+            command: [ "/app/image-builder-migrate-db" ]
+            env:
```
Gundersanne, line 45: I am not sure if execNewPod needs the env specified a second time. This instance of the container definitely needs the additional `MIGRATIONS_DIR` defined, but other than that it shares most variables with the main container.

PR #59, opened 2020-11-26
server: Only require Region and ShareWithAccounts for aws upload

Gundersanne:
Secrets should be accessible by image-builder, and added to the request
for composer. The user should only have to give their aws account id to
share the image with.

Region for now should be filled by image-builder as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,5 +42,5 @@ ubi-container:
 
 .PHONY: update-cloudapi
 update-cloudapi:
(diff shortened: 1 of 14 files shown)


teg (dismissed):
Nice!  Only one question: How come region is still required if it is filled in by image-builder?

teg:
(My question can be addressed in a follow-up, feel free to merge).

Gundersanne:
@teg i just removed it now :)

larskarlitski (approved):
Looks good to me. Thank you!

teg (commented):
🥳

PR #61, opened 2020-12-01
tests: Enable rpmlint

atodorov:


atodorov:
I believe I've fixed all rpmlint issues and then rebased to master to pick-up a fix for a failing test but then schutzbot failed with something else.

Gundersanne (approved):

PR #64, opened 2020-12-01
tests: Enable shellcheck

atodorov:
this yields lots of issues and I'm not sure if I should follow all of the suggestions to resolve them or not.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,7 +6,7 @@ on:
       - "*"
   push:
     branches:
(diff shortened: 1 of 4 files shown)


Gundersanne (approved):
thanks! 

edit: lets' wait for schutzbot though, as schutzbot uses these scripts :)

atodorov:
> lets' wait for schutzbot though, as schutzbot uses these scripts :)

That's what I was waiting for. All is green so merging.

PR #67, opened 2020-12-04
Add tests for ConfigureClient

atodorov:


internal/cloudapi/client_test.go
```diff
@@ -33,3 +33,15 @@ func TestOsbuildClientGetNoError(t *testing.T) {
 	require.NotNil(t, new_client)
 	require.Equal(t, "localhost:8086/", new_client.Server)
 }
+
+func TestOsbuildClientConfigureClientReturnsNilWhenNotUsingHttps(t *testing.T) {
+	// we're not using HTTPS URL on purpose
+	osbuild_client := NewOsbuildClient("localhost:8086/", nil, nil, nil)
```
Gundersanne, line 39: hmm tbh for this I'd generate some certificates, put them in a temporary directory. And create the client with those. Otherwise this test kinda doesn't do anything new compared to the ones above (as ConfigureClient does nothing without certs).

atodorov: this one checks the result of ConfigureClient explicitly which indeed isn't much, but that result is also used in NewClient itself. While the above test doesn't check anything related to the execution of ConfigureClient.

My next scenario would be with certificates.

internal/cloudapi/client_test.go
```diff
@@ -33,3 +35,31 @@ func TestOsbuildClientGetNoError(t *testing.T) {
...
+	osbuild_client := NewOsbuildClient("https://localhost:8086/", &myCert, &myKey, &myCA)
+
+	openapi_client := Client{
+		Server: osbuild_client.osbuildURL,
+	}
+
+	result := osbuild_client.ConfigureClient(&openapi_client)
+	require.Nil(t, result)
+	require.IsType(t, &http.Client{}, openapi_client.Client)
+	// require.IsType(t, &tls.Config{}, openapi_client.Client.Transport.TLSClientConfig)
```
atodorov, line 64: @Gundersanne this one line is giving me grief. I do not understand why. See:
https://github.com/osbuild/image-builder/runs/1499497778?check_suite_focus=true

Gundersanne: Hm the way this is supposed to be used is `osbuildClient.Get()`. This will create a client and configure it.

```
osbuildClient := NewOsbuildClient....

cloudapiClient, err := osbuildClient.Get()
require.NoError(t, err)
```

After that, at most I'd check if `cloudapiClient` has the `Compose` and `ComposeStatus` functions, as those are used by the server to contact osbuild-composer. Potentially using `require.Implements` but i'm not sure there.

atodorov: Note: the 2 scenarios introduced by this PR are meant to exercise and assert the results of calling ConfigureClient directly.  This is different (although almost the same) from what you describe above calling `.Get`.  

When I initially said that I was having problems I meant that I'm having troubles with the last assertion, not understanding why that is. I don't have issues by calling `ConfigureClient` directly event if it wasn't meant to be used like that.

atodorov: Note2: I think my issue comes from the fact that Client.Client is an interface (HttpRequestDoer) with a concrete implementation/value as http.Client but I'm unable to convert this in order to assert that TLS was configured on the connection.

Gundersanne: Ah I think I understand it now. The problem is that the generated cloudapi_client.go defines the Client struct as having HttpRequestDoer, yet at runtime Client gets overwritten with the http.Client by ConfigureClient. This results in all kinds of compiler errors, and really the generated code should just use `http.Client`.

Tbh I'm tempted to just not use this generated client and write our own cuz it's so messy, and the generated client contains 2 clients for some reason.

Anyway the only way to actually check this is to use reflection, and make sure the Transport field is there, the very basic version would be something like:
```
	s := reflect.ValueOf(openapi_client.Client).Elem()
	f := s.FieldByName("Transport")
	require.True(t, f.IsValid())
```

But honestly I need to rethink this client.

Gundersanne (approved):
See the above comment. If you wanna merge this that's fine. And then I'll change the client so it makes a bit more sense.

atodorov:
> See the above comment. If you wanna merge this that's fine. I'll change the client so it makes a bit more sense in future.

Gundersanne (approved):

PR #69, opened 2020-12-08
More tests for cloudapi_client

atodorov:
This builds on top of #67

### DIFF
--- a/internal/cloudapi/client_test.go
+++ b/internal/cloudapi/client_test.go
@@ -3,6 +3,7 @@ package cloudapi
 import (
 	"github.com/stretchr/testify/require"
 	"net/http"
+	"reflect"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


Gundersanne:
Hm the generated cloudapi_client got removed by #70, and the `client_test.go` now tests all there is. And without the mess with the types the `reflect` bit isn't necessary either.

So if you're good with it, let's just close this one?

atodorov:
> Hm the generated cloudapi_client got removed by #70, and the client_test.go now tests all there is. And without the mess with the types the reflect bit isn't necessary either.

Yup, will close. 

However do we still want to assert that the client has TLS configured when passing certs and a https URL. I can remove the reflection from https://github.com/osbuild/image-builder/pull/69/commits/1748bab737988a1f6b717d0295dea5caac95fccf and use `require.IsType` directly. IMO it's a valid assertion.

PR #70, opened 2020-12-09
cloudapi: Stop using generated http clients

Gundersanne:
The generated clients were bloated, and difficult to test. Instead use a
simple http client.


atodorov (approved):
LGTM

PR #71, opened 2020-12-09
tests: Assert that we have TLS configured

atodorov:


atodorov:
Hmm, I'm getting `type http.RoundTripper has no field or method TLSClientConfig` so it looks like reflection may still be needed.

Gundersanne:
> Hmm, I'm getting `type http.RoundTripper has no field or method TLSClientConfig` so it looks like reflection may still be needed.

I mostly referred to the reflect thing to indicate how far we'd have to go to test it. I don't think using reflection like this would be a good idea, it's an implementation detail of the http package. http.RoundTripper is an interface on top of that, not a struct, so that means it should just have the 'RoundTrip' function implemented.

If we want to test tls as well we should construct a test tls server as in`TestComposeStatusWithHTTPServer`. However it might be a bit tricky because the client certificates would need to be generated with different -subj as in `generate-certs.sh`

atodorov:
> > Hmm, I'm getting `type http.RoundTripper has no field or method TLSClientConfig` so it looks like reflection may still be needed.
> 
> I mostly referred to the reflect thing to indicate how far we'd have to go to test it. I don't think using reflection like this would be a good idea, it's an implementation detail of the http package. http.RoundTripper is an interface on top of that, not a struct, so that means it should just have the 'RoundTrip' function implemented.
> 
> If we want to test tls as well we should construct a test tls server as in`TestComposeStatusWithHTTPServer`. However it might be a bit tricky because the client certificates would need to be generated with different -subj as in `generate-certs.sh`

Thanks for your explanation. This turns out to be more hassle than I anticipated originally. Scraping this PR then. If need be we'll come back to this.

PR #73, opened 2020-12-10
Add tests for logger package

atodorov:


internal/logger/logger_test.go
```diff
@@ -0,0 +1,58 @@
+package logger
+
+import (
+	"github.com/sirupsen/logrus"
```
Gundersanne, line 4: imports should always be formatted as 
```
import (
        "some/builtin/package"
        "some/builtin/package2"

        "github.com/some/external/package"

        "github.com/osbuild/image-builder/internal/ourownstuff"
)

Gundersanne (changes requested):
Just the import formatting thing, otherwise looks good.

internal/logger/logger_test.go
atodorov: Prepare source says otherwise https://github.com/osbuild/image-builder/pull/73/checks?check_run_id=1537038729. I will also try with a blank line in between but I don't think that matters.

atodorov: OK, I was wrong. The blank lines do matter. Didn't know that.

Gundersanne: Yea it's not checked in prepare-source. It's just a convention I guess.

Gundersanne (approved):

PR #74, opened 2020-12-11
Add tests for distributions.go

atodorov:


internal/server/distribution_test.go
```diff
@@ -0,0 +1,16 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestAvailableDistributions(t *testing.T) {
+	result, err := AvailableDistributions()
+	require.NoError(t, err)
+	for _, distro := range result {
+		// require.Contains(t, ["fedora-32", "rhel-8"], distro.Name)
```
atodorov, line 13: @Gundersanne the commented out line fails with:
```
Error: internal/server/distribution_test.go:13:35: expected ']', found ','
Error: internal/server/distribution_test.go:15:3: missing ',' in argument list
```
https://github.com/osbuild/image-builder/runs/1537517575?check_suite_focus=true

and I'd love to know why. Spent too much time trying to figure it out. Reusing example from:
https://godoc.org/github.com/stretchr/testify/require#Contains

Gundersanne: The correct way to construct a slice inline is `[]type{}`, as you do below. So you can just remove this line :)

Gundersanne (changes requested):

internal/server/distribution_test.go
atodorov: Does this mean the examples in testify's documentation are wrong ?

atodorov: > So you can just remove this line as the check below does exactly that.

Line removed.

Gundersanne (approved):

PR #78, opened 2020-12-16
templates: Add osbuild-composer secrets to deployment

Gundersanne:
- [ ] secrets need to be in place
- [x] might need a rebase on master intead of on `db`


larskarlitski (approved):
This looks good to me, as far as I can judge it :roll_eyes:

PR #79, opened 2020-12-16
db: Add database

Gundersanne:
- [x] fork main to stable for prod deployment


teg:
@larskarlitski looked into this in the context of osbuild-composer. Maybe we should have a chat to figure out the way forward to make sure we are all on the same page, or did you guys already speak about it?

internal/db/db.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (db *DB) InsertCompose(uuid string, request json.RawMessage, accountId, orgId string) error {
+	ctx := context.Background()
+	conn, err := db.Pool.Acquire(ctx)
+	if err != nil {
+		return err
+	}
+	defer conn.Release()
+
+	_, err = conn.Query(ctx, "INSERT INTO composes(id, request, created_at, account_id, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4);", uuid, request, accountId, orgId)
```
teg, line 45: I stumbled on the `created_at` as it repeats data that should be in `composer`. However, I realize that we (will probably) need this in order paginate, and not having to query composer for all composes all the time, but only the ones we want to show on screen. So: all good! (this was a vacuous comment)

teg (changes requested):
Love this!

Only request: Please add a more detailed commit message to explain the overall scheme (how migrations work, what belongs in this database and what does not, ...).

Gundersanne:
So coverage dropped a lot here, but since the db code pretty much needs a database running i'm not sure how to do it within a unit test so that it would actually result in some added value.

The integration test covers it of course.

internal/server/server.go
```diff
@@ -259,6 +331,26 @@ func (h *Handlers) ComposeImage(ctx echo.Context) error {
...
+	ih := ctx.Get("IdentityHeader")
+	if ih == nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, "Identity Header missing in request handler")
+	}
+	idHeader, ok := ih.(IdentityHeader)
+	if !ok {
+		return echo.NewHTTPError(http.StatusInternalServerError, "Identity Header invalid in request handler")
+	}
+
+	err = h.server.db.InsertCompose(composeResult.Id, idHeader.Identity.AccountNumber, idHeader.Identity.Internal.OrgId, rawCR)
```
teg, line 349: In the future we may want to do this before creating the actual compose request, to make sure we only schedule a compose once we have successfully recorded it. Especially if we enforce caps.

internal/server/server.go
```diff
@@ -196,6 +208,66 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId string) error {
...
+			len(composes),
+		},
+		Links: struct {
+			First string `json:"first"`
+			Last  string `json:"last"`
+		}{
+			fmt.Sprintf("%v/v%v/composes?offset=%v&limit=%v",
+				RoutePrefix(), spec.Info.Version, offset, limit),
+			fmt.Sprintf("%v/v%v/composes?offset=%v&limit=%v",
+				RoutePrefix(), spec.Info.Version, len(composes)-1, limit),
```
teg, line 265: What exactly should the semantics of the First and Last links be? In particular, should the `offset` be rounded down to start on the last `limit` interval?

internal/db/migrations/1_create_table_images.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS composes;
```
teg, line 1: What's the intended use of this?

internal/server/api.yaml
```diff
@@ -70,6 +70,27 @@ paths:
...
                 $ref: '#/components/schemas/Architectures'
+  /composes:
+    get:
+      summary: get a collection of previous compose requests for the logged in user
+      parameters:
+        - in: query
+          name: limit
+          schema:
+            type: integer
+          description: max amount of packages, default 100
```
teg, line 81: `composes`, not `packages`

internal/server/api.yaml
```diff
@@ -70,6 +70,27 @@ paths:
...
+        - in: query
+          name: offset
+          schema:
+            type: integer
+          description: packages page offset, default 0
```
teg, line 86: ditto

internal/server/api.yaml
```diff
@@ -283,6 +304,46 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/UploadRequest'
+    ComposesResponse:
+      required:
```
teg, line 308: Where are these documented? I assume they are standard?

teg (dismissed):
Very minor questions / nit, nothing that can't be answered / addressed after merging.

internal/db/db.go
Gundersanne: yea I mean created_at maybe shld be inserted_at, that could be more correct

internal/server/server.go
Gundersanne: I agree, but the problem is we need the job uuid as that's the primary key here. What we could do is of course insert it with a random key, and then once we get the actual uuid, update the row.

internal/server/server.go
Gundersanne: I think we have some freedom here, but this was mostly a copy and paste from the packages endpoint. I'm not entirely sure what's more correct, I guess offset = lastOffset+limit is more correct.

internal/db/migrations/1_create_table_images.down.sql
Gundersanne: Completeness, for every migration up there should be a migration down. In case we ever reach a situation where we wanna migrate down, it's very unlikely we'll reach this migration ever, but idk completeness :)

internal/server/api.yaml
Gundersanne: Copy and paste error :facepalm:

internal/server/api.yaml
Gundersanne: Copy and paste error :facepalm:

internal/server/api.yaml
Gundersanne: Hm what do you mean?

internal/server/server.go
teg: Ah.

internal/server/api.yaml
teg: What are the `meta` and `links` properties? How are they meant to be used? Why did you add them? I assumed they were some standard thing, in which case, where can I read more about them?

teg (approved):
Love this! My pagination nit-pick could be addressed in a follow-up together with any other optional things from the guidelines.

larskarlitski (approved):
Overall I like it, and most crucially I think it can go in as is. Thanks a lot for you work on this! ❤️ 

Some notes:
* I'm sceptical about including down-migrations. They're not used and might give people the wrong idea (like "down migrations work"). I don't think we could even make them work reliably in this model of running the migration when the container comes up.
* I think I'd be happier using an existing migration tool instead of rolling our own.
* Tying the schema and its version to the code makes a sense here, because image-builder is likely the only service accessing the database. For osbuild-composer, I separated it, because I thought multiple services will access it. This makes it more complex, because then the services have to check at startup if they're talking to a db with the right schemas applied, and we'll have to find a way to do migrations in a separate process. Good to go simple for now :)
* In a follow-up, it would be nice to pass the `context.Context` from requests instead of using `context.Background()` in db calls.

ondrejbudai (approved):
I quickly skimmed through this PR and it's amazing! :heart_eyes:

PR #80, opened 2020-12-17
More tests for functions talking to osbuild-composer

atodorov:


atodorov:
@Gundersanne, @ondrejbudai managed to get this working faster than I thought. Please review.

.github/workflows/tests.yml
```diff
@@ -47,7 +47,7 @@ jobs:
           ls -lR /etc/osbuild-composer-test/ca
 
       - name: Run unit tests
-        run: go test -v -race -covermode=atomic -coverprofile=coverage.txt -coverpkg=./... ./...
+        run: go test -p 1 -v -race -covermode=atomic -coverprofile=coverage.txt -coverpkg=./... ./...
```
Gundersanne, line 50: Is the `-p 1` still needed? I thought grouping the tests would have fixed that.

atodorov: Looks like it isn't. However I still see `Error starting echo server: http: Server close` in the logs regardless of the fact if we have `-p 1` or not. See

without `-p 1`:
https://github.com/osbuild/image-builder/pull/80/checks?check_run_id=1642988882

with `-p 1`:
https://github.com/osbuild/image-builder/runs/1575651107?check_suite_focus=true

all tests report PASS so I think we're good to merge.

ondrejbudai (approved):
Looks good!

PR #85, opened 2020-12-23
templates: Bump osbuild-composer version

Gundersanne:
We need 946a0b4 so we can run the composer-api on a non-privileged port.


larskarlitski (approved):
*nod*

PR #97, opened 2021-01-14
logger: Use hostname as cloudwatch stream

Gundersanne:


internal/logger/logger.go
```diff
@@ -114,10 +104,12 @@ func NewLogger(level string, key *string, secret *string, region *string, group
 	}
 
 	if key != nil {
-		log.SetFormatter(NewCloudwatchFormatter())
```
msehnout, line 117: Why was this line removed?

Gundersanne: it moved to line 108. Just assigning the formatter to a variable since it's used on line 112 as well.

msehnout (approved):
Looks good to me! Thanks for figuring this out, hope it helps :-)

PR #98, opened 2021-01-14
tests: Start using http.Client everywhere

atodorov:


Gundersanne (approved):
Nice! Good to have those generated clients gone :)

PR #107, opened 2021-01-19
server: Allow filtering on org_id from the identity header

Gundersanne:
For now when the environment is empty the filter won't apply.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -2,6 +2,7 @@
 package server
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -22,6 +23,7 @@ type Server struct {
 	echo     *echo.Echo
 	client   cloudapi.OsbuildClient
 	awsCreds *awsCreds
+	orgIds   string
 }
 
 type awsCreds struct {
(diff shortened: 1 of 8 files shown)


internal/server/server.go
```diff
@@ -270,18 +273,47 @@ func (b binder) Bind(i interface{}, ctx echo.Context) error {
...
+	} `json:"identity"`
+}
+
+// clouddot guidelines requires 404 instead of 403 when unauthorized
+func (s *Server) VerifyIdentityHeader(nextHandler echo.HandlerFunc) echo.HandlerFunc {
 	return func(ctx echo.Context) error {
 		request := ctx.Request()
 
 		// For now just check it's there, in future we might want to
 		// decode the b64 string and check a specific entitlement
```
teg, line 290: We decode it now, so drop the comment?

teg (changes requested):
Love it! This is exactly what we need :) One nitpick, and there were a couple of warnings about missing tests. Could you either add the tests, or add a comment to the commit message to say why not?

Gundersanne:
@teg I added a bunch of unit tests. The integration tests only test the expect path. Do you want me to replicate the unit tests in the integration tests?

teg:
Nah, no need to replicate the tests. I'm totally fine with saying "the warnings are moot because of reasons". Just thought I'd ask since we have them enabled ;)

Gundersanne:
Oh, i tuned those out automatically, probably a bad reflix! Hm I could easily add a test for one of those. I'll do that and then ask for a rereview :)

teg:
> Oh, i tuned those out automatically, probably a bad reflix! Hm I could easily add a test for one of those. I'll do that and then ask for a rereview :)

Hehe, I know the reflex. Reviewers are here to tune them back in :P

teg (approved):
Great! Thanks for this!

PR #114, opened 2021-01-27
distribution: Remove db variables echo from startup script

Gundersanne:
Let's start doing this when the database PR is in.


larskarlitski (approved):
Ya. Sorry.

PR #116, opened 2021-01-29
cloudapi: Move path prefix to the client itself

Gundersanne:


major (approved):
This looks good to me.

PR #117, opened 2021-01-29
server: Fix edge case in organization id filter

Gundersanne:
The previous check to see if the id was contained in a string didn't
take into account that organization ids might differ in length. And thus
if any id was a substring of the allowed ids, it could get acccess.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"strings"
 
 	"github.com/osbuild/image-builder/internal/cloudapi"
 	"github.com/osbuild/image-builder/internal/logger"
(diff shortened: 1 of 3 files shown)


teg (approved):
Nice! Do we want to address the codecov warnings, or what should we make the policy around this?

Gundersanne:
@teg depends on the PR really. In this case the reduced .35% coverage is because there's 2 extra lines in main which isn't tested and is hard to capture in a unit test.  As a general rule for any new code we should of course add tests, but this is a followup/fix of PR #107 which increased the coverage by .83%. so eh personally I don't think we should be super dogmatic about it, as long as any new code is tested, especially if it's in internal modules.

teg:
> @teg depends on the PR really. In this case the reduced .35% coverage is because there's 2 extra lines in main which isn't tested and is hard to capture in a unit test.  As a general rule for any new code we should of course add tests, but this is a followup/fix of PR #107 which increased the coverage by .83%. so eh personally I don't think we should be super dogmatic about it, as long as any new code is tested, especially if it's in internal modules.

ah, yeah, i don't mind the percentages. just thought of the inline warnings. i don't have an answer, but might make sense to think of some guidelines. "ok to ignore when,..."

PR #121, opened 2021-02-08
Dnf getpackages

Gundersanne:


internal/server/repository.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (err *DNFError) Error() string {
+	return fmt.Sprintf("DNF error occured: %s: %s", err.Kind, err.Reason)
+}
+
+type Repo struct {
+	Baseurl    *string `json:"baseurl,omitempty"`
+	Metalink   *string `json:"metalink,omitempty"`
+	Mirrorlist *string `json:"mirrorlist,omitempty"`
+	Rhsm       bool    `json:"rhsm"`
```
teg, line 28: We need the container to be subscribed now, I guess?

teg (approved):
Nice! Though do we need some extra magic for RHSM?

internal/server/repository.go
Gundersanne: ah dang you're right.

templates/image-builder.yml
```diff
@@ -210,3 +212,6 @@ parameters:
   - name: ALLOWED_ORG_IDS
     description: Organization ids allowed to access the api, wildcard means everyone
     value: ""
+  - name: DISTRIBUTIONS_DIR
+    description: Directory which dnf-json uses as a cache
```
teg, line 216: Wrong description?

dnf-json
```diff
@@ -0,0 +1,200 @@
+#!/usr/bin/python3
```
teg, line 1: This is now used only to populate the distro files, right? In which case: we should probably document somewhere how to re-run that?

teg (approved):
Thanks for fixing this up! Happy to merge as-is and address any minor issues in follow-ups!

Gundersanne:
@teg waiting on jenkins currently before merging, agree with all of your comments :)

dnf-json
Gundersanne: Made https://github.com/osbuild/image-builder/issues/126 to track that

teg (approved):
Perfection!

PR #123, opened 2021-02-08
distribution: Change to ubi container instead of minimal

Gundersanne:


major (approved):
Checked this locally and it works fine. The container will be a little bigger in quay, but it's not a huge deal.

PR #128, opened 2021-02-11
schutzbot: ⚡Trigger schutzbot using github actions

Gundersanne:
I'm not sure if there's a reason for security concern here, do we need specific branch protections or other repo settings for this ?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -52,6 +52,17 @@ jobs:
       - name: Send coverage to codecov.io
         run: bash <(curl -s https://codecov.io/bash)
 
+      - name: Trigger Schutzbot
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          WEBHOOK_PAYLOAD: ${{ toJSON(github.event) }}
+          SQS_REGION: us-east-1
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -52,6 +52,19 @@ jobs:
       - name: Send coverage to codecov.io
         run: bash <(curl -s https://codecov.io/bash)
 
+      - name: Trigger Schutzbot
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          WEBHOOK_PAYLOAD: ${{ toJSON(github.event) }}
+          SQS_REGION: us-east-1
+          SQS_QUEUE_URL: "https://sqs.us-east-1.amazonaws.com/933752197999/schutzbot_webhook_sqs-staging"
+          AWS_ACCESS_KEY_ID: ${{ secrets.WEBHOOK_AWS_ACCESS_KEY_ID }}
```
major, line 61: You can remove the credentials here since we're using anonymous unsigned requests to drop off messages.

major (changes requested):

major (approved):

PR #130, opened 2021-02-15
Readiness probe

msehnout:


teg (dismissed):
Nice!

msehnout:
I've made 1 more force-push to make the test less confusing. The "fake" osbuild-composer now returns:
```
fmt.Fprint(w, "{\"version\":\"fake\"}")
```
which does not have any impact given that I don't even check the body of the response, but I use a request to /version endpoint so this makes more sense.

teg (approved):

Gundersanne (approved):

Gundersanne:
Thanks, looks good!

PR #131, opened 2021-02-15
Bump golang version to 1.14.14

msehnout:
This is the same version osbuild-composer uses.


ondrejbudai (changes requested):
Please, also change the version in `.github/workflows/tests.yml`

msehnout:
> Please, also change the version in `.github/workflows/tests.yml`

Good catch, now I see why grep for `1.12` didn't help.

ondrejbudai (approved):

PR #133, opened 2021-02-17
distributions: add centos stream 8 support

ondrejbudai:
The only question is: Should we use the `mirrorlist`? It's the official way to consume CentOS repos but if get unlucky, we can have a lot of failed builds... Maybe we should just choose one mirror and pin it for now?


teg:
Let's start out with the official one, and then see what we do if that does not work out?

teg (approved):
So cool! Thanks!

ondrejbudai:
Please don't merge it yet. I need to update composer in Schutzbot. ;)

ondrejbudai:
Ah, Sanne already updated composer in `main`, so rebase was everything I needed. This is now ready to go in!

Gundersanne (approved):

PR #137, opened 2021-02-24
template: Use the new readiness probe endpoint

msehnout:
Use separate endpoints for the liveness and the readiness probes as now
image-builder has both /status and /ready.

Depends on: #130


Gundersanne (dismissed):
Happy to merge this, is 5 seconds enough?

msehnout:
> Happy to merge this, is 5 seconds enough?

It's hard to say. I made it 10.

Gundersanne (approved):

PR #139, opened 2021-02-24
server: Add support for gcp target

Gundersanne:
- [x] osbuild/osbuild-composer#1245

### DIFF
--- a/cmd/image-builder-tests/main_test.go
+++ b/cmd/image-builder-tests/main_test.go
@@ -69,7 +69,7 @@ func RunTestWithClient(t *testing.T, ib string)  {
 		ImageRequests: []server.ImageRequest{
 			server.ImageRequest{
 				Architecture: arch,
-				ImageType: architectures[0].ImageTypes[0],
+				ImageType: "ami",
 				UploadRequests: []server.UploadRequest{
 					{
 						Type: "aws",
@@ -93,7 +93,6 @@ func RunTestWithClient(t *testing.T, ib string)  {
 	var composeResp server.ComposeResponse
 	err = json.Unmarshal([]byte(body), &composeResp)
 	require.NoError(t, err)
-
 	id := composeResp.Id
 
 	response, body = tutils.GetResponseBody(t, ib + "/composes/" + id, &tutils.AuthString0)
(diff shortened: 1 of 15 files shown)


internal/server/api.yaml
```diff
@@ -291,16 +300,64 @@ components:
...
+                service account. For example, 'my-other-app@appspot.gserviceaccount.com'.
+              - 'group:{emailid}': An email address that represents a Google group.
+                For example, 'admins@example.com'.
+              - 'domain:{domain}': The G Suite domain (primary) that represents all
+                the users of that domain. For example, 'google.com' or 'example.com'.
+                If not specified, the imported Compute Node image is not shared with any
+                account.
           items:
             type: string
+    Customizations:
```
thozza, line 356: It seems like the `packages` customization got removed as part of these changes. Is this intentional?

Gundersanne: It just needed rebasing, this was a bit of a mess.

cmd/image-builder-tests/main_test.go
```diff
@@ -106,6 +105,42 @@ func RunTestWithClient(t *testing.T, ib string)  {
 	require.NoError(t, err)
 	require.Contains(t, []string{"pending", "running"}, composeStatus.ImageStatus.Status)
 
+	// A compose with gcp as a target
+	composeRequest = server.ComposeRequest{
+		Distribution: distro,
+		ImageRequests: []server.ImageRequest{
+			server.ImageRequest{
+				Architecture: arch,
+				ImageType: architectures[0].ImageTypes[0],
```
thozza, line 114: My assumption is that this is the distro-specific list of supported architectures and image types. So from looking at `internal/distro/rhel84/distro.go` from `osbuild-composer`, the `[0]` arch is x86_64, which is fine, but the Image type `[0]` is AMI. This should be changed to VHD or VMDK image types for now.

cmd/image-builder-tests/main_test.go
```diff
@@ -106,6 +105,42 @@ func RunTestWithClient(t *testing.T, ib string)  {
...
+				UploadRequests: []server.UploadRequest{
+					{
+						Type: "gcp",
+						Options: server.GCPUploadRequestOptions{
+							Bucket: "some-bucket",
+							ShareWithAccounts: &[]string{"somebody@example.com"},
```
thozza, line 120: The account format must be one of:
- `user:{emailid}`
- `serviceAccount:{emailid}`
- `group:{emailid}`
- `domain:{domain}`

Plain email won't work.

Gundersanne: Ah ok, I'll change that. It doesn't matter hugely since it never actually completes the compose, it just makes sure the request sent is valid.

cmd/image-builder-tests/main_test.go
Gundersanne: ah good point

teg (dismissed):
Thanks!

templates/image-builder.yml
```diff
@@ -215,3 +220,9 @@ parameters:
   - name: DISTRIBUTIONS_DIR
     description: Directory which contains json files detailing available distributions, their repositories, and their packages
     value: "/app/distributions"
+  - name: OSBUILD_GCP_REGION
+    description: Region in GCP to upload to
+    value: "us-east4"
```
Gundersanne, line 225: @thozza do these values make sense?

templates/image-builder.yml
```diff
@@ -215,3 +220,9 @@ parameters:
...
+  - name: OSBUILD_GCP_BUCKET
+    description: Bucket in GCP to upload to
+    value: "image-upload-bkt-us-east4"
```
thozza, line 228: This should be just "image-upload-bkt-us". The Storage bucket is multi-regional, so there is no need to have one per each region. There will be just one bucket for each - US, EU, ASIA (https://cloud.google.com/about/locations#multi-region)

templates/image-builder.yml
thozza: This value makes sense and it is OK!

teg (approved):

PR #147, opened 2021-03-07
Add support for Azure

ondrejbudai:
This commit adds support for the Azure target throughout all image-builder layers. Also, osbuild-composer's version in provision-composer was bumped to support the azure target.


teg (approved):
Nice! \o/

PR #153, opened 2021-03-10
README: Move test README to test/ and add project description

croissanne:
I'm not sure if we should already reference the target environment, as we're not actually deployed there yet publicly. But maybe at some point we should add a link to where it's running :)

---

- [x] update the test README


teg:
No problem to link to it, we do it all over the place (check out osbuild.org/guides).

PR #156, opened 2021-03-17
Expose upload options via API and extend integration testing for all cloud providers

thozza:
Update composer cloudapi types related to upload options.

Expose upload options, returned from composer, in the image-builder API.
These contain all the necessary information to identify the uploaded
image in the specific cloud provider environment.

Integration testing:
- Test compose request for all supported cloud providers (AWS, Azure,
  GCP).
- Verify the returned upload options for each cloud provider. Use it
  to verify the resulting image from the compose by creating a VM in the
  respective cloud provider environment and connecting to it via SSH
  (except for Azure, where it is not possible because of a bug).
- Modify the way test cases are executed based on PR#79. Specifically
  that the compose is tested only once for `BaseUrl`.
- Clean up WORKDIR at the end of the test run.
- Update test/README.md section on integration testing.

schutzbot/deploy.sh:
- Create osbuild-worker configuration directory.
- Create osbuild-worker's configuration file with necessary credentials
  for Azure or GCP based on available environment variables.
- Pass necessary configuration environment variables to the
  image-builder container.

schutzbot/Jenkinsfile:
- After previous discussion test image-builder only on RHEL-8 in
  parallel for each supported cloud provider.
- Set necessary environment variables for each cloud-provider
  integration testing.

### DIFF
--- a/internal/server/api.yaml
+++ b/internal/server/api.yaml
@@ -239,21 +239,50 @@ components:
       required:
         - status
         - type
+        - options
       properties:
         status:
           type: string
           enum: ['success', 'failure', 'pending', 'running']
         type:
           $ref: '#/components/schemas/UploadTypes'
-    UploadTypes:
-      type: string
-      enum: ['aws', 'gcp', 'azure']
+        options:
+          oneOf:
+            - $ref: '#/components/schemas/AWSUploadStatus'
+            - $ref: '#/components/schemas/GCPUploadStatus'
+            - $ref: '#/components/schemas/AzureUploadStatus'
     AWSUploadStatus:
       type: object
+      required:
+        - ami
+        - region
       properties:
-        ami_id:
+        ami:
           type: string
           example: 'ami-0c830793775595d4b'
+        region:
+          type: string
+          example: 'eu-west-1'
+    GCPUploadStatus:
+      type: object
+      required:
+        - project_id
+        - image_name
+      properties:
+        project_id:
+          type: string
+          example: 'ascendant-braid-303513'
+        image_name:
+          type: string
+          example: 'my-image'
+    AzureUploadStatus:
+      type: object
+      required:
+        - image_name
+      properties:
+        image_name:
+          type: string
+          example: 'my-image'
     ComposeRequest:
       type: object
       required:
@@ -302,6 +331,9 @@ components:
             - $ref: '#/components/schemas/AWSUploadRequestOptions'
             - $ref: '#/components/schemas/GCPUploadRequestOptions'
             - $ref: '#/components/schemas/AzureUploadRequestOptions'
+    UploadTypes:
+      type: string
+      enum: ['aws', 'gcp', 'azure']
     AWSUploadRequestOptions:
       type: object
       required:
--- a/schutzbot/deploy.sh
+++ b/schutzbot/deploy.sh
@@ -33,6 +33,33 @@ sudo dnf install -y osbuild-composer composer-cli
 sudo mkdir -p /etc/osbuild-composer
 sudo cp -a schutzbot/osbuild-composer.toml /etc/osbuild-composer/
 
+sudo mkdir -p /etc/osbuild-worker
+
+# if GCP credentials are defined in the ENV, add them to the worker's configuration
+GOOGLE_APPLICATION_CREDENTIALS="${GOOGLE_APPLICATION_CREDENTIALS:-}"
+if [[ -n "$GOOGLE_APPLICATION_CREDENTIALS" ]]; then
+     # The credentials file must be copied to a different location. Jenkins places
+     # it into /tmp and as a result, the worker would not see it due to using PrivateTmp=true.
+     GCP_CREDS_WORKER_PATH="/etc/osbuild-worker/gcp-credentials.json"
+     sudo cp "$GOOGLE_APPLICATION_CREDENTIALS" "$GCP_CREDS_WORKER_PATH"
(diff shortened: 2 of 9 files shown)


schutzbot/deploy.sh
```diff
@@ -33,6 +33,35 @@ sudo dnf install -y osbuild-composer composer-cli
...
+     GCP_CREDS_WORKER_PATH="/etc/osbuild-worker/gcp-credentials.json"
+     sudo cp "$GOOGLE_APPLICATION_CREDENTIALS" "$GCP_CREDS_WORKER_PATH"
+     echo -e "\n[gcp]\ncredentials = \"$GCP_CREDS_WORKER_PATH\"\n" | sudo tee -a /etc/osbuild-worker/osbuild-worker.toml
+fi
+
+# if Azure credentials are defined in the env, create the credentials file
+AZURE_CLIENT_ID="${AZURE_CLIENT_ID:-}"
+AZURE_CLIENT_SECRET="${AZURE_CLIENT_SECRET:-}"
+if [[ -n "$AZURE_CLIENT_ID" && -n "$AZURE_CLIENT_SECRET" ]]; then
+     set +x
```
msehnout, line 52: Why the set +x here?

thozza: to not to leak credentials in CI log

Gundersanne: It would dump the secrets in the logs otherwise but I have no idea if we take the same precautions in composer.

msehnout (commented):
I can't really tell if the test/cases/api.sh changes are correct as I do not know all the CLI clients, but otherwise this looks very cool. And the docs: :heart:

schutzbot/deploy.sh
msehnout: Shouldn't Jenkins take care of this? It's set to mask these, isn't it?

thozza: we do the same thing in composer. I basically copied the Azure part from composer tests 1:1

schutzbot/deploy.sh
```diff
@@ -33,6 +33,35 @@ sudo dnf install -y osbuild-composer composer-cli
...
+if [[ -n "$GOOGLE_APPLICATION_CREDENTIALS" ]]; then
+     # The credentials file must be copied to a different location. Jenkins places
+     # it into /tmp and as a result, the worker would not see it due to using PrivateTmp=true.
+     GCP_CREDS_WORKER_PATH="/etc/osbuild-worker/gcp-credentials.json"
+     sudo cp "$GOOGLE_APPLICATION_CREDENTIALS" "$GCP_CREDS_WORKER_PATH"
+     echo -e "\n[gcp]\ncredentials = \"$GCP_CREDS_WORKER_PATH\"\n" | sudo tee -a /etc/osbuild-worker/osbuild-worker.toml
+fi
+
+# if Azure credentials are defined in the env, create the credentials file
+AZURE_CLIENT_ID="${AZURE_CLIENT_ID:-}"
```
msehnout, line 49: If anything the set +x should go up here, no? Because here it "leaks" anyway.

thozza: No, it's fine - https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fimage-builder/detail/PR-156/7/pipeline/25#step-66-log-191

schutzbot/deploy.sh
thozza: Maybe @ondrejbudai can share his motivation as the original author of this piece of code in composer...

schutzbot/deploy.sh
ondrejbudai: Yeah, I didn't want to expose any secrets. `set +x` should definitely be several lines higher - I had it right originally but I think I got it wrong after several iterations.

msehnout: So it can be removed completely, right? Because Jenkins masks the "leaks" anyway.

thozza: it's removed now...

internal/server/api.yaml
```diff
@@ -239,21 +239,50 @@ components:
       required:
         - status
         - type
+        - options
```
Gundersanne, line 242: Hm I know we marked this as required in composer as well, but really it'll only contain meaningful data if the status property === 'success', just a thought.

Gundersanne (approved):
Looking good! Just have a thought/comment I'd like your take on.

internal/server/api.yaml
thozza: Yes, your comment makes sense. I probably didn't hit this issue, because the upload always passed in CI so far... I'll change it to not be required (and test it with e.g. non existent GCP account for sharing). Another thought I just had while writing documentation for image builder API is that it seems that we don't return any errors, just status. This may not be ideal for the user because they may not be able to figure out the reason for the failure...

thozza: So this is not really a problem, because the `upload_status` is not part of the response at all in case the upload failed. See https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fimage-builder/detail/PR-156/9/pipeline#step-36-log-788

Gundersanne: Alright keep it as required then, I think it that makes sense

teg (approved):
This is really great! Love the testing and docs! ❤️

Converted to `draft` as the last commit is marked as DON'T MERGE.

thozza:
I rebased the branch on top of `main` and removed the last commit, which was simulating image upload failure.

Gundersanne (approved):
Nice! thank you :)

PR #157, opened 2021-03-18
server: un-disable gosec

chloenayon:
Previously contained gosec overrides for G304 and G307, amend code so that the
overrides can be safely removed without triggering gosec violations. Clean
filepaths before calling os.Open and remove instances of deferring f.Close
so that errors can be caught before return.

### DIFF
--- a/internal/server/distribution.go
+++ b/internal/server/distribution.go
@@ -38,11 +38,17 @@ func ReadDistributions(distsDir, distro string) ([]DistributionFile, error) {
 			return nil
 		}
 
-		f, err := os.Open(path) // #nosec G304
+		cleanPath := filepath.Clean(path)
+		f, err := os.Open(cleanPath)
 		if err != nil {
 			return err
 		}
-		defer f.Close() // #nosec G307
+		defer func() {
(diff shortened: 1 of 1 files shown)


internal/server/distribution.go
```diff
@@ -38,16 +38,20 @@ func ReadDistributions(distsDir, distro string) ([]DistributionFile, error) {
...
 		if err != nil {
 			return err
 		}
-		defer f.Close() // #nosec G307
 		var d DistributionFile
 		err = json.NewDecoder(f).Decode(&d)
 		if err != nil {
 			return err
 		}
+		err = f.Close()
```
Gundersanne, line 51: Hmm, if the decoding errors out the file will never be closed now. I think what's best is to change the defer function a little, see https://github.com/securego/gosec/issues/512#issuecomment-675286833 ; wdyt?

chloenayon: Yes, that makes sense - thanks for pointing that out! I'll make changes

teg (approved):
Really nice cleanups :)

PR #159, opened 2021-03-19
server: return bad request when distribution unsupported

chloenayon:
Return 400 Bad Request instead of 500 when the distribution is
not supported.

### DIFF
--- a/internal/server/distribution.go
+++ b/internal/server/distribution.go
@@ -3,12 +3,14 @@ package server
 import (
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"os"
 	"path"
 	"path/filepath"
 	"sort"
 	"strings"
 
+	"github.com/labstack/echo/v4"
 	"github.com/osbuild/image-builder/internal/cloudapi"
 )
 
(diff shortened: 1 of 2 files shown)


internal/server/server.go
```diff
@@ -210,7 +210,12 @@ func (h *Handlers) ComposeImage(ctx echo.Context) error {
 
 	repositories, err := RepositoriesForImage(h.server.distsDir, composeRequest.Distribution, composeRequest.ImageRequests[0].Architecture)
 	if err != nil {
-		return err
+		switch err.Error() {
+		case "Architecture not supported":
```
Gundersanne, line 214: I don't think we should be relying on the error message in order to return the correct error. 

Imo there's 2 options:
1. Since 'distribution.go' is still in `internal/server`, how about we just call `echo.NewHTTPError` inside the distribution?

2. Alternatively we define our own set of errors which can then be compared compliant with https://github.com/golang/go/wiki/ErrorValueFAQ (Error.Is())

I think I prefer 1 for simplicity's sake. Since we'd just end up redefining all possible http errors.

Gundersanne (commented):
Not sure if I like the approach of comparing error messages. Curious about your thoughts regarding alternatives.

internal/server/server.go
chloenayon: I agree with you, I think 1 is preferable - thanks for the input!

teg (approved):
Thanks for this!

PR #163, opened 2021-03-24
Use valid Cloud API values for `image_status.status`

thozza:
This PR is related to:
- https://github.com/osbuild/osbuild-composer/pull/1319
- https://github.com/osbuild/image-builder-frontend/pull/121

Cloud API definition of `ImageStatus.status` value has been updated to use reusable enum. This caused the OpenAPI generator to create a type alias for the value and also to define constant for each possible enum value. Update `internal/server/server.go` to use the type alias.

`osbuild-composer` is currently returning an invalid `ImageStatus.status` value for an in-progress image compose, specifically the value "running". The Cloud API specification does not allow this value.

Update Cloud API client tests and server tests to use valid `ImageStatus.status` value, instead of "running".

Update API test case to check for each valid `ImageStatus.status` value. Keep the "running" value as valid for some time to ensure backward compatibility with older `osbuild-composer` versions.

Update the oapi-codegen version to the same version as used in `osbuild-composer`. Regenerate the code for cloudapi and server APIs. The result of this change is that constant values are defined for reusable string enums.

Use generated `UploadTypes` values in the server implementation, instead of strings.

I didn't define the `ImageStatus.status` in the Image Builder API as reusable enum, because in my opinion it does not bring any value. Image Builder is basically copying the value from the `osbuild-composer`'s response without doing any further processing. This means that the generated constant values would not be used anywhere in the code. For the Cloud API this is a different story, as it makes sense to keep the exact copy of its definition from `osbuild-composer`, even if it results in some constant values not used in Image Builder's code.


teg (approved):
Nice! Thanks for the cleanups :)

Gundersanne (approved):
lgtm!

PR #169, opened 2021-03-31
server: validate input

chloenayon:
Use deepmap's OapiRequestValidator function to validate input.


ondrejbudai:
Please run `tools/prepare-source.sh`, commit and push the changes, something needs to be regenerated. :)

Gundersanne:
After trying a few things out, it seems we need to specify the relative baseurls in order for this to work properly. Try adding 
```
servers:
- url: "/api/image-builder/v1"
  description: Running locally
- url: "/api/image-builder/v1.0"
  description: Running locally
```
to the spec between `paths:` and `info:`

internal/server/api.yaml
```diff
@@ -8,6 +8,9 @@ info:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
 
+servers:
+  - url: "/api/image-builder/v1"
+  - url: "/api/image-builder/v1.0"
```
Gundersanne, line 13: nitpick, newline between `servers:` and `paths:`, just to be consistent with the newline between `info:` and `servers:`

internal/server/api.yaml
```diff
@@ -486,7 +489,6 @@ components:
           example: 'subscription.rhsm.redhat.com'
         base-url:
           type: string
-          format: url
```
Gundersanne, line 489: ah was this erroring?

internal/server/server_test.go
```diff
@@ -278,20 +278,30 @@ func TestComposeImage(t *testing.T) {
...
 				ImageRequest{
-					Architecture:  "x86_64",
-					ImageType:     "qcow2",
-					UploadRequest: UploadRequest{},
+					Architecture: "x86_64",
+					ImageType:    "qcow2",
+					UploadRequest: UploadRequest{
+						Type: "aws",
+						Options: AWSUploadRequestOptions{
+							ShareWithAccounts: []string{"test-account"},
```
Gundersanne, line 286: ah the accounts here are 12-digit numbers, so like `012345678912`, not sure if we can somehow force that format in the openapi spec though

Gundersanne (commented):
looking good so far, I know you're still iterating on the unit tests but some preliminary comments

internal/server/api.yaml
chloenayon: yes, I think it could be replaced with uri but wasn't sure at the time if they were interchangeable

internal/server/server.go
```diff
@@ -551,6 +553,35 @@ func (s *Server) VerifyIdentityHeader(nextHandler echo.HandlerFunc) echo.Handler
 	return func(ctx echo.Context) error {
 		request := ctx.Request()
 
+		loader := openapi3.NewSwaggerLoader()
+		doc, err := loader.LoadSwaggerFromFile("api.yaml")
```
Gundersanne, line 557: This won't actually work since api.yaml will not be in the same directory as the binary, in fact it's not really shipped at all.

What I recommend is doing the same thing `GetSwagger` is doing once when setting up the server (in `NewServer`), and the section here until `NewRouter`. And then assigning that router as a private field in the `Server` struct.

internal/server/server.go
```diff
@@ -551,6 +553,35 @@ func (s *Server) VerifyIdentityHeader(nextHandler echo.HandlerFunc) echo.Handler
...
+
+		if err := doc.Validate(loader.Context); err != nil {
+			panic(err)
+		}
+
+		router, err := legacyrouter.NewRouter(doc)
+		if err != nil {
+			panic(err)
+		}
+		route, params, err := router.FindRoute(request)
```
Gundersanne, line 570: This should move to a separate middleware imo. Which is slotted in after the  `VerifyIdentityHeader` middleware.

internal/server/server.go
```diff
@@ -551,6 +553,35 @@ func (s *Server) VerifyIdentityHeader(nextHandler echo.HandlerFunc) echo.Handler
...
+		if err != nil {
+			panic(err)
```
Gundersanne, line 572: Better not to use panic here, juts return the error. The Errorhandler will turn it into a 500, with panic the server will crash when handling requests. (Inside of `NewServer` it's fine to panic however, if anything goes wrong there we want to crash the server).

internal/server/server_test.go
```diff
@@ -308,7 +318,7 @@ func TestComposeImage(t *testing.T) {
 		}
 		response, body := tutils.PostResponseBody(t, "http://localhost:8086/api/image-builder/v1/compose", payload)
 		require.Equal(t, 400, response.StatusCode)
-		require.Contains(t, body, "Exactly one upload request should be included")
+		require.Contains(t, body, "Value is not nullable")
```
Gundersanne, line 321: Hmm I think it would make sense to expand the error message matching here a little to something less generic, the full error is 
```
{"errors":[{"detail":"request body has an error: doesn't match the schema: Error at \"/image_requests/0/upload_request/options\": Value is not nullable\nSchema:\n  {\n    \"anyOf\": [\n      {\n        \"$ref\": \"#/components/schemas/AWSUploadRequestOptions\"\n      },\n      {\n        \"$ref\": \"#/components/schemas/GCPUploadRequestOptions\"\n      },\n      {\n        \"$ref\": \"#/components/schemas/AzureUploadRequestOptions\"\n      }\n    ]\n  }\n\nValue:\n  null\n","title":"400"}]}
```

Like matching `image_requests/0/upload_request/options` to make sure it's the options which aren't nullable.

internal/server/server_test.go
```diff
@@ -355,7 +365,7 @@ func TestComposeImage(t *testing.T) {
 		}
 		response, body := tutils.PostResponseBody(t, "http://localhost:8086/api/image-builder/v1/compose", payload)
 		require.Equal(t, 400, response.StatusCode)
-		require.Contains(t, body, "Unknown UploadRequest type")
+		require.Contains(t, body, "value is not one of the allowed values")
```
Gundersanne, line 368: Same here, `...Error at \"/image_requests/0/upload_request/type\": value is not one of the allowed value...`, think it would be useful to include the type.

internal/server/server_test.go
chloenayon: Sounds good, I'll make changes, thanks!

internal/server/server.go
```diff
@@ -579,6 +595,29 @@ func (s *Server) VerifyIdentityHeader(nextHandler echo.HandlerFunc) echo.Handler
...
 }
 
+func (s *Server) ValidateRequest(nextHandler echo.HandlerFunc) echo.HandlerFunc {
+	return func(ctx echo.Context) error {
+		request := ctx.Request()
+		loader := openapi3.NewSwaggerLoader()
+
+		route, params, err := s.router.FindRoute(request)
+		if err != nil {
+			return echo.NewHTTPError(http.StatusBadRequest, err)
```
Gundersanne, line 605: I think here it would be an internal server error, no? Since any non-existing route would have returned 404 before it ever hits this. So just `return err`.

internal/server/server.go
```diff
@@ -66,12 +70,24 @@ func NewServer(logger *logrus.Logger, client cloudapi.OsbuildClient, dbase db.DB
...
+	loader := openapi3.NewSwaggerLoader()
+
+	if err := spec.Validate(loader.Context); err != nil {
+		panic(err)
+	}
+
+	router, err := legacyrouter.NewRouter(spec)
+	if err != nil {
+		panic(err)
+	}
```
Gundersanne, line 83: nitpick, but I'd space it like this:

```
        ...

        loader := openapi3.NewSwaggerLoader()
	if err := spec.Validate(loader.Context); err != nil {
		panic(err)
	}
	router, err := legacyrouter.NewRouter(spec)
	if err != nil {
		panic(err)
	}

        majorVersion := strings.Split(spec.Info.Version, ".")[0]
```

just to have each paragraph do a single thing.

internal/server/server.go
```diff
@@ -579,6 +595,29 @@ func (s *Server) VerifyIdentityHeader(nextHandler echo.HandlerFunc) echo.Handler
...
+		}
+
+		requestValidationInput := &openapi3filter.RequestValidationInput{
+			Request:    request,
+			PathParams: params,
+			Route:      route,
+		}
+
+		context := context.TODO()
+		if err := openapi3filter.ValidateRequest(context, requestValidationInput); err != nil {
```
ondrejbudai, line 614: I think we can just `ctx` here instead of creating a new context.

internal/server/api.yaml
```diff
@@ -386,7 +390,7 @@ components:
         type:
           $ref: '#/components/schemas/UploadTypes'
         options:
-          oneOf:
+          anyOf:
```
ondrejbudai, line 393: What's the reason for this change? I believe that we indeed want `oneOf` here instead of `anyOf`.

internal/server/server.go
chloenayon: I believe ctx is the wrong type - ValidateRequest takes context.Context and ctx is echo.context

internal/server/api.yaml
chloenayon: I agree that oneOf is preferable, the validator complains because AWSUploadRequestOptions and GCPUploadRequestOptions are identical except for the name.

internal/server/server.go
ondrejbudai: Ah, you're right, we need to get it from the request. `request.Context()` should return the right thing.

internal/server/api.yaml
ondrejbudai: Interesting, how can I reproduce the validator's complaint? I would still prefer somehow suppressing the complaint to changing the schema because I believe that `anyOf` doesn't reflect well the reality.

chloenayon: If you change it back to oneOf and run go test -v ./... you should get "{"errors":[{"detail":"request body has an error: doesn't match the schema: input matches more than one oneOf schemas","title":"400"}]}

ondrejbudai: Hmm, we can change the spec like this: https://github.com/ondrejbudai/image-builder/commit/fb4451c2b6e6a4d502b0d1be45b40d27d472d756

chloenayon: hmm, having some issues - did this work when you ran it?

Gundersanne: The problem here is that the aws and gcp upload requests share the exact same property names. So `AnyOf` might actually be correct from a purely syntactical pov. Logically there's a difference of course but well, the validator can't know that with just the spec in hand.

And the `type` enum is still a string and gets validated as such.

Gundersanne: Hm I was wrong about the enum type, the validator does consider it as more than a string. So moving oneOf up a level would work. I am a little hesitant for 2 reasons however:

1. I don't think it makes sense how the type turns into an enum with length of 1. Essentially acting as a constant, just to avoid the fact of having duplicate named fields. On top of that not using UploadTypes as well, the uploadtypes seems weird.
2. A more practical reason: oneOf/anyOf isn't really supported by the deepmap generator. So we have the type field set in the uploadrequest just to make parsing the options easier. Otherwise we'd need to parse the uploadrequest as a map first, then get the type, then parse it again as the correct upload request. 

So I'm not sure if it's worth it? Both options are not ideal I have to admit.

chloenayon: I'm leaning towards using anyOf, oneOf is what we want but anyOf does apply accurately to our spec as we've defined it

Gundersanne (approved):
Looks great, thank you. Let's get this in!

PR #173, opened 2021-04-22
server: Allow filtering based on account numbers

Gundersanne:


teg (approved):
At some point we may want to optimise this logic, and also be stricter about parsing it? I think for now it does not matter though.
I'm imagining a set of integers, with `*` treated specially, and failing during initialisation if the lists of numbers cannot be parsed properly.
For now: let's merge this!

PR #174, opened 2021-04-23
server: expand openapi spec

chloenayon:
Expand the openapi to describe content and parameters more specifically,
also edit some of the tests to reflect changes from the addition of the input 
validation logic.


internal/server/server_test.go
```diff
@@ -117,23 +117,28 @@ func TestWithoutOsbuildComposerBackend(t *testing.T) {
 		p1 := result.Data[0]
 		p2 := result.Data[1]
 
-		response, body = tutils.GetResponseBody(t, "http://localhost:8086/api/image-builder/v1/packages?distribution=rhel-8&architecture=x86_64&search=ssh&limit=1", &tutils.AuthString0)
+		response, body = tutils.GetResponseBody(t, "http://localhost:8086/api/image-builder/v1/packages?distribution=rhel-8&architecture=x86_64&search=ssh&limit=10", &tutils.AuthString0)
```
Gundersanne, line 120: Why up the limit to 10 here? I think testing the edges and just over the edges (so limit=1 and limit=0) is a good idea.

internal/server/server_test.go
```diff
@@ -117,23 +117,28 @@ func TestWithoutOsbuildComposerBackend(t *testing.T) {
...
+
 		response, body = tutils.GetResponseBody(t, "http://localhost:8086/api/image-builder/v1/packages?distribution=rhel-8&architecture=x86_64&search=ssh&limit=1&offset=1", &tutils.AuthString0)
 		require.Equal(t, 200, response.StatusCode)
 		err = json.Unmarshal([]byte(body), &result)
 		require.NoError(t, err)
 		require.Greater(t, result.Meta.Count, 1)
 		require.Equal(t, result.Data[0], p2)
 
-		response, _ = tutils.GetResponseBody(t, "http://localhost:8086/api/image-builder/v1/packages?distribution=rhel-8&architecture=x86_64&search=ssh&limit=-13", &tutils.AuthString0)
+		response, _ = tutils.GetResponseBody(t, "http://localhost:8086/api/image-builder/v1/packages?distribution=rhel-8&architecture=x86_64&search=ssh&limit=13", &tutils.AuthString0)
```
Gundersanne, line 139: I'd rather keep the negative values here and check it returns 400, the good cases are already tested above

internal/server/server_test.go
```diff
@@ -117,23 +117,28 @@ func TestWithoutOsbuildComposerBackend(t *testing.T) {
...
 		require.Equal(t, 200, response.StatusCode)
-		response, _ = tutils.GetResponseBody(t, "http://localhost:8086/api/image-builder/v1/packages?distribution=rhel-8&architecture=x86_64&search=ssh&limit=-13&offset=-2193", &tutils.AuthString0)
```
Gundersanne, line 136: How about we change this into a positive limit and a negative offset (and then check it returns 400?)

internal/server/server_test.go
```diff
@@ -117,23 +117,28 @@ func TestWithoutOsbuildComposerBackend(t *testing.T) {
...
 
-		response, body = tutils.GetResponseBody(t, "http://localhost:8086/api/image-builder/v1/packages?distribution=rhel-8&architecture=x86_64&search=ssh&limit=1", &tutils.AuthString0)
+		response, body = tutils.GetResponseBody(t, "http://localhost:8086/api/image-builder/v1/packages?distribution=rhel-8&architecture=x86_64&search=ssh&limit=10", &tutils.AuthString0)
 		require.Equal(t, 200, response.StatusCode)
 		err = json.Unmarshal([]byte(body), &result)
 		require.NoError(t, err)
 		require.Greater(t, result.Meta.Count, 1)
 		require.Equal(t, result.Data[0], p1)
 
+		response, body = tutils.GetResponseBody(t, "http://localhost:8086/api/image-builder/v1/packages?distribution=rhel-8&architecture=x86_64&search=ssh&limit=-10", &tutils.AuthString0)
```
Gundersanne, line 127: We should test this inded, but how about we just convert the last coupel of cases in this test to test invalid values?

Gundersanne (changes requested):
I like the spec changes! Have some comments on the added tests. 

And maybe we could add a unit test for the `uniqueItems` constraint as well? So for those we could add a test with multiple items, and a test with multiple items with a duplicate item.

internal/server/server_test.go
```diff
@@ -318,8 +318,8 @@ func TestComposeImage(t *testing.T) {
 		}
 		response, body := tutils.PostResponseBody(t, "http://localhost:8086/api/image-builder/v1/compose", payload)
 		require.Equal(t, 400, response.StatusCode)
-		require.Contains(t, body, "image_requests/0/upload_request/options")
-		require.Contains(t, body, "Value is not nullable")
+		require.Regexp(t, "image_requests/0/upload_request/options|image_requests/0/upload_request/type", body)
```
chloenayon, line 321: It seems that the order in which openapi3filter checks for invalid input is not strict - I've seen both of these errors regularly and both are equally technically correct, hence this change from require.Contains to require.Regexp in order to catch both.

Gundersanne (approved):
Thank youuuuuuuuuu

PR #175, opened 2021-04-27
test: Add paragraph on migrating the db when running locally

Gundersanne:


chloenayon (approved):
Tested it locally and it worked, looks good to me!

PR #180, opened 2021-05-12
server: buildCustomizations never returns an error

### LINKED ISSUE #88: buildCustomizations() always returns nil for error
From server.go:

```
    custom, err := buildCustomizations(composeRequest.Customizations)
    if err != nil {
        return err
    }
```

but the current implementation of `buildCustomizations()` function always returns `nil` for error. I'm not sure if it is missing an error condition check or there's simply no way for customizations to error out at the moment. 

Comment from @ondrejbudai:
> hmm... I guess there might be more validation
> I'm not sure how Sanne wants to validate the API requests.

Opening issue so we don't forget about this.

Gundersanne:
Fix #88

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -320,14 +320,9 @@ func (h *Handlers) ComposeImage(ctx echo.Context) error {
 		return err
 	}
 
-	custom, err := buildCustomizations(composeRequest.Customizations)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


atodorov (approved):

ondrejbudai (approved):
Thanks!

PR #186, opened 2021-05-20
prometheus: Use regular counters instead of countervectors

Gundersanne:
Regular counters get initiated to 0, and with countervectors we're
collecting a bunch of data we don't want. Let's target the metrics we
want to collect very specifically, and make our job easier when writing
rules around those metrics.

### DIFF
--- a/internal/server/prometheus.go
+++ b/internal/server/prometheus.go
@@ -1,6 +1,8 @@
 package server
 
 import (
+	"strings"
+
 	"github.com/labstack/echo/v4"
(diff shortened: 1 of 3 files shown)


Gundersanne:
I'll follow this up with a dashboard configmap update after it's merged (that way the metrics become available). I want to have something small but sensible before I get the dashbord into prod, and I find the countervectors are just very hard to work with, or at least aren't suited for what I'm trying to do.

teg (approved):

PR #188, opened 2021-05-20
distributions: Update rhel-8 package list

### LINKED ISSUE #185: Update rhel package lists and distribution names
`dist` is now rhel8.4

Gundersanne:
Fixes #185

blocked on regenerating the pkglist with the most recent version of every pkg


teg (dismissed):
How come you grew 6x?

Gundersanne:
> How come you grew 6x?

That does not make any sense does it. I used the stage cdn this time, and I think the previous list was with the download.devel repos. I'm not sure which is more correct? The stage cdn list includes different package versions too, and not just the latest ones.

Gundersanne:
@teg this diff looks slightly more reasonable :D  I only grew by 4% this time

teg (approved):

PR #189, opened 2021-05-21
dashboards: Adapt to simple counters

Gundersanne:
I also added a variable selecting the range. Followup of #186


teg (approved):
Makes sense! Next time, could you add a comment about how the diff was generated in the commit message? If I understand correctly, it wasn't hand written, right?

Gundersanne:
@teg will do. It's made using the grafana gui, and then exported as json.

PR #190, opened 2021-05-26
Add subscription test case in aws and gcp

yuxisun1217:
Add subscription test case and add environment to CI.

Signed-off-by: Yuxin Sun <yuxisun@redhat.com>

### DIFF
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -145,6 +145,24 @@ function instanceWaitSSH() {
   done
 }
 
+function instanceCheck() {
+  echo "✔️ Instance checking"
+  local _ssh="$1"
+
+  # Check if postgres is installed
+  $_ssh rpm -q postgresql
+
+  # Verify subscribe status
+  subscribe_org_id=$($_ssh sudo subscription-manager identity | grep 'org ID')
+  [[ "$subscribe_org_id" == "org ID: $API_TEST_SUBSCRIPTION_ORG_ID" ]]
+
+  # Verify yum install a small package. It will fail if no available repo.
+  $_ssh sudo dnf -y install dos2unix
+
+  # Unregister subscription
+  $_ssh sudo subscription-manager unregister
+}
+
 ############### AWS-specific functions ################
 
 function checkEnvAWS() {
@@ -185,7 +203,14 @@ function createReqFileAWS() {
   "customizations": {
     "packages": [
       "postgresql"
-    ]
+    ],
+    "subscription": {
+      "organization": ${API_TEST_SUBSCRIPTION_ORG_ID:-},
+      "activation-key": "${API_TEST_SUBSCRIPTION_ACTIVATION_KEY:-}",
+      "base-url": "https://cdn.redhat.com/",
+      "server-url": "subscription.rhsm.redhat.com",
+      "insights": true
+    }
   }
 }
 EOF
@@ -235,7 +260,14 @@ function createReqFileGCP() {
   "customizations": {
     "packages": [
       "postgresql"
-    ]
+    ],
+    "subscription": {
+      "organization": ${API_TEST_SUBSCRIPTION_ORG_ID:-},
+      "activation-key": "${API_TEST_SUBSCRIPTION_ACTIVATION_KEY:-}",
+      "base-url": "https://cdn.redhat.com/",
+      "server-url": "subscription.rhsm.redhat.com",
+      "insights": true
+    }
   }
 }
 EOF
@@ -286,7 +318,14 @@ function createReqFileAzure() {
   "customizations": {
     "packages": [
       "postgresql"
-    ]
+    ],
+    "subscription": {
+      "organization": ${API_TEST_SUBSCRIPTION_ORG_ID:-},
+      "activation-key": "${API_TEST_SUBSCRIPTION_ACTIVATION_KEY:-}",
+      "base-url": "https://cdn.redhat.com/",
+      "server-url": "subscription.rhsm.redhat.com",
+      "insights": true
+    }
   }
 }
 EOF
@@ -412,8 +451,9 @@ function Test_verifyComposeResultAWS() {
   echo "⏱ Waiting for AWS instance to respond to ssh"
   instanceWaitSSH "$HOST"
 
-  # Check if postgres is installed
-  ssh -oStrictHostKeyChecking=no -i ./keypair.pem "$SSH_USER"@"$HOST" rpm -q postgresql
+  # Verify image
+  _ssh="ssh -oStrictHostKeyChecking=no -i ./keypair.pem $SSH_USER@$HOST"
+  instanceCheck "$_ssh"
 }
 
 ### Case: verify the result (image) of a finished compose in GCP
@@ -470,8 +510,9 @@ function Test_verifyComposeResultGCP() {
   echo "⏱ Waiting for GCP instance to respond to ssh"
   instanceWaitSSH "$HOST"
 
-  # Check if postgres is installed
-  ssh -oStrictHostKeyChecking=no -i "$GCP_SSH_KEY" "$SSH_USER"@"$HOST" rpm -q postgresql
+  # Verify image
+  _ssh="ssh -oStrictHostKeyChecking=no -i $GCP_SSH_KEY $SSH_USER@$HOST"
+  instanceCheck "$_ssh"
 }
 
 ### Case: verify the result (image) of a finished compose in Azure
--- a/schutzbot/Jenkinsfile
+++ b/schutzbot/Jenkinsfile
@@ -28,6 +28,10 @@ pipeline {
         }
 
         stage("Test EL8 ✅") {
+            environment {
+                API_TEST_SUBSCRIPTION_ORG_ID = credentials('api-test-subscription-org-id')
+                API_TEST_SUBSCRIPTION_ACTIVATION_KEY = credentials('api-test-subscription-activation-key')
+            }
             parallel {
                 stage('AWS') {
                     agent { label "rhel8cloudbase && x86_64 && aws" }


Gundersanne (approved):

Gundersanne (changes requested):
So on it's own this won't do much. Any reason to not have the changes to https://github.com/osbuild/image-builder/blob/main/test/cases/api.sh in the same PR?

yuxisun1217:
> So on it's own this won't do much. Any reason to not have the changes to https://github.com/osbuild/image-builder/blob/main/test/cases/api.sh in the same PR?

Sorry, I'm still adding the case. Just one question: do we need to cover both "pass" and “not pass" subscription scenarios? If yes, may need to build 2 images in one test. Thanks!

Gundersanne:
> Sorry, I'm still adding the case. Just one question: do we need to cover both "pass" and “not pass" subscription scenarios? If yes, may need to build 2 images in one test. Thanks!

No need to apologize! So I'd just cover the expected code path for subscriptions (so just "pass").

yuxisun1217:
Hi @Gundersanne ，

I've added the case in aws and gcp. For azure, I also add the subscription block in the request file, but no checking case now because of no boot up step in Azure testing.

test/cases/api.sh
```diff
@@ -99,6 +99,15 @@ BASEURLMAJORVERSION="http://$ADDRESS:$PORT/api/image-builder/v1"
 REQUEST_FILE="${WORKDIR}/request.json"
 ARCH=$(uname -m)
 
+RH_SUBSCRIPTION=",
```
Gundersanne, line 102: I'd actually prefer the duplication in each of the createReqFile*() functions rather than using a variable, makes it easier to read the exact request from the code.

test/cases/api.sh
```diff
@@ -145,6 +154,20 @@ function instanceWaitSSH() {
...
 
+function instanceCheck() {
+  local _ssh="$1"
+
+  # Check if postgres is installed
+  $_ssh rpm -q postgresql
+
+  # Verify subscribe status
+  subscribe_status=$($_ssh subscription-manager status | grep 'Overall Status')
+  [[ ! "$subscribe_status" =~ "Unknown" ]]
```
Gundersanne, line 165: Hm is there a string that we could match rather than a negation? "Any string but something with Unknown" seems a bit too loose of a test :)

test/cases/api.sh
```diff
@@ -145,6 +154,20 @@ function instanceWaitSSH() {
...
+
+  # Verify yum install package
+  $_ssh yum -y install dos2unix
```
Gundersanne, line 168: Why not dnf?

Gundersanne (changes requested):
A few questions and nitpicks, but looking good!

test/cases/api.sh
yuxisun1217: Thanks! Use dnf instead.

test/cases/api.sh
yuxisun1217: Updated. Thanks!

test/cases/api.sh
yuxisun1217: Change to use "subscription-manager identify". If it's registered, the output is like:
```
system identity: xxxxxx
name: wala84perftest05060901-vm1
org name: 14651757
org ID: 14651757
```
I grep the "org ID" line and compare it with "org ID: $API_TEST_SUBSCRIPTION_ORG_ID" to verify the subscribe status.

If it is not registered, the output is:
```
This system is not yet registered. Try 'subscription-manager register --help' for more information.
```

yuxisun1217:
> A few questions and nitpicks, but looking good!
> 
> Also a general remark: We also need to unregister the machine, otherwise the 100 machine limit on the subscription will run out real quick I think. So verify that it's subscribed, and then immediately unsusbscribe the machine, hopefully we won't run into the limit all that quickly then.

I unregister the machine in "instanceCheck" function because IMO it only needs to be unregistered if it was subscribed successfully. Do you think it's acceptable? Or should I move it to cleanup function? Thanks! @Gundersanne

yuxisun1217:
Hi @Gundersanne ,
The azure/gcp branches are passed, but aws is failed. Seems in aws the it's not registered successfully:([CI build link](https://osbuildci.cloud.paas.psi.redhat.com/blue/organizations/jenkins/osbuild%2Fimage-builder/detail/PR-190/4/pipeline/))
```
[2021-05-31T03:51:00.116Z] ++ ssh -oStrictHostKeyChecking=no -i ./keypair.pem cloud-user@3.141.27.21 sudo subscription-manager identity
[2021-05-31T03:51:00.116Z] ++ grep 'org ID'
[2021-05-31T03:51:01.498Z] This system is not yet registered. Try 'subscription-manager register --help' for more information.
```

test/cases/api.sh
```diff
@@ -145,6 +145,23 @@ function instanceWaitSSH() {
...
+
+  # Check if postgres is installed
+  $_ssh rpm -q postgresql
+
+  # Verify subscribe status
+  subscribe_org_id=$($_ssh sudo subscription-manager identity | grep 'org ID')
+  [[ "$subscribe_org_id" == "org ID: $API_TEST_SUBSCRIPTION_ORG_ID" ]]
+
+  # Verify yum install a small package
+  $_ssh sudo dnf -y install dos2unix
```
atodorov, line 159: I would suggest that first we check with rpm -q that dos2unix is not installed before attempting to install it with dnf b/c dnf returns 0 if the package is already present.

test/cases/api.sh
```diff
@@ -185,7 +202,14 @@ function createReqFileAWS() {
   "customizations": {
     "packages": [
       "postgresql"
-    ]
+    ],
+    "subscription": {
+      "organization": ${API_TEST_SUBSCRIPTION_ORG_ID:-},
+      "activation-key": "${API_TEST_SUBSCRIPTION_ACTIVATION_KEY:-}",
+      "base-url": "http://cdn.redhat.com/",
```
atodorov, line 209: Is that an `http://` or an `https://` URL ?  IMO should be https

atodorov (changes requested):

test/cases/api.sh
yuxisun1217: Thanks @atodorov ! Updated.

test/cases/api.sh
yuxisun1217: Thanks! Updated. But I've no idea why http also works in gcp...Perhaps both can work? Anyway the default url is indeed https.

test/cases/api.sh
```diff
@@ -145,6 +145,25 @@ function instanceWaitSSH() {
...
+
+  # Check if postgres is installed
+  $_ssh rpm -q postgresql
+
+  # Verify subscribe status
+  subscribe_org_id=$($_ssh sudo subscription-manager identity | grep 'org ID')
+  [[ "$subscribe_org_id" == "org ID: $API_TEST_SUBSCRIPTION_ORG_ID" ]]
+
+  # Remove dos2unix before install it
+  $_ssh rpm -q dos2unix && $_ssh sudo rpm -e dos2unix
```
atodorov, line 159: IMO you don't need to remove the rpm b/c if `rpm -q` returns non-zero the second part will not be executed anyway.  On top of that we have `set -e` enabled which should cause the entire script to exit with non-zero immediately.

atodorov (changes requested):

test/cases/api.sh
yuxisun1217: Hi @atodorov ,
I've tried it in my fedora33 that the "&&" can catch the previous exit code. Here is the logic:
if dos2unix is pre-installed, "rpm -q dos2unix" return 0, run "rpm -e dos2unix"
else if dos2unix is not pre-installed, "rpm -q dos2unix" return 1, then skip "rpm -e dos2unix" step and run "dnf -y install dos2unix"...
The only risk here I think is if "rpm -e dos2unix" failed the script will exit. But I think it's very rare to happen. And if the package was installed but failed to be removed, the next step "dnf install" will not be able to verify the package installation. 
Could you please double check that? Thanks!

atodorov: According to man `The shell does not exit if the command that fails is ... part of  any  command  exe‐cuted  in  a  &&  or || list except the command following the final && or ||,` so it looks like you are correct and I was wrong. 

rpm -e can indeed fail if there are other packages requiring dos2unix. But in this case we control what gets installed onto the system and which packages we use for testing purposes. It is unlikely that would fail and it is probably equally unlikely that we'll have to remove the package before installing it again. I don't have a strong opinion in any direction. 

@Gundersanne what do you think about  checking & removing the package before installing it vs. only checking that it's not installed and failing the entire script if it is ?

test/cases/api.sh
```diff
@@ -145,6 +145,25 @@ function instanceWaitSSH() {
...
+  # Verify yum install a small package
+  $_ssh sudo dnf -y install dos2unix
+
+  # Unregister subscription
+  $_ssh sudo subscription-manager unregister || true
```
Gundersanne, line 164: Does this fail often? Why the `|| true`?

test/cases/api.sh
Gundersanne: Hm are we expecting the package to sometimes be installed? Or can we just pick a package that's never installed?

Plus wouldn't `dnf install $installed_package` fail anyway if the system isn't registered?

test/cases/api.sh
```diff
@@ -470,8 +512,9 @@ function Test_verifyComposeResultGCP() {
   echo "⏱ Waiting for GCP instance to respond to ssh"
   instanceWaitSSH "$HOST"
 
-  # Check if postgres is installed
-  ssh -oStrictHostKeyChecking=no -i "$GCP_SSH_KEY" "$SSH_USER"@"$HOST" rpm -q postgresql
+  # Verify image
+  _ssh="ssh -oStrictHostKeyChecking=no -i $GCP_SSH_KEY $SSH_USER@$HOST"
```
Gundersanne, line 516: Since yo uadded the `echo "✔️ Instance checking"` in the AWS check, might as well add it here :)

Gundersanne (commented):
Looking good! I do have some additional questions.

test/cases/api.sh
yuxisun1217: I've never met the failed case. Just to avoid the risk because I think it's not a checkpoint and if it's failed it should not make the whole test failed. Do you suggest to remove the || true? Thanks!

atodorov: "Need it robust, make it fragile" -> https://www.youtube.com/watch?v=WOy9zhzyMOE

IMO let this fail the test if unregister fails. Otherwise we will exhaust the pool of subscriptions and the subsequent tests will still fail but be a bit harder to debug.

test/cases/api.sh
yuxisun1217: Ah actually yes I think. If there's no repo, the yum install returns 1 no matter if the package is installed:
```
# yum install dos2unix
Error: There are no enabled repositories in "/etc/yum.repos.d", "/etc/yum/repos.d", "/etc/distro.repos.d".
# echo $?
1
```
So I think I should remove the package check before dnf install?

test/cases/api.sh
yuxisun1217: Thanks @Gundersanne ! Move it into the function to reduce duplicate lines.

test/cases/api.sh
yuxisun1217: OK thanks @atodorov ! Removed it.

test/cases/api.sh
yuxisun1217: Just removed the package pre-check. Please let me know if it's not a good solution. Thanks!

atodorov: Once we migrate to GitLab the runners will come with some epositories pre-configured sometimes.

If you want this to be a robust check then remove/disable  everything under /etc/yum.repos.d before subscribing but still keep the `rpm -q dos2unix` check.  That's a pre-condition and we want to fail if the pre-condition isn't met.

Gundersanne (approved):
nice :) Thank you!

test/cases/api.sh
yuxisun1217: Thanks @atodorov ! Since this PR is merged I opened a new issue #195 to track this, and will fix it soon.

PR #193, opened 2021-06-01
Build Edge commit and installer and upload to S3

achilleas-k:
Counterpart to https://github.com/osbuild/osbuild-composer/pull/1439

This PR adds a new upload type labelled `aws.s3` for uploading artifacts to an S3 bucket.

## EDIT

There are some things I'd like to include before marking as ready:
- [x] OSTree parameters
- [x] Edge installer: COMPOSER-926
- [x] Tests

The edge installer would receive the same treatment as the edge commit, but it will need to pull the edge commit payload to embed it.


teg (commented):
This is great! Thanks!

distributions/rhel-85.json
```diff
@@ -0,0 +1,15 @@
+{
+  "module_platform_id": "platform:el8",
+  "distribution": {
+    "name": "rhel-85",
+    "description": "Red Hat Enterprise Linux (RHEL) 8"
+  },
+  "x86_64": {
+    "image_types": [ "ami", "vhd", "edge-commit" ],
```
teg, line 8: I think this should just be added as part of 8.4 instead, no?

teg (changes requested):
I think if we change the patch to add 8.5 support to instead add edge support to 8.4, this can be landed (as soon as the composer counter-part).

distributions/rhel-85.json
Gundersanne: if we just want edge support in GA then we can just add it to the current rhel-8 file

teg: GA?

achilleas-k:
Added `rhel-edge-commit` to the RHEL 8.4 distribution config.

teg (dismissed):
This is really great! Could we hook up some basic test too before merging it?

achilleas-k:
I've added basic unit tests.  Tested building a `rhel-edge-installer` locally and no further changes are needed other than including it in the distribution file.

teg:
I think a fixup commit is still left over. @Gundersanne do we do full build/upload tests for the other image types yet, do we need the same here?

achilleas-k:
oops, sorry for the fixup commit leftover.  Just noticed it now while adding the installer to the distro config.
Fixed.

Gundersanne:
> I think a fixup commit is still left over. @Gundersanne do we do full build/upload tests for the other image types yet, do we need the same here?

Yea all the images are built/uploaded (rhel only though). I'm ok with it being added in a followup however.

Gundersanne (approved):
This looks fine. We probably want to include an integration test? But I'm fine with doing that in a followup.

Gundersanne:
Added a blocked label for osbuild/osbuild-composer#1439 ; as soon as that's in we can redeploy composer and merge this!

achilleas-k:
Moved the edge image types from `rhel-8.json` to `rhel-84.json` (see #202)

achilleas-k:
Changed the location of the ostree options in the request payload.  See also https://github.com/osbuild/osbuild-composer/pull/1439#issuecomment-863157741.

teg:
I think we are very, very close to having the composer PR in, this will need a rebase though.

achilleas-k:
Unblocked and rebased.

teg (approved):
Thanks!

PR #198, opened 2021-06-09
Add Azure boot up test and clean up Azure resources after test

yuxisun1217:
1. Boot up Azure VM and check image
2. Delete os_disk, NIC, public-ip, nsg, vnet, image and blob in cleanupAzure.

Signed-off-by: Yuxin Sun <yuxisun@redhat.com>

### DIFF
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -52,12 +52,30 @@ function cleanupAzure() {
 
   # do not run clean-up if the image name is not yet defined
   if [[ -n "$AZURE_CMD" && -n "$AZURE_IMAGE_NAME" ]]; then
+    # When clean up, need to delete os_disk, NIC, public-ip, nsg, vnet, image and blob.
     set +e
-    $AZURE_CMD image delete --resource-group sharing-research --name "$AZURE_IMAGE_NAME"
+    $AZURE_CMD image delete --resource-group "$AZURE_RESOURCE_GROUP" --name "$AZURE_IMAGE_NAME"
+    # Get all the resources ids
+    VM_ID=$(jq -r '.id' "$WORKDIR"/vm_details.json)
+    OSDISK_ID=$(jq -r '.storageProfile.osDisk.managedDisk.id' "$WORKDIR"/vm_details.json)
+    NIC_ID=$(jq -r '.networkProfile.networkInterfaces[0].id' "$WORKDIR"/vm_details.json)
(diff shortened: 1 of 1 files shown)


yuxisun1217:
The Azure test has passed. AWS stage failed because of #199 .

test/cases/api.sh
```diff
@@ -52,12 +52,28 @@ function cleanupAzure() {
...
+    NIC_ID=$(jq -r '.networkProfile.networkInterfaces[0].id' "$WORKDIR"/vm_details.json)
+    "$AZURE_CMD" network nic show --ids "$NIC_ID" > "$WORKDIR"/nic_details.json
+    NSG_ID=$(jq -r '.networkSecurityGroup.id' "$WORKDIR"/nic_details.json)
+    PUBLICIP_ID=$(jq -r '.ipConfigurations[0].publicIpAddress.id' "$WORKDIR"/nic_details.json)
+    VNET_ID=$(jq -r '.ipConfigurations[0].subnet.id' "$WORKDIR"/nic_details.json | cut -d '/' -f1-9)
+
+    # Delete resources one by one
+    for id in "$VM_ID" "$OSDISK_ID" "$NIC_ID" "$NSG_ID" "$PUBLICIP_ID" "$VNET_ID"; do
+      echo "Deleting $id..."
+      "$AZURE_CMD" resource delete --ids "$id"
```
atodorov, line 70: Does the order in which resources are deleted matter?  If yes its a good idea to document that with a comment.

test/cases/api.sh
```diff
@@ -533,6 +549,31 @@ function Test_verifyComposeResultAzure() {
   # https://github.com/Azure/azure-cli/issues/17123
   # Without this issue fixed or worked around, I'm not able to delete the disk
   # attached to the VM.
+
```
atodorov, line 552: Remove commented section above.

atodorov (changes requested):

test/cases/api.sh
yuxisun1217: Yes. I'll add comment. Besides, I just aware that the vnet will be reused by other VMs in the same resource group so I change to not delete it. Due to MSFT's document "Virtual Network in Azure is free of charge.", so it won't be charged if no traffic.
Just updated. Please help to review again. Thanks!

atodorov (approved):

atodorov (approved):

PR #202, opened 2021-06-16
distributions: Duplicate rhel-8 into rhel-84

Gundersanne:


achilleas-k (approved):
LGTM.
I'll update #193 as soon as this is merged.

ondrejbudai (approved):
LGTM. Should we regenerate the package list? There might be a lot of differences between 8.3 and 8.4. Can be done in a follow-up.

PR #203, opened 2021-06-17
server: add rhel-84 support to api packages route

jkozol:
When fetching packages the user should be able to specify both rhel-8 and rhel-84 as the distribution. The packages server test is updated to use the rhel-84 distribution since this is what the image-builder-frontend currently uses as its rhel distro.


jkozol:
This should fix https://github.com/RedHatInsights/image-builder-frontend/issues/242

teg (approved):

jrusz:
Just FYI, I came across this and I restarted the Azure test, it failed to download azure-cli, I've seen this before, it just sometimes fails to do that and the next try is usually fine... I don't know why it happens but I think it would be worth adding some retries to all of our tests that use azure-cli...

ondrejbudai:
@jrusz can you please create an issue from the failed Azure test? Just a snippet from the log and link to the failed job. :)

Gundersanne (approved):
I might have broken this yesterday :) Only on stage though, release the frontend is blocked until this is released.

Gundersanne:
Released @jkozol , should be fixed on stage!

PR #207, opened 2021-06-23
Added ansible to the rhel 8.4 distribution files.

AaronH88:
Added Ansible to the list of available packages for anyone building rhel 8.4 images.

Jira ticket: https://issues.redhat.com/browse/COMPOSER-981 

In order to keep CentOS 8 Streams as close as possible to rhel, ansibile has also been added to CentOS. 

TODO: 

- [x]  Test

### DIFF
--- a/distributions/rhel-84.json
+++ b/distributions/rhel-84.json
@@ -12,9 +12,9 @@
     }, {
       "baseurl": "https://cdn.redhat.com/content/dist/rhel8/8/x86_64/appstream/os",
       "rhsm": true
-    }, {                                                                                                                                                      
-      "baseurl": "https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/ansible/2.9/os",                                                                   
-      "rhsm": true                                                                                                                                            
+    }, {
+      "baseurl": "https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/ansible/2/os",
+      "rhsm": true
     }]
   }
 }
--- a/distributions/rhel-84.json
+++ b/distributions/rhel-84.json
@@ -12,6 +12,9 @@
     }, {
       "baseurl": "https://cdn.redhat.com/content/dist/rhel8/8/x86_64/appstream/os",
       "rhsm": true
+    }, {                                                                                                                                                      
+      "baseurl": "https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/ansible/2.9/os",                                                                   
+      "rhsm": true                                                                                                                                            
     }]
   }
 }
--- a/distributions/centos-8-x86_64-packages.json
+++ b/distributions/centos-8-x86_64-packages.json
@@ -18,72 +18,72 @@
     {
       "summary": "Network connection manager and user applications",
       "name": "NetworkManager",
-      "version": "1.30.0"
+      "version": "1.32.0"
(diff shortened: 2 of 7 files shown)


distributions/rhel-84.json
```diff
@@ -12,6 +12,9 @@
     }, {
       "baseurl": "https://cdn.redhat.com/content/dist/rhel8/8/x86_64/appstream/os",
       "rhsm": true
+    }, {                                                                                                                                                      
+      "baseurl": "https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/ansible/2.9/os",                                                                   
+      "rhsm": true
```
Gundersanne, line 17: How about we add a field here `"requires_subscription": [ 'SKUxxxxx', 'ansible', etc... ]`. If present image-builder will check if the logged in user has access to these, and adds the packages (which should be located in a separate file), when querying the `/packages` endpoint. That way the user can search for/select them in the UI.

 The separate packages file should have a predictable name I think. Something like

`rhel-84-x86_64-packages.$SKU.json` with $SKU corresponding to the subscription name.

I'm not sure if we'll use SKU numbers or a name (for example `ansible`), whatever is easiest, both should be treated as strings anyway.

Gundersanne (dismissed):
So by adding it to the main rhel84 packages file the result is that the ansible packages will seem to be available to anyone wanting to build the rhel-84 image. I don't think that's the intention?

If I understood it correctly, the aim is to add this repository and packages depending on whether or not the user has a valid ansible subscription.

So this will require code changes as well, to `internal/server/distribution.go` specifically.

distributions/rhel-84.json
Gundersanne, line 17: Hmm when I enable it on a rhel-8.4 vm (`subscription-manager repos --enable ansible-2-for-rhel-8-x86_64-rpms`) I get:

```Repo ID:   ansible-2-for-rhel-8-x86_64-rpms
Repo Name: Red Hat Ansible Engine 2 for RHEL 8 x86_64 (RPMs)
Repo URL:  https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/ansible/2/os
Enabled:   1
```

Is there a reason to pin it to `2.9?`

Gundersanne (changes requested):
Thank you! Dismissed my earlier review as the requirements changed.

I have one question, and this will need a rebase to make the tests work  (due to https://github.com/osbuild/image-builder/commit/924a2d01b74609435569b45af2c8fc0698368cd4).

teg:
> So by adding it to the main rhel84 packages file the result is that the ansible packages will seem to be available to anyone wanting to build the rhel-84 image. I don't think that's the intention?

Turns out this was a misunderstanding on my part. ansible-2.9 is available to anyone with a RHEL subscription.

@AaronH88 should we enable the same for CentOS Stream 8 though, do the equivalent ansible repos exist there?

teg:
(C8S could be a separate PR, of course)

teg (approved):
Looking good!

distributions/rhel-84.json
AaronH88: Agreed and changed

teg (dismissed):

Gundersanne (approved):

PR #208, opened 2021-06-24
templates: Update dashboard to match alerts

Gundersanne:
image

---

These are generated so the diff doesn't matter all that much.


Gundersanne:
@teg I changed the dashboard (again) to match the alerts that got merged. That way we can refer to them from the alerts.

teg (dismissed):

teg (approved):

PR #213, opened 2021-07-05
test edge commit image

yih-redhat:
This PR contains:

1. Disabled subscription for rhel8.4 runners
2. Added test script to test edge commit image

@atodorov Can you please help to review this one? Appreciate if you can help to start to run this PR, it passed locally but it's the first time to run in image-builder CI, I need to run it early and fix the environment problems.

PR #217, opened 2021-07-13
tests: remove workarounds for non-subscribed RHEL

ondrejbudai:
#213 introduced some workarounds for non-subscribed image. They weren't
in the end needed in the PR because we decided to use a subscribed image.
This commit removes them.

This is entirely my fault because I gave @yih-redhat bad suggestions without
fully understanding the intent of his PR. Therefore I'm cleaning my own
mess.

BTW it would be cool if we can somehow test the service on future RHEL versions - CentOS Stream sounds like a good option for this kind of test.


yih-redhat:
Hi Ondrej,
Glad to know that we will use subscribed images, and thanks for your help that I can have my test case running now. I can continue to add more test cases, which is important for me to make sure image-builder can generate edge images correctly.

I don't think they are bad suggestions, although we need to remove the workaround if we use subscribed images, it helped me to continue my work, thanks.

Gundersanne (approved):

PR #218, opened 2021-07-14
add test case for edge iso image

yih-redhat:
This pr contains:

Test edge installer iso image, includes:
- call rest api to generate iso image, 
- use mkksiso to put ks file into iso, 
- install edge with iso file
- run ansible scirpt to check edge vm.

### DIFF
--- a/image-builder.spec
+++ b/image-builder.spec
@@ -103,6 +103,7 @@ Requires:   qemu-kvm
 Requires:   libvirt-client
 Requires:   libvirt-daemon-kvm
 Requires:   virt-install
+Requires:   wget
 
 %description tests
 Integration tests to be run on a system with image-builder installed.
--- a/test/cases/edge.sh
+++ b/test/cases/edge.sh
@@ -9,7 +9,11 @@
 # 2. Download commit image, extract it to /var/www/html, and serve it over httpd
 # 3. Install Edge vm based on commit repo url, and run ansible playbook to check it.
 #
-# Test case for installer ISO image will be added soon.
+# ------------ Test iso image --------------------------------------------------------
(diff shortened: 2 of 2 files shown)


yih-redhat:
CI runners has been changed to subscribed image, see #217. The gitlab CI job of this pull request will fail until #217 passed.

teg (dismissed):
Thanks for this! Looking great, exactly what we need :)

Gundersanne (approved):
thanks! I'll wait for tests to be green and merge :)

Gundersanne:
This is really not ideal 
image

I've retried the test a couple of times now and it seems really flaky, and hard to debug because the log output. I'll retry a few more times but this really should be fixed, so if we merge this an issue should be created and followed up on.

yih-redhat:
> This is really not ideal
> image
> 
> I've retried the test a couple of times now and it seems really flaky, and hard to debug because of the log output. I'll retry a few more times but this really should be fixed, so if we merge this an issue should be created and followed up on.

Hi Sanne,

I have seen this several times before, the root reason is that my test script contains a step to display the metadata of image, which produces lots of log.
I have disabled that step, until we can increase the log size of CI job. Now we can check the failures if job failed.

Pushed and trigger the test again.


Thanks,
Yi

ondrejbudai:
If you need to store bigger logs, just use Gitlab's artifacts: https://docs.gitlab.com/ee/ci/pipelines/job_artifacts.html

yih-redhat:
@Gundersanne Can you please help to merge this pr?

yih-redhat:
> If you need to store bigger logs, just use Gitlab's artifacts: https://docs.gitlab.com/ee/ci/pipelines/job_artifacts.html

Thanks, Ondrej. I will try it later.

Gundersanne (approved):

PR #219, opened 2021-07-15
image-types/rhel84: add rhel-edge-container

teg:
This is an alternative to rhel-edge-commit, which has the same payload, but in 8.4 it
supports embedding ssh keys, which rhel-edge-commit does not. The Edge API would
like to use this, so add the image type to the allow-list.

cc @aaronh88

### DIFF
--- a/distributions/rhel-84.json
+++ b/distributions/rhel-84.json
@@ -5,7 +5,7 @@
     "description": "Red Hat Enterprise Linux (RHEL) 8"
   },
   "x86_64": {
-    "image_types": [ "ami", "vhd", "rhel-edge-commit", "rhel-edge-installer" ],
+    "image_types": [ "ami", "vhd", "rhel-edge-commit", "rhel-edge-installer", "rhel-edge-container" ],
     "repositories": [{
       "baseurl": "https://cdn.redhat.com/content/dist/rhel8/8/x86_64/baseos/os",
       "rhsm": true


yih-redhat:
@teg Hi Tom, I think this pr is just to add the image type to the allow-list, and there will be follow up tasks to implement the Rest API in image-builder to build container image, right?

Gundersanne (approved):

teg:
> @teg Hi Tom, I think this pr is just to add the image type to the allow-list, and there will be follow up tasks to implement the Rest API in image-builder to build container image, right?

This adds it to the allow-list, and as it already is implemented in composer that should be all we need for container building to work. Does that answer your question, or did I misunderstand?

yih-redhat:
> > @teg Hi Tom, I think this pr is just to add the image type to the allow-list, and there will be follow up tasks to implement the Rest API in image-builder to build container image, right?
> 
> This adds it to the allow-list, and as it already is implemented in composer that should be all we need for container building to work. Does that answer your question, or did I misunderstand?

Hi @teg ,

Sorry if I didn't make myself clear.

Yes, I know that osbuild-composer can build commit, container and iso. And as I know, image-builder can build commit and iso, which were implemented by Achilleas, so actually my question is:
1. Is there any extra work need to be done in image-builder to build container image?
2. Is there any clear requirement from edge team that they need to use container image? cause I didn't see they use container in last edge demo meeting.

 Thanks a lot if you can answer these questions :) , I may need to add test cases for container in image-builder.

Thanks,
Yi

teg:
Hi @yih-redhat!
I'm sorry, I missed your question. In case it is still relevant:
1. No, this is all.
2. This turned out not to be needed by them, as they found another work-around.
I would not put any more work into this as it is now, as I don't think it is used. We may want to revisit this when we can actually push these artefacts to container registries, then they would bring real benefit over the commit tarballs.

PR #223, opened 2021-07-20
dnf-json: Fetch from osbuild-composer when needed

lavocatt:
dnf-json is fetched from
https://raw.githubusercontent.com/osbuild/osbuild-composer/main/dnf-json
and is no more versioned in this repo.


tools/generate-package-lists
```diff
@@ -8,9 +8,13 @@ import os
 import subprocess
 import sys
 import time
+import requests
 
 PROJECT_ROOT=os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
 DISTRIBUTION_DIR=os.path.join(PROJECT_ROOT, 'distributions')
+r = requests.get("https://raw.githubusercontent.com/osbuild/osbuild-composer/main/dnf-json")
```
Gundersanne, line 15: I'd move the actual request into the main function, under the argument parsing. On top it's just the definition of the constants.

tools/generate-package-lists
```diff
@@ -8,9 +8,13 @@ import os
...
+with open(os.path.join(PROJECT_ROOT, "dnf-json"), 'w') as f:
```
Gundersanne, line 16: When moved below, could just use the DNF_JSON constant instead of the path.join.

Gundersanne (changes requested):
Thanks!

`dnf-json` needs to be marked as executable as well. After writing the file add `os.chmod(DNF_JSON, 0o744)`.  You can test quite easily using `tools/generate-package-lists --distro ./distributions/centos-8.json` or another distro.

lavocatt:
> Thanks!
> 
> `dnf-json` needs to be marked as executable as well. After writing the file add `os.chmod(DNF_JSON, 0o744)`. You can test quite easily using `tools/generate-package-lists --distro ./distributions/centos-8.json` or another distro.

Indeed ! I was testing it with `./tools/generate-package-lists` only which obviously didn't thrown any bad exception.

Gundersanne (approved):
Thanks!

PR #227, opened 2021-08-07
remove duplicate packages in rhel-84-x86_64-packages.json

matoval:
Remove duplicate packages in rhel-84-x86_64-packages.json to resolve issue with receiving duplicate packages as described in this [ticket](https://issues.redhat.com/browse/RHELPLAN-90237).

### DIFF
--- a/distributions/rhel-84-x86_64-packages.json
+++ b/distributions/rhel-84-x86_64-packages.json
@@ -126,11 +126,6 @@
       "version": "1.56.1"
     },
     {
-      "summary": "The NetBSD Editline library",
-      "name": "libedit",
-      "version": "3.1"
-    },
-    {
       "summary": "Libsmbios C/C++ shared libraries",
       "name": "libsmbios",
       "version": "2.4.1"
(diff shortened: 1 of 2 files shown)


Gundersanne (changes requested):
Thank you for working on this!

I'd prefer this to be fixed in [tools/generate-package-lists](https://github.com/osbuild/image-builder/blob/main/tools/generate-package-lists)  though. I suggest to drop the `version` from the data, and then just filter out duplicate names before writing it all to a file.

matoval:
> Thank you for working on this!
> 
> I'd prefer this to be fixed in [tools/generate-package-lists](https://github.com/osbuild/image-builder/blob/main/tools/generate-package-lists)  though. I suggest to drop the `version` from the data, and then just filter out duplicate names before writing it all to a file.

Sounds good, I'll make those changes and push a new commit.

matoval:
@Gundersanne I added the functionality to remove duplicate packages when generating the package list. Currently I left version in the package object and when generating a list there would be multiple versions of the same package.

Gundersanne:
I integrated your commits in https://github.com/osbuild/image-builder/pull/230 ; there's some extra work I wanna base off of this. I hope that's ok. 

Mostly I want to drop the versioning field entirely, as we don't allow specifying exact versions for packages anyway, you always get the latest version. So I'm closing this, but it's included in #230.

PR #228, opened 2021-08-11
tests: remove hard coded ssh keys

jabia99:
Use newly generated ssh keys instead of hard codded ones


jabia99:
> Looks good, is this ready to be merged?

Yes, the keys are not used anywhere else

Gundersanne (approved):
Nice thanks!

PR #230, opened 2021-08-18
Remove duplication from packages

Gundersanne:


tools/generate-package-lists
```diff
@@ -70,6 +70,7 @@ def main():
             pkgs = [ {"name": p["name"], "summary": p["summary"]} for p in json.loads(out)['packages'] ]
             filtered = []
             [ filtered.append(p) for p in pkgs if p["name"] not in list(map(lambda p: p["name"], filtered)) ]
+            filtered.sort(key=lambda p: p["name"].lower())
```
Gundersanne, line 73: I know they're all lowercase, but to be triple sure :)

teg:
Looking good, but does this break backwards compatibility in any way? Will we fail if a caller passes a version field?

Gundersanne:
> Looking good, but does this break backwards compatibility in any way? Will we fail if a caller passes a version field?

We never required a version field to specify packages, just the package name.

This breaks compatibility a little if the caller depended on the version field being present when querying /packages, none of the current users (IBFE and edge) do.  And honestly i've kind of given up on backwards compatibility until we reach public beta .-.

teg (approved):
If you checked the only two real users, then I think we are good

PR #241, opened 2021-08-31
Compose: No cross user compose access

lavocatt:
Before accessing a compose status, the value of the identity header is
checked against the database to ensure that the user getting the
information is the same that made the request.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -52,7 +52,7 @@ func (db *dB) InsertCompose(jobId, accountId, orgId string, request json.RawMess
 	}
 	defer conn.Release()
 
-	_, err = conn.Query(ctx, "INSERT INTO composes(job_id, request, created_at, account_id, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4);", jobId, request, accountId, orgId)
+	_, err = conn.Exec(ctx, "INSERT INTO composes(job_id, request, created_at, account_id, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4)", jobId, request, accountId, orgId)
 	return err
 }
 
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -3,11 +3,16 @@ package db
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
 )
+
+// ComposeNotFoundError occurs when no compose request is found for a user.
+var ComposeNotFoundError = errors.New("Compose not found")
 
 type dB struct {
 	Pool *pgxpool.Pool
@@ -22,6 +27,7 @@ type ComposeEntry struct {
 type DB interface {
 	InsertCompose(jobId, accountId, orgId string, request json.RawMessage) error
 	GetComposes(accountId string, limit, offset int) ([]ComposeEntry, int, error)
+	GetCompose(jobId string, accountId string) (*ComposeEntry, error)
 }
 
 func InitDBConnectionPool(connStr string) (DB, error) {
@@ -48,6 +54,30 @@ func (db *dB) InsertCompose(jobId, accountId, orgId string, request json.RawMess
 
 	_, err = conn.Query(ctx, "INSERT INTO composes(job_id, request, created_at, account_id, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4);", jobId, request, accountId, orgId)
 	return err
+}
+
+func (db *dB) GetCompose(jobId string, accountId string) (*ComposeEntry, error) {
+	ctx := context.Background()
+	conn, err := db.Pool.Acquire(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Release()
+
+	result := conn.QueryRow(ctx, "SELECT job_id, request, created_at FROM composes WHERE account_id=$1 and job_id=$2",
+		accountId, jobId)
+
+	var compose ComposeEntry
+	err = result.Scan(&compose.Id, &compose.Request, &compose.CreatedAt)
+	if err != nil {
+		if errors.As(err, &pgx.ErrNoRows) {
+			return nil, ComposeNotFoundError
+		} else {
+			return nil, err
+		}
+	}
+
+	return &compose, nil
 }
 
 func (db *dB) GetComposes(accountId string, limit, offset int) ([]ComposeEntry, int, error) {
--- a/internal/tutils/tutils.go
+++ b/internal/tutils/tutils.go
@@ -75,11 +75,13 @@ func PostResponseBody(t *testing.T, url string, compose interface{}) (*http.Resp
 }
 
 type dB struct {
-	entries []db.ComposeEntry
+	accountOwernship map[string][]db.ComposeEntry
 }
 
 func InitDB() db.DB {
-	return &dB{[]db.ComposeEntry{}}
+	return &dB{
+		make(map[string][]db.ComposeEntry),
+	}
 }
 
 func (d *dB) InsertCompose(jobId, accountId, orgId string, request json.RawMessage) error {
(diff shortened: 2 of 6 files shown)


internal/db/db.go
```diff
@@ -22,6 +22,7 @@ type ComposeEntry struct {
 type DB interface {
 	InsertCompose(jobId, accountId, orgId string, request json.RawMessage) error
 	GetComposes(accountId string, limit, offset int) ([]ComposeEntry, int, error)
+	DoesComposesBelongsToAccount(accountId string, composeId string) (bool, error)
```
Gundersanne, line 25: How about `GetCompose (accountId string, composeId uuid) (*ComposeEntry, error)` ? Less logic in the db that way, and might be useful in future. 

And then define an error which can be used to distinguish between 'real errors' and not found.

```
type NoRowsError struct {}

func (err *NoRowsError) Error() string {
    return "No rows found"
}
``` 
Also see https://go.dev/blog/go1.13-errors.

internal/db/db.go
```diff
@@ -50,6 +51,30 @@ func (db *dB) InsertCompose(jobId, accountId, orgId string, request json.RawMess
...
+	var count int
+	err = conn.QueryRow(ctx,
+		"SELECT COUNT(*) FROM composes WHERE account_id=$1 AND job_id=$2",
+		accountId,
+		composeId).Scan(&count)
+	if err != nil {
+		return false, err
+	} else {
+		return count > 0, nil
+	}
```
Gundersanne, line 75: With `GetCompose` you can just scan the entire row into a ComposeEntry. And `.Scan` will return `pgx.ErrNoRows` which can easily be checked like `err != nil && errors.As(err, &pgx.ErrNoRows)` in case no row was found. In that case `GetCompose` would return `nil, NoRowsError`.

Gundersanne (changes requested):
Thank you! A few thoughts.

I think it might be nice to add a unit test in `server_test.go` as well to have it covered for faster local dev (even if it runs against the mock DB).

Also currently we show past composes purely based on account number, but it might be worth being able to list organization composes too (so based on org_id). Unsure yet however, also not in the scope of this PR.

Also opened #242 rereading the db code ^_^

internal/db/db.go
```diff
@@ -50,6 +51,30 @@ func (db *dB) InsertCompose(jobId, accountId, orgId string, request json.RawMess
 	return err
 }
 
+// Returns (true, nil) if the composeId belongs to the accountId (false, nil)
+// otherwise.
+// Returns (false, err) when an error occurs
+func (db *dB) DoesComposesBelongsToAccount(accountId string, composeId string) (
```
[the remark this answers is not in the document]
lavocatt: Alright then, I'm gonna make a special vim configuration. I tend to prefer to have 80 cols because with this I can have more vertical splits, and the more I have, the better I feel ^^

internal/v1/server.go
```diff
@@ -201,7 +201,27 @@ func (h *Handlers) GetArchitectures(ctx echo.Context, distribution string) error
...
 
+func (h *Handlers) composeBelongsToId(ctx echo.Context, composeId string) (bool, error) {
+	ih := ctx.Get("IdentityHeader")
+	if ih == nil {
+		return false, echo.NewHTTPError(http.StatusInternalServerError, "Identity Header missing in request handler")
+	}
+	idHeader, ok := ih.(IdentityHeader)
+	if !ok {
+		return false, echo.NewHTTPError(http.StatusInternalServerError, "Identity Header invalid in request handler")
+	}
```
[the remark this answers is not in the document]
lavocatt: Got it, makes sense.

internal/v1/server.go
```diff
@@ -201,7 +201,27 @@ func (h *Handlers) GetArchitectures(ctx echo.Context, distribution string) error
...
+
+	return h.server.db.DoesComposesBelongsToAccount(idHeader.Identity.AccountNumber, composeId)
+}
+
 func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId string) error {
+	belongs, err := h.composeBelongsToId(ctx, composeId)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, err)
```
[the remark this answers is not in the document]
lavocatt: Good catch!

test/cases/api.sh
```diff
@@ -105,12 +105,16 @@ trap cleanup EXIT
...
 
+# org_id 000001 (valid org_id)
+VALIDAUTHSTRING2="eyJlbnRpdGxlbWVudHMiOnsiaW5zaWdodHMiOnsiaXNfZW50aXRsZWQiOnRydWV9LCAic21hcnRfbWFuYWdlbWVudCI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sICJvcGVuc2hpZnQiOnsiaXNfZW50aXRsZWQiOnRydWV9LCAiaHlicmlkIjp7ImlzX2VudGl0bGVkIjp0cnVlfSwgIm1pZ3JhdGlvbnMiOnsiaXNfZW50aXRsZWQiOnRydWV9LCAiYW5zaWJsZSI6eyJpc19lbnRpdGxlZCI6dHJ1ZX19LCAiaWRlbnRpdHkiOnsiYWNjb3VudF9udW1iZXIiOiIwMDAwMDEiLCAidHlwZSI6IlVzZXIiLCAidXNlciI6eyJ1c2VybmFtZSI6InVzZXIiLCAiZW1haWwiOiJ1c2VyQHVzZXIudXNlciIsICJmaXJzdF9uYW1lIjoidXNlciIsICJsYXN0X25hbWUiOiJ1c2VyIiwgImlzX2FjdGl2ZSI6dHJ1ZSwgImlzX29yZ19hZG1pbiI6dHJ1ZSwgImlzX2ludGVybmFsIjp0cnVlLCJsb2NhbGUiOiJlbi1VUyJ9LCAiaW50ZXJuYWwiOnsib3JnX2lkIjoiMDAwMDAxIn19fQo="
+
```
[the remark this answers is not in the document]
lavocatt: Oh, yeah! Didn't see this coming really. I was testing on my local machine and the `VALIDAUTHSTRING2` was working correctly. Well I'll do what you propose, seems nice.

internal/db/db.go
lavocatt: I was quite happy with the logic right there, but you proposition is nice too.

internal/db/db.go
lavocatt: Oh yes ! I'll do that too. Thanks.

internal/db/db.go
Gundersanne: I'm not entirely against having a specific function which just checks if there was an element found, but in that case `ComposeBelongsToAccount` or `DoesComposeBelongToAccount` is correcter.

test/cases/api.sh
```diff
@@ -105,12 +105,16 @@ trap cleanup EXIT
...
 # org_id 000001 (invalid org_id)
 INVALIDAUTHSTRING="eyJlbnRpdGxlbWVudHMiOnsiaW5zaWdodHMiOnsiaXNfZW50aXRsZWQiOnRydWV9LCJzbWFydF9tYW5hZ2VtZW50Ijp7ImlzX2VudGl0bGVkIjp0cnVlfSwib3BlbnNoaWZ0Ijp7ImlzX2VudGl0bGVkIjp0cnVlfSwiaHlicmlkIjp7ImlzX2VudGl0bGVkIjp0cnVlfSwibWlncmF0aW9ucyI6eyJpc19lbnRpdGxlZCI6dHJ1ZX0sImFuc2libGUiOnsiaXNfZW50aXRsZWQiOnRydWV9fSwiaWRlbnRpdHkiOnsiYWNjb3VudF9udW1iZXIiOiIwMDAwMDAiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJ1c2VyIiwiZW1haWwiOiJ1c2VyQHVzZXIudXNlciIsImZpcnN0X25hbWUiOiJ1c2VyIiwibGFzdF9uYW1lIjoidXNlciIsImlzX2FjdGl2ZSI6dHJ1ZSwiaXNfb3JnX2FkbWluIjp0cnVlLCJpc19pbnRlcm5hbCI6dHJ1ZSwibG9jYWxlIjoiZW4tVVMifSwiaW50ZXJuYWwiOnsib3JnX2lkIjoiMDAwMDAxIn19fQ=="
 
 PORT="8086"
 CURLCMD='curl -w %{http_code}'
 HEADER="x-rh-identity: $VALIDAUTHSTRING"
+HEADER2="x-rh-identity: $VALIDAUTHSTRING2"
```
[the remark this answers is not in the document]
lavocatt: Sure!

internal/db/db.go
lavocatt: Oh yes, but I'm already implementing your proposition, but thanks for the correction for the name, it was indeed a bit off.

lavocatt: I went for a go 1.13 error definition:

`var ComposeNotFoundError = errors.New("Compose not found")`

Gundersanne (approved):
Thank you! Looks nice.

Would still be nice to have an extra unit test, but it's covered in the integration tests so it's ok.

lavocatt:
Yep, I'm on it :)
Thanks for the review!

internal/tutils/tutils.go
```diff
@@ -75,30 +75,50 @@ func PostResponseBody(t *testing.T, url string, compose interface{}) (*http.Resp
 }
 
 type dB struct {
-	entries []db.ComposeEntry
+	entries          []db.ComposeEntry
```
Gundersanne, line 78: I don't think entries is used at all right now? So we can just drop that field and use map[string][]db.ComposeEntry

internal/tutils/tutils.go
```diff
@@ -75,30 +75,50 @@ func PostResponseBody(t *testing.T, url string, compose interface{}) (*http.Resp
...
 		return err
 	}
-	d.entries = append(d.entries, db.ComposeEntry{
+	dbEntry := db.ComposeEntry{
 		Id:        id,
 		Request:   request,
 		CreatedAt: time.Now(),
-	})
+	}
+	d.entries = append(d.entries, dbEntry)
```
Gundersanne, line 99: This line can be dropped as well.

Gundersanne (changes requested):
Final comments, otherwise looks good.

Integration tests should've been fixed just now so let's see on the next run.

internal/tutils/tutils.go
lavocatt: Sure!

Gundersanne:
CI got fixed in #245, all it needs is a rebase on main

Gundersanne (approved):
LGTM! Thank you

PR #245, opened 2021-09-03
schutzbot: Always grab versions from terraform repo

Gundersanne:
Exit 1 in case this fails. We always want to test IB against the current
version in stage.

### DIFF
--- a/schutzbot/deploy.sh
+++ b/schutzbot/deploy.sh
@@ -22,57 +22,25 @@ fi
 REPO_DIR=repo/image-builder/"${CI_PIPELINE_ID}"
 sudo dnf localinstall -y "$REPO_DIR"/*"${ARCH}".rpm
 
-# Set up osbuild-composer repo
-OSBUILD_DNF_REPO_BASEURL=http://osbuild-composer-repos.s3-website.us-east-2.amazonaws.com
-COMPOSER_DNF_REPO_BASEURL=http://osbuild-composer-repos.s3.amazonaws.com
-# Default values
-OSBUILD_COMMIT=eb74ddf2ef8c203b5d354dfb97adee65babdc147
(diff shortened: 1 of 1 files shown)


schutzbot/deploy.sh
```diff
@@ -22,57 +22,25 @@ fi
...
-          CURL_URL="https://api.github.com/repos/osbuild/image-builder-terraform/contents/terraform.tfvars.json?ref=$IB_TERRAFORM_BRANCH"
-          TERRAFORM_VARS=$(curl -H "Authorization: token $GITHUB_TOKEN" -H "Accept: application/vnd.github.v3.raw" -L "$CURL_URL")
-          OSBUILD_COMMIT=$(echo "$TERRAFORM_VARS" | jq -r .osbuild_commit)
-          OSBUILD_COMPOSER_COMMIT=$(echo "$TERRAFORM_VARS" | jq -r .composer_commit)
-     fi
-fi
-
+# Retrieve composer&osbuild version from current stage deployment
+pushd "$(mktemp -d)"
+chmod 400 "$TERRAFORM_REPO_DEPLOY_KEY"
```
jrusz, line 27: Maybe it needs `sudo` here? Similar to line 17 when making the registration script executable?

Gundersanne: I think the entire thing mihgt need to be sudo, if it's 400 for root then it won't be readable by the user executing the pipeline.

Gundersanne: eh i'll just keep iterating until it's green here :D

teg (approved):
Makes sense. As a follow up is it possible to actually run tests against stage itself though?

Gundersanne:
> Makes sense. As a follow up is it possible to actually run tests against stage itself though?

Yea that's a good idea, but let's unbreak CI first :D

PR #246, opened 2021-09-06
db: Rename account_id to account_number

### LINKED ISSUE #242: Rename references to account_id/accountId to account_number
In CRC `org_id` is used to refer to a number that organization admins have access to really. And account_number is the shared number over an organization.

So let's rename account_id/accountId references to account_number (but keep org_id as is). To keep it consistent and avoid future confusion. Migrate the db as well, since we currently have a single table without any references to account_id, we can just rename it to account_number.

Gundersanne:
Let's avoid confusion, account_numbers are actually shared.

Fixes #242

### DIFF
--- a/cmd/image-builder-db-test/main_test.go
+++ b/cmd/image-builder-db-test/main_test.go
@@ -1,0 +1,158 @@
+// +build integration
+
+package main
+
+import (
+	"context"
+	"fmt"
+        "testing"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4"
+	"github.com/stretchr/testify/require"
+
+	"github.com/osbuild/image-builder/internal/config"
+	"github.com/osbuild/image-builder/internal/db"
+	"github.com/osbuild/image-builder/internal/logger"
+)
+
+const (
+	ANR1 = "000001"
+	ANR2 = "000002"
+	ANR3 = "000003"
+
+	ORGID1= "100000"
+)
+
+func conf(t *testing.T) *config.ImageBuilderConfig {
+	c := config.ImageBuilderConfig{
+		ListenAddress: "unused",
+		LogLevel:      "INFO",
+		MigrationsDir: "/usr/share/image-builder/migrations",
+		PGHost:        "localhost",
+		PGPort:        "5432",
+		PGDatabase:    "imagebuilder",
+		PGUser:        "postgres",
+		PGPassword:    "foobar",
+		PGSSLMode:     "disable",
+	}
+
+	err := config.LoadConfigFromEnv(&c)
+	require.NoError(t, err)
+
+	return &c
+}
+
+func connStr(t *testing.T) string {
+	c := conf(t)
+	return fmt.Sprintf("postgres://%s:%s@%s:%s/%s?sslmode=%s", c.PGUser, c.PGPassword, c.PGHost, c.PGPort, c.PGDatabase, c.PGSSLMode)
+}
+
+func migrateOneStep(t *testing.T) {
+	c := conf(t)
+
+	log, err := logger.NewLogger(c.LogLevel, nil, nil, nil, nil)
+	require.NoError(t, err)
+	require.NotNil(t, log)
+
+	err = db.MigrateSteps(connStr(t), c.MigrationsDir, 1, log)
+	require.NoError(t, err)
+}
+
+func migrateUp(t *testing.T) {
+	c := conf(t)
+
+	log, err := logger.NewLogger(c.LogLevel, nil, nil, nil, nil)
+	require.NoError(t, err)
+	require.NotNil(t, log)
+
+	err = db.Migrate(connStr(t), c.MigrationsDir, log)
+	require.NoError(t, err)
+}
+
+func connect(t *testing.T) *pgx.Conn {
+	conn, err := pgx.Connect(context.Background(), connStr(t))
+	require.NoError(t, err)
+	return conn
+}
+
+// after each migration insert some data which follows the schema of that migration
+func insertData1(t *testing.T) {
+	conn := connect(t)
+	defer conn.Close(context.Background())
+
+	insert := "INSERT INTO composes(job_id, request, created_at, account_id, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4)"
+
+	_, err := conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR1, ORGID1)
+	require.NoError(t, err)
+
+	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR1, ORGID1)
+	require.NoError(t, err)
+
+	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR2, ORGID1)
+	require.NoError(t, err)
+}
+
+func insertData2(t *testing.T) {
+	conn := connect(t)
+	defer conn.Close(context.Background())
+
+	insert := "INSERT INTO composes(job_id, request, created_at, account_number, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4)"
+
+	_, err := conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR1, ORGID1)
+	require.NoError(t, err)
+
+	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR1, ORGID1)
+	require.NoError(t, err)
+
+	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR2, ORGID1)
+	require.NoError(t, err)
+}
+
+func TestMain(t *testing.T) {
+	migrateOneStep(t)
+	insertData1(t)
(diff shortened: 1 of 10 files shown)


Gundersanne:
adding needs work, will clean up unit tests.

teg (dismissed):

Gundersanne:
This will be a first solid test of the DB migration on stage :) Of course it's already tested in the integration test but let's see.

Gundersanne:
adding blocked, let's get #241 in first since this will be easier to rebase.

Gundersanne:
I added a test which migrates one step at a time, inserting data after each step. This way we simulate the real migration process/DB better, having actual data in there.

cmd/image-builder-db-test/main_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR2, ORGID1)
+	require.NoError(t, err)
+}
+
+func insertData2(t *testing.T) {
+	conn := connect(t)
+	defer conn.Close(context.Background())
+
+	insert := "INSERT INTO composes(job_id, request, created_at, account_number, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4)"
```
lavocatt, line 88: Why not using `func (db *dB) InsertCompose(jobId, accountId, orgId string, request json.RawMessage) error {` here ?

lavocatt (commented):
This is nice! I really like the db testing.
Just one comment for me.

cmd/image-builder-db-test/main_test.go
Gundersanne: For insertData2 could use that here, as that is the latest version. But when insertData3 comes around we'd have to change that into sql queries again, which will make the diff a bit uglier for that PR. So really just preventing future changes.

internal/db/migrate.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func Migrate(connStr string, migrationsDir string, logger *logrus.Logger) error {
+func Migrate(connStr string, migrationsDir string, steps int, logger *logrus.Logger) error {
```
lavocatt, line 12: Maybe we can do it in a way that leaves the compatibility with the previous API of migrate?

`func Migrate(connStr string, migrationsDir string, logger *logrus.Logger) error {` does the full migration
`func MigrateStep(connStr string, migrationsDir string, steps int, logger *logrus.Logger) error {` executes the given step for the migration. With an precodition on steps being > 0.

This way we can avoid the Hack of the step 0.

Another solution, could be to make `Migrate` invoke a callback function between each step.

What are your thoughts ?

cmd/image-builder-db-test/main_test.go
lavocatt: Ah yes, I haven't thought about that!

internal/db/migrate.go
Gundersanne: I'll add a 2nd function, I think that'll make it the easiest to read as well.

lavocatt (approved):
LGTM thanks :)

Gundersanne:
I added an additional Up() migration after all migrations were applied, to make sure that runs fine and doesn't return any errors. As that's the scenario that is executed the most (on updates to main without changes in the database schema).

internal/db/migrate.go
```diff
@@ -10,48 +11,80 @@ import (
...
+	}()
+	if err != nil {
+		return err
+	}
+	return finish(m, m.Up(), logger)
+}
+
+func MigrateSteps(connStr string, migrationsDir string, steps int, logger *logrus.Logger) error {
+	m, err := prepare(connStr, migrationsDir, logger)
+	defer func() {
```
lavocatt, line 35: I see that you've extracted these checks from finish, but I don't understand the reason, can you explain me why ?

cmd/image-builder-db-test/main_test.go
```diff
@@ -0,0 +1,157 @@
...
+	migrateOneStep(t)
+	insertData2(t)
+
+	// make sure migrating a fully migrated db doesn't error out
+	migrateUp(t)
+
+	d, err := db.InitDBConnectionPool(connStr(t))
+	require.NoError(t, err)
+
+	composes, count, err := d.GetComposes(ANR1, 100, 0)
```
lavocatt, line 125: Thinking about this part, to ease future amendments, I'd define tests scenarios with their own functions.

```
test_migration()
test_count_returns_total_in_db()
test_selected_compose_equals_inserted_compose()
test_no_cross_account_access()
test_compose_insertion()
```

what are your thoughts ?

internal/db/migrations/2_alter_table_rename_account.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE composes RENAME COLUMN account_number TO account_id;
```
lavocatt, line 1: Should we also test migrating down ?

lavocatt (commented):
Hi, looking good!
I left some comments and questions, thanks.

internal/db/migrations/2_alter_table_rename_account.down.sql
Gundersanne: They're there mostly for completeness and potential future debugging of db issues. We never migrate a live database down, only up. (So say a mistake in a migration.sql file got made, it would have to be fixed by an additional up migration if at all possible).

cmd/image-builder-db-test/main_test.go
Gundersanne: Depends what you mean; 

If it's introducing more helpers, I think it's too little for now to split up into more helper functions.

If you mean to make more Test* functions: we can't rely on the order of unit tests,  and those should be able to be executed in parallel. So we can really only have one Test* function per database, and there's only one database atm. It's possible I suppose if we move the podman run call into the test itself, and do the migration per test.

internal/db/migrate.go
Gundersanne: This was actually a small bug, where errors would result in the migration/connection not being closed. I know the placement of the err != nil check and defer might seem a little weird, but the problem is that the migration gets opened in prepare:

If prepare returns an error we need to return early and close the migration, but m might be nil because it could have failed in `migrate.New` already. So we need the nil check in defer.

defer here can't be part of finish because the migration might have been opened in `prepare`, errored out in `prepare` after opening, and it would never reach `finish`, so it would never close, which wouldn't fix the bug.

---

The alternative is moving the migrate.New call into the `Migrate*` functions again, maybe a bit more readable:

```
func MigrateSteps(connStr string, migrationsDir string, steps int, logger *logrus.Logger) error {
	logger.Infoln("Applying migrations from directory ", fmt.Sprintf("file://%s", migrationsDir))
	m, err := migrate.New(fmt.Sprintf("file://%s", migrationsDir), connStr)
	if err != nil {
		return nil, err
	}

	defer func() {
		srcErr, dbErr := m.Close()
		if srcErr != nil {
			logger.Errorln("Source error when closing migration", srcErr)
		}
		if dbErr != nil {
			logger.Errorln("DB error when closing migration", dbErr)
		}
	}()


	m, err := prepare(m, connStr, migrationsDir, logger)
	if err != nil {
		return err
	}
	return finish(m, m.Steps(steps), logger)
}
```

cmd/image-builder-db-test/main_test.go
lavocatt: Yest, I see it as 'introducing more helpers'. Just to keep track of the meaning of each steps of the test.

internal/db/migrations/2_alter_table_rename_account.down.sql
lavocatt: I see, okay, it makes sense.

internal/db/migrate.go
lavocatt: I see, that's a good point indeed!

You could also define a new function:
```
func cleanup(m *migrate.Migrate) error {
	if m == nil {
		return
	}
	srcErr, dbErr := m.Close()
	if srcErr != nil {
		logger.Errorln("Source error when closing migration", srcErr)
	}
	if dbErr != nil {
		logger.Errorln("DB error when closing migration", dbErr)
	}
}
```

and call `defer cleanup(m)`

cmd/image-builder-db-test/main_test.go
lavocatt: But yeah also, as you said, this is a simple test, so it might be too much splitting, your call :)

Gundersanne: Right, not a huge fan I think because the state of the DB would change in between those helpers, and the helpers would read as subtests. I think it's fine to split it up more if each unit test would get it's own database. But let's do that in a followup then.

I added some comments in stead. And then let's [make the test invoke podman](https://podman.io/blogs/2020/08/10/podman-go-bindings.html) in a future iteration.

lavocatt: I completely agree.

lavocatt (approved):

Issue #247, opened 2021-09-06
Rate limiting of image builds


Gundersanne:
- Per account_number
- Each account_number is able to have it's own limit (some accounts may be allowed more builds than others)
- Limit per 2 week period?

Gundersanne:
Implemenation:
- This list should not live upstream, but mapped to the container (as a configmap? can we define this list in app-intf?)
- This should not require any changes in the db migrations since all entries already carry a timestamp + account number

Gundersanne:
@teg So say the jobs keep failing (you started 100 image builds and they all failed). Do we wanna count those towards the limit?

teg:
Good question!

I would say no, the quota should be on the number of images we store.

Might also want a second (much higher) limit on the number of (attempted) image builds to avoid someone DoS'ing us with failing builds.

Gundersanne:
> I would say no, the quota should be on the number of images we store.

So currently we store all images, but should this have a time limit as well? Like the amount of images stored in the past 14 days?


> Might also want a second (much higher) limit on the number of (attempted) image builds to avoid someone DoS'ing us with failing builds.

teg:
Yeah, that was what I was thinking: put the quote on the number of stored images in the past 14 days, under the assumption we delete images after 14 days.

Gundersanne:
That's gonna be a bit harder because we need the final status of each of those builds then. So we'd need to query composer for each of the builds they did in the past 14 days. Not sure if that's a good idea.

teg:
Ah, we can definitely use the number of attempted builds as an approximation for now, but yeah we should figure out a long term plan

Gundersanne:
Let's have a generous default for now, and then long term let's see,  I think we can't really do this in any reasonable sense without having IB store/cache the result.

PR #255, opened 2021-09-14
DB: search for composes for a time period

lavocatt:
This commit introduces a new method to search for composes made a by a
user for a given time period. This adresses the first part of the
issue #247

### DIFF
--- a/cmd/image-builder-db-test/main_test.go
+++ b/cmd/image-builder-db-test/main_test.go
@@ -77,69 +77,92 @@ func connect(t *testing.T) *pgx.Conn {
 	return conn
 }
 
-// after each migration insert some data which follows the schema of that migration
-func insertDataForMigrationStep1(t *testing.T) {
+func tearDown(t *testing.T) {
 	conn := connect(t)
 	defer conn.Close(context.Background())
+	conn.Exec(context.Background(), "drop table composes")
+	conn.Exec(context.Background(), "drop table schema_migrations")
+}
 
+func testMigration(t *testing.T) {
+	defer tearDown(t) // tear-down cleanup the database
+
+	migrateOneStep(t) //migrate to step 1
+
+	conn := connect(t)
+	defer conn.Close(context.Background())
 	insert := "INSERT INTO composes(job_id, request, created_at, account_id, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4)"
-
 	_, err := conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR1, ORGID1)
 	require.NoError(t, err)
 
-	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR1, ORGID1)
-	require.NoError(t, err)
+	migrateOneStep(t) // migrate to step 2
 
+	insert = "INSERT INTO composes(job_id, request, created_at, account_number, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4)"
 	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR2, ORGID1)
 	require.NoError(t, err)
 
-	insert = "INSERT INTO composes(job_id, request, created_at, account_id, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP - interval '2 days', $3, $4)"
-
-	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR3, ORGID1)
-	require.NoError(t, err)
-
-	insert = "INSERT INTO composes(job_id, request, created_at, account_id, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP - interval '3 days', $3, $4)"
-
-	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR3, ORGID1)
-	require.NoError(t, err)
-
-	insert = "INSERT INTO composes(job_id, request, created_at, account_id, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP - interval '4 days', $3, $4)"
-
-	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR3, ORGID1)
-	require.NoError(t, err)
-}
-
-func insertDataForMigrationStep2(t *testing.T) {
-	conn := connect(t)
-	defer conn.Close(context.Background())
-
-	insert := "INSERT INTO composes(job_id, request, created_at, account_number, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4)"
-
-	_, err := conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR1, ORGID1)
-	require.NoError(t, err)
-
+	// inserting data referring to account_id should fail after migration step 2
+	insert = "INSERT INTO composes(job_id, request, created_at, account_id, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4)"
 	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR1, ORGID1)
-	require.NoError(t, err)
+	require.Error(t, err)
 
-	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR2, ORGID1)
-	require.NoError(t, err)
-}
+	migrateOneStep(t) // migrate to step 3
 
-func TestMain(t *testing.T) {
-	migrateOneStep(t)
-	insertDataForMigrationStep1(t)
-
-	migrateOneStep(t)
-	insertDataForMigrationStep2(t)
-
-	migrateOneStep(t)
+	// Verify that after migration step 3 adding a compose request to the db requires a non empty account number.
+	d, err := db.InitDBConnectionPool(connStr(t))
+	err = d.InsertCompose(uuid.New().String(), "", ORGID1, []byte("{}"))
+	require.Error(t, err)
 
 	// make sure migrating a fully migrated db doesn't error out
 	migrateUp(t)
 
+	// Check data inserted at migration step 1 and 2 are still accessible
+	_, count, err := d.GetComposes(ANR1, 100, 0)
+	require.NoError(t, err)
+	require.Equal(t, 1, count)
+	_, count, err = d.GetComposes(ANR2, 100, 0)
+	require.NoError(t, err)
+	require.Equal(t, 1, count)
+}
+
+func testInsertCompose(t *testing.T) {
 	d, err := db.InitDBConnectionPool(connStr(t))
 	require.NoError(t, err)
 
+	// teardwon
+	defer tearDown(t)
+
+	// setup
+	migrateUp(t)
+
+	// test
+	err = d.InsertCompose(uuid.New().String(), ANR1, ORGID1, []byte("{}"))
+	require.NoError(t, err)
+	err = d.InsertCompose("toto", ANR1, ORGID1, []byte("{}"))
+	require.Error(t, err)
+	err = d.InsertCompose(uuid.New().String(), "", ORGID1, []byte("{}"))
+	require.Error(t, err)
+}
+
+func testGetCompose(t *testing.T) {
+	d, err := db.InitDBConnectionPool(connStr(t))
+	require.NoError(t, err)
+
+	// teardwon
+	defer tearDown(t)
+
+	// setup
+	migrateUp(t)
+	err = d.InsertCompose(uuid.New().String(), ANR1, ORGID1, []byte("{}"))
+	require.NoError(t, err)
+	err = d.InsertCompose(uuid.New().String(), ANR1, ORGID1, []byte("{}"))
+	require.NoError(t, err)
+	err = d.InsertCompose(uuid.New().String(), ANR1, ORGID1, []byte("{}"))
+	require.NoError(t, err)
+	err = d.InsertCompose(uuid.New().String(), ANR1, ORGID1, []byte("{}"))
+	require.NoError(t, err)
+
+	// test
 	// GetComposes works as expected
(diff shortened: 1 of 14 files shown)


cmd/image-builder-db-test/main_test.go
```diff
@@ -109,6 +109,26 @@ func insertData2(t *testing.T) {
...
+
+	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR3, ORGID1)
+	require.NoError(t, err)
+
+	insert = "INSERT INTO composes(job_id, request, created_at, account_number, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP - interval '4 days', $3, $4)"
+
+	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR3, ORGID1)
+	require.NoError(t, err)
+}
+
```
Gundersanne, line 131: Since this operates on data that existed since the first migration, and it's not specific to the 3rd migration file, this should be moved to insertData1. I added the comment on line 79 for this but could also rename the functions insertDataMigration1/2/3/...

internal/db/db.go
```diff
@@ -4,13 +4,78 @@ import (
...
+		sign = "-"
+	default:
+		return "", TimeIntervalInvalidOperator
+	}
+	if tI.Amount <= 0 {
+		return "", TimeIntervalInvalidAmount
+	}
+	return fmt.Sprintf("%s interval '%d %s'", sign, tI.Amount, sUnit), nil
+}
+
```
Gundersanne, line 78: I think this is a bit too much for what we need, and furthermore pgx4 supports time.Duration to interval https://github.com/jackc/pgtype/blob/master/interval.go, so if we want something with a bit more functionality let's use `time.Duration`.

internal/db/db.go
```diff
@@ -28,6 +93,10 @@ type DB interface {
 	InsertCompose(jobId, accountNumber, orgId string, request json.RawMessage) error
 	GetComposes(accountNumber string, limit, offset int) ([]ComposeEntry, int, error)
 	GetCompose(jobId string, accountNumber string) (*ComposeEntry, error)
+	// Returns the count of compose requests performed by an accountNumber since the current date for the duration of (the
+	// sum of) the time intervals. The sum of the time intervals should be negative for the db to search in the past, and
+	// positive to search in the future.
+	CountComposesFor(accountNumber string, intervals ...TimeInterval) (int, error)
```
Gundersanne, line 99: "For" isn't super clear here. `CountComposesSince(accountNumber string, since time.Duration)` reads clearer, can drop the comment then as well.

internal/db/db.go
```diff
@@ -117,3 +186,31 @@ func (db *dB) GetComposes(accountNumber string, limit, offset int) ([]ComposeEnt
...
+		if err != nil {
+			return 0, err
+		}
+		intervals = fmt.Sprintf("%s %s", intervals, sTimeInterval)
+	}
+
+	var count int
+	err = conn.QueryRow(ctx, fmt.Sprintf(
+		"SELECT COUNT(*) FROM composes WHERE account_number=$1 AND created_at > CURRENT_TIMESTAMP%s", intervals),
+		accountNumber).Scan(&count)
```
Gundersanne, line 210: Don't use `fmt.Sprintf` in queries, rely on the formatting and sanitizing done by pgx.

And with `time.Duration`: `SELECT COUNT(*) FROM composes WHERE account_number=$1 AND CURRENT_TIMESTAMP - created_at <=  $2` works with `time.Duration` for $2.

Gundersanne (commented):
Thanks! Some comments.

internal/db/migrate.go
```diff
@@ -28,6 +28,15 @@ func MigrateSteps(connStr string, migrationsDir string, steps int, logger *logru
...
 
+func MigrateDown(connStr string, migrationsDir string, logger *logrus.Logger) error {
+	m, err := prepare(connStr, migrationsDir, logger)
+	defer cleanup(m, logger)
+	if err != nil {
+		return err
+	}
+	return finishDown(m, m.Down(), logger)
+}
+
```
Gundersanne, line 39: So I'm not sure how I feel about adding support for something that we'll never do in a deployment. I added the downward migration files for completeness but maybe we should just remove them. Broken migrations should really be fixed by another migration.

I think it's better to just drop/create the database in the test then. But other than that I'm happy with splitting up the tests like you did.

lavocatt: I have to admit that I kinda like the fact that the DB can migrate down. But if that's pointless, we can remove the .down files and drop the table or database manually at the end of the test. And if one day we need to step down a migration we still can develop the feature when needed.

Gundersanne: I think that's best. The only way we'll roll back is via restoring a snapshot in aws rds which doesn't include the migration process. Which would of course be the worst case scenario.

Gundersanne: https://github.com/osbuild/image-builder/pull/258 opened as well; as soon as the db is in some sort of bad state a human needs to investigate.

lavocatt: Yep, this makes complete sense. I'll add a commit to remove all of this .down stuff :)

cmd/image-builder-db-test/main_test.go
```diff
@@ -76,54 +77,88 @@ func connect(t *testing.T) *pgx.Conn {
...
-	insertData1(t)
-
-	migrateOneStep(t)
-	insertData2(t)
+func testInsertCompose(t *testing.T) {
+	d, err := db.InitDBConnectionPool(connStr(t))
+	require.NoError(t, err)
 
-	migrateOneStep(t)
+	// teardwon
```
Gundersanne, line 132: nitpick: "teardown"

internal/common/quota.go
```diff
@@ -0,0 +1,60 @@
...
+	defaultSlidingWindow time.Duration = 2 * week
+	defaultQuota         int           = 100
+)
+
+type quota struct {
+	Quota         int           `json:"quota"`
+	SlidingWindow time.Duration `json:"slidingWindow"`
+}
+
+// Returns true if the number of requests made by accountNumber during a sliding window is inferior to a threshold.
```
Gundersanne, line 25: "inferior to a threshold" sounds a bit unclear to me, how about "below a threshold".

internal/common/quota.go
```diff
@@ -0,0 +1,60 @@
...
+// The duration of the sliding window and the value of the threshold can be set in /config/accounts_quotas.json
```
Gundersanne, line 26: I think we need to make the exact path configurable via the env, so `QUOTA_FILE`  (in internal/config) or something. And everything tends to be in `/app/` (see the dockerfile).

cmd/image-builder-db-test/main_test.go
```diff
@@ -71,75 +71,97 @@ func migrateUp(t *testing.T) {
...
+	defer migrateDown(t)
 
-	// make sure migrating a fully migrated db doesn't error out
+	// setup
 	migrateUp(t)
+	err = d.InsertCompose(uuid.New().String(), ANR1, ORGID1, []byte("{}"))
+	err = d.InsertCompose(uuid.New().String(), ANR1, ORGID1, []byte("{}"))
+	err = d.InsertCompose(uuid.New().String(), ANR1, ORGID1, []byte("{}"))
+	err = d.InsertCompose(uuid.New().String(), ANR1, ORGID1, []byte("{}"))
```
Gundersanne, line 159: Check the errors here.

internal/common/quota.go
```diff
@@ -0,0 +1,60 @@
...
+// By default the values are set to 100 requests per 2 weeks.
+func CheckQuota(accountNumber string, dB db.DB) (bool, error) {
+	authorizedRequests := defaultQuota
+	slidingWindow := defaultSlidingWindow
+
+	// read proper values from quotas' file
+	var quotas map[string]quota
+	jsonFile, err := os.Open("/config/accounts_quotas.json")
+	if _, ok := err.(*os.PathError); ok {
+		fmt.Println("No config file for quotas found")
```
Gundersanne, line 36: return the patherror here.

internal/common/quota.go
```diff
@@ -0,0 +1,60 @@
...
+			return false, err
+		}
+		if quota, ok := quotas[accountNumber]; ok {
+			authorizedRequests = quota.Quota
+			slidingWindow = quota.SlidingWindow
+		} else if quota, ok := quotas["default"]; ok {
+			authorizedRequests = quota.Quota
+			slidingWindow = quota.SlidingWindow
+		} else {
+			fmt.Println("No default values in the quotas' file")
```
Gundersanne, line 50: fmt.Errorf with the quota file path here.

internal/v1/server.go
```diff
@@ -408,8 +408,21 @@ func buildOSTreeOptions(ostreeOptions *OSTree) *cloudapi.OSTree {
...
+	if err != nil {
+		return err
+	}
+
+	quotaOk, err := common.CheckQuota(idHeader.Identity.AccountNumber, h.server.db)
+	if err != nil {
+		return err
+	}
+	if !quotaOk {
+		return echo.NewHTTPError(http.StatusUnauthorized, "Quota exceeded for user")
```
Gundersanne, line 421: http codes are badly named >.> 

`http.StatusForbidden` is actually correct here since that's an authorization failure. `http.StatusUnauthorized` is an authentication failure. And keep in mind that any 401s for CRC would need to be 404s. But in this case a 403 is the correct one.

Gundersanne (changes requested):
Thank you! Looking good, some minor comments.

Also `DB: Migrating down is not an option` can be squashed into `DB: refactor the testing of the DB`, the diff between those two is a bit confusing now.

cmd/image-builder-db-test/main_test.go
lavocatt: I'm not sure we should, since it'll blow up in the test just below.

internal/common/quota.go
lavocatt: Good idea!

internal/common/quota.go
lavocatt: Yeah, this might be literal translation issues ^^

cmd/image-builder-db-test/main_test.go
Gundersanne: failing earlier is better, it'll make it easier to debug. Failures here are unlikely but you'd get an exact line number etc...

lavocatt: Good point indeed.

internal/v1/server.go
lavocatt: I'll try to keep this in mind :sweat_smile:

Gundersanne:
Still marking this as blocked as the app-intf changes need to land beforehand (Just in case someone comes along and merges it).

templates/image-builder.yml
```diff
@@ -63,6 +63,8 @@ objects:
             - name: composer-secrets
               mountPath: "/app/composer-secrets"
               readOnly: true
+            - name: config-volume
+              mountPath: /etc/config/image-builder
```
Gundersanne, line 67: The container doesn't follow FHS. To avoid any conflicts just put everything in `/app`. So `/app/quotas` or something should do.

templates/image-builder.yml
```diff
@@ -159,10 +161,18 @@ objects:
               value: "${ALLOWED_ACCOUNT_NUMBERS}"
             - name: DISTRIBUTIONS_DIR
               value: "${DISTRIBUTIONS_DIR}"
+            - name: QUOTA_FILE
+              value: "/etc/config/image-builder/accounts_quotas.json"
```
Gundersanne, line 165: ditto

internal/v1/server_test.go
```diff
@@ -22,16 +25,55 @@ import (
...
+	// create quotas with only the default values
+	quotas := map[string]common.Quota{
+		"default": {Quota: common.DefaultQuota, SlidingWindow: common.DefaultSlidingWindow},
+	}
+	jsonQuotas, err := json.Marshal(quotas)
+	if err != nil {
+		return "", err
+	}
+
+	fmt.Println("pouet-1")
```
Gundersanne, line 39: Remaning debug print.

internal/common/quota.go
```diff
@@ -0,0 +1,81 @@
...
+// environment variable.
+func CheckQuota(accountNumber string, dB db.DB, quotaFile string) (bool, error) {
+	var authorizedRequests int
+	var slidingWindow time.Duration
+
+	// read proper values from quotas' file
+	var quotas map[string]Quota
+	jsonFile, err := os.Open(filepath.Clean(quotaFile))
+	if _, ok := err.(*os.PathError); ok {
+		fmt.Printf("No config file for quotas found at %s\n", quotaFile)
```
Gundersanne, line 55: We don't capture stdout/stderr, the logger object logs straight to cloudwatch. Either we return the patherror or `return false, fmt.Errorf(....)`

internal/common/quota.go
```diff
@@ -0,0 +1,81 @@
...
+			return false, err
+		}
+		if quota, ok := quotas[accountNumber]; ok {
+			authorizedRequests = quota.Quota
+			slidingWindow = quota.SlidingWindow
+		} else if quota, ok := quotas["default"]; ok {
+			authorizedRequests = quota.Quota
+			slidingWindow = quota.SlidingWindow
+		} else {
+			fmt.Printf("No default values in the quotas' file %s\n", quotaFile)
```
Gundersanne, line 70: Same as above.

Gundersanne (changes requested):
1 remaining comment re: logging

internal/v1/server_test.go
lavocatt: oh yes -_-'

Gundersanne (approved):
Alright, the confimap should be in place. Let's see how it goes :)

Gundersanne (approved):
lgtm!

PR #262, opened 2021-09-23
prometheus: Custom buckets for http request duration histogram

Gundersanne:
Compose requests can take anywhere from 3 to 12 seconds, and the default
buckets put everything above 10 in the infinite bucket. By introducing
more buckets compose requests should be captured better.

---

Note that the first 5 buckets are considered acceptable for non-compose requests. And the first 17 acceptable for compose requests. Anything above 12 for compose requests is considered (really) slow.

### DIFF
--- a/internal/common/prometheus.go
+++ b/internal/common/prometheus.go
@@ -10,8 +10,9 @@ import (
 
 var (
 	httpDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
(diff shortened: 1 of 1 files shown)


teg (dismissed):
Thanks!

teg (approved):

ondrejbudai (approved):
Thanks!

PR #268, opened 2021-10-12
v1: Return version without reading openapi spec

Gundersanne:


kingsleyzissou (approved):
Awesome!

teg (commented):
I wonder if we could just cache the spec?

Issue #256, opened 2021-09-15
db: Check for Err() on Rows result after it's closed


Gundersanne:
In case the SQL was invalid, `Rows.Err()` is nil until the rows are closed.

```
diff --git a/internal/db/db.go b/internal/db/db.go
index 2c0ee44..8c20f0d 100644
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -108,6 +108,9 @@ func (db *dB) GetComposes(accountNumber string, limit, offset int) ([]ComposeEnt
                        createdAt,
                })
        }
+       if result.Err() != nil {
+               return result.Err()
+       }
 
        var count int
        err = conn.QueryRow(ctx, "SELECT COUNT(*) FROM composes WHERE account_number=$1", accountNumber).Scan(&count)
```

`QueryRow` returns the error immediately as that closes the Rows immediately.

PR #275, opened 2021-10-19
db: check for err after rows closed

chloenayon:
Add call to result.Err() after rows are closed by result.Next()
to check for invalid SQL not caught by earlier error handling.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -118,6 +118,9 @@ func (db *dB) GetComposes(accountNumber string, since time.Duration, limit, offs
 			createdAt,
 		})
 	}
+	if err = result.Err(); err != nil {
+		return nil, 0, err
+	}
 
 	var count int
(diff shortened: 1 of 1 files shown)


Gundersanne:
> Sounds good to me, but out of curiosity: How did this come up? I.e., did you run into the error being missed, and if so what would trigger it?

see https://github.com/osbuild/image-builder/issues/256 ; currently we don't actually check result.Err(), which we should. And it should be checked after Rows has been closed (Next() closes it automatically after all rows have been read). 

Any errors in the actual SQL will not be returned by conn.Query; but in the result(.Err()) itself.

Maybe it's useful to update the commit message with a bit more info however.

Also add the Fixes #256 and rebase to get rid of the security errors and test against the live deployment.

teg:
Thanks for the explanation!

Gundersanne (approved):

PR #284, opened 2021-11-04
templates: change dashboard name & id

kingsleyzissou:
The current dashboard is still not showing in stage or production and it's possibly
due to a conflict with the name and id. App-sre has suggested changing this to
resolve the issue.


teg (approved):
Not tested in CI. Makes sense.

PR #285, opened 2021-11-05
CI: Journal-log is accessible and encrypted

lavocatt:


schutzbot/save_journal.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# use tee, otherwise shellcheck complains
+sudo journalctl --boot | tee journal-log >/dev/null
+
+# As it might contain sensitive information and is important for debugging
+# purposes, encrypt journal-log with a symmetric passphrase.
+gpg --batch --yes --passphrase "$GPG_SYMETRIC_PASSPHRASE" -o journal-log.gpg --symmetric journal-log
```
achilleas-k, line 8:
```suggestion
gpg --batch --yes --passphrase "$GPG_SYMMETRIC_PASSPHRASE" -o journal-log.gpg --symmetric journal-log
```
A bit of a nitpick but if there's a typo in the variable name I'm sure it'll trip someone up in the future

achilleas-k (changes requested):

schutzbot/save_journal.sh
lavocatt: Indeed, good catch.

ondrejbudai (commented):
Looks good, needs a rebase though.

achilleas-k (approved):

msehnout:
Can we make the artifact disappear after some predefined time? In case the password ever leaked.

jrusz:
We can set expiration time for artifacts https://docs.gitlab.com/ee/ci/pipelines/job_artifacts.html

lavocatt:
The expiration is 30 days by default https://docs.gitlab.com/ee/user/admin_area/settings/continuous_integration.html#default-artifacts-expiration. If I understand the documentation correctly, the artifacts are kept indefinitely until a new job producing the same artifact runs successfully. Making the new artifact replacing the previous one. So if the password leaks, we should be able to fix the leakage by simply rerun the pipeline with a new password for the GPG key.

msehnout (approved):
Looks good :+1: 

I'd also like to see a proposal for key distribution and rotation, but preferably in a private channel, not on Github :-)

achilleas-k (approved):

PR #287, opened 2021-11-09
Add rhel 85

Gundersanne:


Gundersanne:
I want to switch over the edge test as well actually, to 8.5. But let's do that in a followup just for the sake of CI.

ondrejbudai (approved):
Very nice!

Issue #291, opened 2021-11-12
TLS Cert error in dev stack


kingsleyzissou:
In the dev stack, the backend is unable to communicate with the composer instance because of self-signed certs. Since the composer client is using offline tokens, this is not an issue in production.

I'm not too sure if we want to "fix" this in the dev stack, but I can push a POC fix for this.

![tls](https://user-images.githubusercontent.com/20438192/141465055-10afe153-652b-4c6b-a29e-ff4c20c70175.png)

PR #292, opened 2021-11-12
composer: add tls config

kingsleyzissou:
@Gundersanne @ondrejbudai I've had to re-add some tls config when creating the composer`client. I've tried to make the changes as noninvasive as possible. The reason for this is that in the dev stack the backend is unable to communicate with the composer instance because of self-signed certs (https://github.com/osbuild/image-builder/issues/291)

I'm not sure if we want to include this "fix", so I've marked it as draft for now and happy to hear your thoughts.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -41,7 +41,7 @@ func main() {
 		panic(err)
 	}
 
-	client, err := composer.NewClient(conf.ComposerURL, conf.ComposerTokenURL, conf.ComposerOfflineToken)
+	client, err := composer.NewClient(conf.ComposerURL, conf.ComposerTokenURL, conf.ComposerOfflineToken, conf.ComposerCA)
 	if err != nil {
 		panic(err)
 	}
--- a/internal/composer/client.go
+++ b/internal/composer/client.go
@@ -4,11 +4,15 @@ package composer
 
 import (
 	"bytes"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strings"
 	"sync"
 	"time"
 )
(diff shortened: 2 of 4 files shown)


ondrejbudai:
Can't composer just ignore a bad certificate?

kingsleyzissou:
> Can't composer just ignore a bad certificate?
That sounds like a much simpler solution. What would I need to do in composer to change that?

ondrejbudai:
Chm, I might be confused. Does the local dev stack support OAuth, or is it currently limited to mTLS?

kingsleyzissou:
> Chm, I might be confused. Does the local dev stack support OAuth, or is it currently limited to mTLS?

I've got this draft PR to add oauth support (So it's using `jwt` and not `mTLS`):
https://github.com/RedHatInsights/image-builder-frontend/pull/428
The oauth server is using `http` though, because of the certs.

ondrejbudai:
Right, thanks for your patience with me. :D I don't think that you actually need to provide a key and a cert here, right? I think that supplying just the CA should be fine (we are not using mTLS anymore, so image-builder doesn't actually need to send any certs, it just needs to trust the authority that signed the composer's cert). I would also rename the variable to `COMPOSER_CA_PATH`.

This feels slightly cleaner, just disabling the CA check should be the last resort I think.

kingsleyzissou:
No not at all! Oh yeah of course, let me try it without the key and cert! Agreed about disabling the CA check as the last resort.

kingsleyzissou:
@ondrejbudai yeah it works with just the CA

ondrejbudai (approved):

ondrejbudai:
Ah, GoSec complains, you need to put back the line that set minTLS to 1.2.

kingsleyzissou:
> Ah, GoSec complains, you need to put back the line that set minTLS to 1.2.

Just saw that. Should be good now.

Gundersanne:
you can just turn of tls as well
```
      enable_tls = false
      enable_mtls = false
```

the idea was that in that case you had to enable jwt but since I wrote it the check for that isn't ever executed https://github.com/osbuild/osbuild-composer/blob/main/cmd/osbuild-composer/composer.go#L122 :penguin:

kingsleyzissou:
Yeah `enable_tls=false` is also another option, but we'd need to make the protocol for the workers configurable, since they're hard-coded to `https`. I don't think we would need to enable `.KeysInsecure(bool)` for that solution either.

Gundersanne (approved):

PR #293, opened 2021-11-13
add rhel-85 to package endpoint

matoval:
Add rhel-85 to /package endpoint

I'm getting this error when trying to search for rhel-85 package:
`{
    "detail": "parameter \"distribution\" in query has an error: value is not one of the allowed values\nSchema:\n  {\n    \"enum\": [\n      \"rhel-84\",\n      \"centos-8\"\n    ],\n    \"type\": \"string\"\n  }\n\nValue:\n  \"rhel-85\"\n",
    "title": "400"
}`


Gundersanne (approved):

Gundersanne (changes requested):
Thank you, missed that. 

Can you run `/tools/prepare-source.sh` and push again? It should generate the api, which is why the checks are complaining.

teg:
Would be good to have some tests for this, I think.

Gundersanne:
> Would be good to have some tests for this, I think.

agreed, I can do that in a followup

Gundersanne (approved):
lgtm, thanks!

ondrejbudai (approved):

Gundersanne:
The gcp failure is unrelated, I'll merge this since it's relatively high priority.

PR #296, opened 2021-11-18
v1: Enable rhel-edge-container

Gundersanne:
- [x] https://github.com/osbuild/osbuild-composer/pull/2035

### DIFF
--- a/distributions/rhel-84.json
+++ b/distributions/rhel-84.json
@@ -5,7 +5,7 @@
     "description": "Red Hat Enterprise Linux (RHEL) 8"
   },
   "x86_64": {
-    "image_types": [ "aws", "gcp", "azure", "edge-commit", "edge-installer", "ami", "vhd", "rhel-edge-commit", "rhel-edge-installer", "rhel-edge-container" ],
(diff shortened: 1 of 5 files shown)


achilleas-k (approved):

achilleas-k (changes requested):
Do we also need to add `edge-container` and `rhel-edge-container` to the list of image types for the S3 upload target?
https://github.com/osbuild/image-builder/blob/31e47e202eca0dee785b9c422bd6bf8f0fb17223/internal/v1/server.go#L519-L540

teg (commented):
Hmm... Does the edge team want these containers? I thought they only wanted the simplified installer?

Gundersanne:
> Hmm... Does the edge team want these containers? I thought they only wanted the simplified installer?

I'm not sure actually, @achilleas-k  do you remember? I'll open a PR in the meantime with the extra types for rhel-8.5 because those are important (and unrelated).

achilleas-k:
> > Hmm... Does the edge team want these containers? I thought they only wanted the simplified installer?
> 
> I'm not sure actually, @achilleas-k do you remember? I'll open a PR in the meantime with the extra types for rhel-8.5 because those are important (and unrelated).

I don't remember if they asked for them either.  I think we decided to add it anyway since it's an image we build and it's useful.  The general plan is to add every image type to the backend eventually, isn't it?

teg:
I think I may have been the source of the confusion :-/

Gundersanne:
rhel-edge-container is

PR #299, opened 2021-11-19
image-builder: Drop identity header check

Gundersanne:


teg (dismissed):

teg (dismissed):

Gundersanne:
Latest gcloud version has a bug in it, hopefully it'll be fixed soon :/

teg:
Should we pin a known working version? Assuming the issue is only in the CLI

Gundersanne:
> Should we pin a known working version? Assuming the issue is only in the CLI

Also a workaround with some env var, i'll open a PR

teg (approved):

PR #303, opened 2021-11-30
templates: drop the secrets volume

ondrejbudai:
We don't actually use any file-based secrets anymore so let's just drop it.


ondrejbudai:
Needs https://github.com/osbuild/image-builder/pull/304 to unbreak the CI.

Gundersanne (approved):

PR #311, opened 2022-01-18
templates/dashboards: Set uid to a fixed string

Gundersanne:
This way the generated url on app-sre's grafana instances will always be
`...d/image-builder-crc`.

---

I plan to update all references to the dashboard after this, and then we'll hopefully never have to make sure the urls are up to date ever again :)


kingsleyzissou (approved):
Oh that's genius!

kingsleyzissou (approved):

Gundersanne:
Unrelated failure, no resources available in a specific gcp zone, they'll come back when they come back :)

PR #313, opened 2022-01-21
composer: Lazy token refresh

Gundersanne:


ondrejbudai (commented):
So `ExpiresIn` is now basically unused and we're just relying on 403, right?

Gundersanne:
> So `ExpiresIn` is now basically unused and we're just relying on 403, right?

Ah you're right, I Can drop that. But yea that's the idea, could also add an expiration check but you need the retry on 401 regardless for edge case.

ondrejbudai (approved):
I don't see any harm in doing it this way and it handles the edge-cases better I think. :)

PR #322, opened 2022-02-11
schutzbot: add load test slack notification

jabia99:
load test is being executed weekly but has to be reviewed manually. Send
a slack notification with the test result.

Note: permissions were wrong


jabia99:
It's a reopened PR as I didn't create a new branch. And check gives me a strange error

PR #325, opened 2022-02-11
v1: support image name parameter in compose request 

jkozol:
When a user makes a compose request to image builder they should be able to specify an image name that is stored alongside the compose request similar to the uuid. If no image name is specified than none will be stored or returned when the compose info is requested.


internal/v1/api.yaml
```diff
@@ -433,6 +436,8 @@ components:
         request: {}
         created_at:
           type: string
+        image_name:
```
jkozol, line 439: By setting the image_name as an optional return value the type becomes `*string` This seems to cause issues returning the string. @Gundersanne do you have any thoughts about what I'm doing wrong here?

ondrejbudai: You have to return `*string` in this case. I would help more but I need to see the error and the line of code that's causing it.

internal/v1/server.go
```diff
@@ -469,12 +476,25 @@ func (h *Handlers) ComposeImage(ctx echo.Context) error {
...
+
+	rawCR, err := json.Marshal(dbComposeRequest)
 	if err != nil {
 		return err
 	}
 
-	err = h.server.db.InsertCompose(composeResult.Id, idHeader.Identity.AccountNumber, idHeader.Identity.Internal.OrgId, rawCR)
+	var imageName string
+	if composeRequest.ImageName == nil {
+		imageName = ""
```
Gundersanne, line 492: default value of  a string is `""` so you can simplify to
```
if composeRequest.ImageName != nil {
     imageName = *....
}
```

internal/db/migrations/4_alter_table_add_image_name.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE composes ADD COLUMN image_name VARCHAR NOT NULL DEFAULT '';
```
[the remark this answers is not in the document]
jkozol: So, when I tried leaving the default value as NULL it creates issues when mapping from the db to a golang struct. I get error messages of can't map NULL to *string. Do you how I could change the `.Scan()` so that we can map from NULL to *string. Everything I could find online recommended using packages to handle typing between the db and the golang structs.

cmd/image-builder-db-test/main_test.go
```diff
@@ -257,7 +266,7 @@ func testCountGetComposesSince(t *testing.T) {
 	require.Equal(t, job1, composes[0].Id)
 
 	// correct ordering (recent first)
-	composes, count, err = d.GetComposes(ANR3, fortnight * 2, 100, 0)
+	composes, count, err = d.GetComposes(ANR3, fortnight*2, 100, 0)
```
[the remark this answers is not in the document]
jkozol: This is coming from gofmt. I can make a separate commit for the style update.

internal/v1/api.yaml
jkozol: I guess my confusion about returning *string is that wouldn't that return a pointer to the user?

internal/v1/server.go
Gundersanne: Don't derefence it, if it's non-required in the response, it won't be specified.

And I think the UI can then just check, if 'name' is in the response -> use the name, otherwise fallback to the uuid

Gundersanne: I wouldn't ever dereference. Just store the ComposeEntry with a pointer directly.

internal/db/migrations/4_alter_table_add_image_name.up.sql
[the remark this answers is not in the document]
Gundersanne: jackc/pgx handles this just fine if you change ComposeEntry.ImageName to `*string`

jkozol:
Ok. I made all the changes to the `*string` type for the name. Everything seems to work as expected now :)

internal/v1/server.go
```diff
@@ -469,12 +476,18 @@ func (h *Handlers) ComposeImage(ctx echo.Context) error {
 		return err
 	}
 
-	rawCR, err := json.Marshal(composeRequest)
+	dbComposeRequest := DBComposeRequest{
```
Gundersanne, line 479: Why the separate object?  The idea is to just throw the request in there as it comes in.

cmd/image-builder-db-test/main_test.go
```diff
@@ -110,9 +110,18 @@ func testMigration(t *testing.T) {
 
 	migrateOneStep(t) // migrate to step 3
 
+	migrateOneStep(t) // migrate to step 4
+
+	IMGNM := "MyImageName"
```
Gundersanne, line 115: Nitpick but use regular variable naming here, just imageName

internal/v1/server.go
jkozol: I made it a separate object to not store the image name in duplicate. I am happy to pass it all in inside of the request but then I don't think we need to store it in a separate column in the database.

cmd/image-builder-db-test/main_test.go
jkozol: Will update.

internal/v1/server.go
Gundersanne: The reason to store the request in the db is just so users can replay/recreate it easily later. Attributes that we want to maybe index on, and in future potentially search on, and are defined in the api spec (request is just of type {}), we should store those in a proper column imo.

jkozol: Makes a lot of sense. Fixed.

Gundersanne (commented):
Can you also add a test case in the db/migration test where you insert a compose with 'nil' as the imageName? And then verify it's nil coming out.

cmd/image-builder-db-test/main_test.go
```diff
@@ -125,6 +139,12 @@ func testMigration(t *testing.T) {
 	_, count, err = d.GetComposes(ANR2, fortnight, 100, 0)
 	require.NoError(t, err)
 	require.Equal(t, 1, count)
+
+	// check that the image names are as expected
+	composeEntries, count, err := d.GetComposes(ANR3, fortnight, 100, 0)
+	var nilString *string = nil
```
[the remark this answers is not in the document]
Gundersanne: Use `require.Nil` https://pkg.go.dev/github.com/stretchr/testify/require#Nil, no need for the variable here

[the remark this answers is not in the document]
jkozol: @Gundersanne fixed

cmd/image-builder-db-test/main_test.go
```diff
@@ -110,9 +110,23 @@ func testMigration(t *testing.T) {
...
+	// inserting empty image_name should succeed
+	insert = "INSERT INTO composes(job_id, request, created_at, account_number, org_id, image_name) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4, $5)"
+	_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", ANR3, ORGID1, nil)
+	require.NoError(t, err)
+
 	// Verify that after migration step 3 adding a compose request to the db requires a non empty account number.
 	d, err := db.InitDBConnectionPool(connStr(t))
-	err = d.InsertCompose(uuid.New().String(), "", ORGID1, []byte("{}"))
+	err = d.InsertCompose(uuid.New().String(), "", ORGID1, &imageName, []byte("{}"))
 	require.Error(t, err)
```
Gundersanne, line 130: Ah I only now noticed this, these 4 lines should move up to in between migration 3 and 4 :/

jkozol: Moved

cmd/image-builder-db-test/main_test.go
```diff
@@ -112,9 +112,23 @@ func testMigration(t *testing.T) {
 
 	// Verify that after migration step 3 adding a compose request to the db requires a non empty account number.
 	d, err := db.InitDBConnectionPool(connStr(t))
-	err = d.InsertCompose(uuid.New().String(), "", ORGID1, []byte("{}"))
+	err = d.InsertCompose(uuid.New().String(), "", ORGID1, nil, []byte("{}"))
```
Gundersanne, line 115: Because this would be using current defined Insertcomposer against a migration-3 db, it needs to be transformed into a direct query:

```
insert = "INSERT INTO composes(job_id, request, created_at, account_id, org_id) VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4)"
_, err = conn.Exec(context.Background(), insert, uuid.New().String(), "{}", "", ORGID1)
require.Error(t, err)
```

Sorry should've mentioned that. Last thing and it should be good :)

jkozol: ahh makes sense.

Gundersanne (approved):
LGTM, thank you!

PR #326, opened 2022-02-16
add test tags to cloud resources

jabia99:
scheduled cloud cleaner (running daily on osbuild-composer repo), removes old
enough and tagged resources. Adding the "gitlab-ci-test" tag to all
testing resources so they can get cleaned if the cleanup function fails.

### DIFF
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -348,7 +348,20 @@ function installClientAzure() {
   $AZURE_CMD version
 }
 
+source /etc/os-release
+
+CI="${CI:-false}"
+if [[ "$CI" == true ]]; then
+  DISTRO_CODE="${DISTRO_CODE:-${ID}-${VERSION_ID//./}}"
+  TEST_ID="$DISTRO_CODE-$ARCH-$CI_COMMIT_BRANCH-$CI_BUILD_ID"
+else
+  # if not running in Jenkins, generate ID not relying on specific env variables
+  TEST_ID=$(uuidgen);
+fi
+
 function createReqFileAzure() {
+  AZURE_IMAGE_NAME="image-$TEST_ID"
+
   cat > "$REQUEST_FILE" << EOF
 {
   "distribution": "$DISTRO",
@@ -361,7 +374,8 @@ function createReqFileAzure() {
         "options": {
           "tenant_id": "${AZURE_TENANT_ID}",
           "subscription_id": "${AZURE_SUBSCRIPTION_ID}",
-          "resource_group": "${AZURE_RESOURCE_GROUP}"
+          "resource_group": "${AZURE_RESOURCE_GROUP}",
+	  "image-name": "${AZURE_IMAGE_NAME}"
         }
       }
     }
@@ -473,12 +487,19 @@ function Test_verifyComposeResultAWS() {
   REGION=$(echo "$UPLOAD_OPTIONS" | jq -r '.region')
   [[ "$REGION" = "$AWS_REGION" ]]
 
+  # Tag image and snapshot with "gitlab-ci-test" tag
+  $AWS_CMD ec2 create-tags \
+    --resources "${AMI_IMAGE_ID}" \
+    --tags Key=gitlab-ci-test,Value=true
+
   # Create key-pair
   $AWS_CMD ec2 create-key-pair --key-name "key-for-$AMI_IMAGE_ID" --query 'KeyMaterial' --output text > keypair.pem
   chmod 400 ./keypair.pem
 
   # Create an instance based on the ami
-  $AWS_CMD ec2 run-instances --image-id "$AMI_IMAGE_ID" --count 1 --instance-type t2.micro --key-name "key-for-$AMI_IMAGE_ID" > "$WORKDIR/instances.json"
+  $AWS_CMD ec2 run-instances --image-id "$AMI_IMAGE_ID" --count 1 --instance-type t2.micro \
+	  --key-name "key-for-$AMI_IMAGE_ID" \
+	  --tag-specifications 'ResourceType=instance,Tags=[{Key=gitlab-ci-test,Value=true}]' > "$WORKDIR/instances.json"
   AWS_INSTANCE_ID=$(jq -r '.Instances[].InstanceId' "$WORKDIR/instances.json")
 
   $AWS_CMD ec2 wait instance-running --instance-ids "$AWS_INSTANCE_ID"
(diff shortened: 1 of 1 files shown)


test/cases/api.sh
```diff
@@ -524,8 +534,12 @@ function Test_verifyComposeResultGCP() {
     --zone="$GCP_REGION-a" \
     --image-project="$GCP_IMAGE_BUILDER_PROJECT" \
     --image="$GCP_IMAGE_NAME" \
+    --labels=gitlab-ci-test=true \
     --metadata-from-file=ssh-keys="$GCP_SSH_METADATA_FILE"
   HOST=$($GCP_CMD compute instances describe "$GCP_INSTANCE_NAME" --zone="$GCP_REGION-a" --format='get(networkInterfaces[0].accessConfigs[0].natIP)')
+  
+  echo $GCP_INSTANCE_NAME
+  echo $HOST
```
Gundersanne, line 542: These should be clear from the `-x` flag no?

jabia99: That's true, I just discovered the test fails before creating the instance

Gundersanne (dismissed):

test/cases/api.sh
```diff
@@ -516,6 +523,9 @@ function Test_verifyComposeResultGCP() {
 
   echo "${SSH_USER}:$(cat "$GCP_SSH_KEY".pub)" > "$GCP_SSH_METADATA_FILE"
 
+  # add "gitlab-ci-test" label to the image
+  # $GCP_CMD compute images add-labels "$GCP_IMAGE_NAME" --labels=gitlab-ci-test=true
+
```
Gundersanne, line 528: redundant comment?

test/cases/api.sh
```diff
@@ -524,8 +534,10 @@ function Test_verifyComposeResultGCP() {
     --zone="$GCP_REGION-a" \
     --image-project="$GCP_IMAGE_BUILDER_PROJECT" \
     --image="$GCP_IMAGE_NAME" \
+    --labels=gitlab-ci-test=true \
     --metadata-from-file=ssh-keys="$GCP_SSH_METADATA_FILE"
   HOST=$($GCP_CMD compute instances describe "$GCP_INSTANCE_NAME" --zone="$GCP_REGION-a" --format='get(networkInterfaces[0].accessConfigs[0].natIP)')
+
```
Gundersanne, line 540: nitpick, rogue newline

test/cases/api.sh
jabia99: We need this line to add the label to the image. I'm doing some debugging, as for some reason, that line gives me the next error: 
https://gitlab.com/osbuild/ci/image-builder/-/jobs/2117306772#L2971

Gundersanne: Ah IC, yea this image is made by the stage deployment, and as such isn't in the cockpituous account. It's located in the account used for the service, these get automatically cleaned up after 14 days.

jabia99: Ohh, that's why. Thanks!!

jabia99:
@jrusz do you think scheduled cloud cleaner will clean properly azure images from this test? I've read your code but you are more familiar with it

jrusz (changes requested):
> @jrusz do you think scheduled cloud cleaner will clean properly azure images from this test? I've read your code but you are more familiar with it

No it won't it's using the cloud-cleaner written in GO which expects a specific names of the resources ( https://github.com/osbuild/osbuild-composer/blob/main/internal/boot/azuretest/azure.go#L187 and https://github.com/osbuild/osbuild-composer/blob/main/internal/boot/azuretest/azure.go#L104 ) . The `api.sh` test here is missing something like this https://github.com/osbuild/osbuild-composer/commit/d927bfbd30d8394e0df4f2965fffb587ec2c40f0 ... It only cleans the resources in the resource group that matches the `$TEST_ID` prepended with name of the resource like image,nsg etc... @sanne also added some code to clean all storage accounts I think that one is good, but it currently won't clean anything left by this test.

jabia99:
> but it currently won't clean anything left by this test.
What do you think is the best approach here? Change the name of the generated images? or add labels to testing azure images and rewrite scheduled cloud cleaner so removes tagged resources?

jrusz:
> > but it currently won't clean anything left by this test.
> What do you think is the best approach here? Change the name of the generated images? or add labels to testing azure images and rewrite scheduled cloud cleaner so removes tagged resources?

I'd just name all the resources in this `api.sh` test the same way they're named in `osbuild-composer` repository and make sure it's using the same `AZURE_RESOURCE_GROUP` and any other vars that will affect the final destination of the resources.

jrusz (changes requested):
Looks good, just one tiny nitpick. Did you test it in any way? 

Also I wonder if we want to be able to run this easily manually, in that case we could use a similar condition for `TEST+ID` like we have in `osbuild-compsoer` repo https://github.com/osbuild/osbuild-composer/blob/main/test/cases/api.sh#L437 It's probably just a nice to have...

jabia99:
> Looks good, just one tiny nitpick. Did you test it in any way?

For testing, I manually went to the cloud and saw that the resources where tagged, (in GCP is not needed as the bucket they are uploaded get automatically cleaned. For azure it the name had the `TEST_ID` format). Do you think I should test it a bit more? I could remove the cleanup part of `api.sh` test and check the resources get removed.

I didn't test it much more as the code is the same as in osbuild-composer.

test/cases/api.sh
```diff
@@ -348,7 +348,19 @@ function installClientAzure() {
   $AZURE_CMD version
 }
 
+source /etc/os-release
+
+if [[ "$CI" == true ]]; then
```
jrusz, line 353: Uh since `set -u` is set it's going to fail outside of CI cause that variable will not be set, you'd need something like `CI="${CI:-false}"` for it to work :)

jrusz (changes requested):
> For testing, I manually went to the cloud and saw that the resources where tagged, (in GCP is not needed as the bucket they are uploaded get automatically cleaned. For azure it the name had the `TEST_ID` format). Do you think I should test it a bit more? I could remove the cleanup part of `api.sh` test and check the resources get removed.
> 
> I didn't test it much more as the code is the same as in osbuild-composer.

Yeah right, I guess it's fine then. One more tiny nitpick since you already wen through the effort to add the condition for `TEST_ID` :)

test/cases/api.sh
jabia99: Uups, that makes sense

jrusz (dismissed):
LGTM

jrusz (approved):

PR #331, opened 2022-03-02
Image type tags

Gundersanne:


distributions/centos-8/centos-8.json
```diff
@@ -7,14 +7,22 @@
...
       "rhsm": false
     }, {
+      "id": "extras",
       "baseurl": "http://mirror.centos.org/centos/8-stream/extras/x86_64/os/",
       "rhsm": false
+    }, {
+      "id": "google-compute-engine",
+      "baseurl": "https://packages.cloud.google.com/yum/repos/google-compute-engine-el8-x86_64-stable",
+      "rhsm": false,
+      "image_type_tags": ["gcp"]
```
thozza, line 25: We need to define 
```ini
[google-cloud-sdk]
name=Google Cloud SDK
baseurl=https://packages.cloud.google.com/yum/repos/cloud-sdk-el8-x86_64
enabled=1
gpgcheck=1
repo_gpgcheck=0
gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg
    https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
```
as well, because the GCE image installs the GCP SDK by default. The reason is that the current RHEL image built by Google contains it and the goal was to replicate the image for now.

distributions/centos-9/centos-9.json
```diff
@@ -7,9 +7,11 @@
...
     "image_types": [ "ami", "vhd", "aws", "gcp", "azure", "edge-commit", "edge-installer", "edge-container", "rhel-edge-commit", "rhel-edge-installer", "guest-image", "image-installer", "vsphere" ],
     "repositories": [{
+      "id": "baseos",
       "baseurl": "http://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/",
       "rhsm": false
     }, {
+      "id": "appstream",
       "baseurl": "http://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/",
       "rhsm": false
     }]
```
thozza, line 17: The Google repos should be defined for C9S as well.

internal/distribution/distribution.go
```diff
@@ -0,0 +1,161 @@
...
+		df, err := ReadDistribution(distsDir, f)
+		if err != nil {
+			return nil, err
+		}
+		distros = append(distros, df.Distribution)
+	}
+	return distros, nil
+}
+
+func RepositoriesForImage(distsDir, distro, arch string) ([]Repository, error) {
```
thozza, line 118: nitpick: I find the function name misleading. I would expected something like `RepositoriesForArchitecture` based on what it does.

Gundersanne: makes sense!

distributions/centos-8/centos-8.json
Gundersanne: done! can't believe i missed that many :facepalm:

internal/distribution/distribution.go
```diff
@@ -0,0 +1,161 @@
...
+
+type Distributions []DistributionItem
+
+type DistributionFile struct {
+	ModulePlatformID string           `json:"module_platform_id"`
+	Distribution     DistributionItem `json:"distribution"`
+	ArchX86          *X86_64          `json:"x86_64,omitempty"`
+}
+
+type X86_64 struct {
```
thozza, line 32: super-"nice to have" nitpick: Sorry that I missed this initially. This structure could have been named e.g. `Architecture`, because I expect that it could be reused as is, for any additional architectures supported by image-builder.

So now the `DistributionFile` would look like 
```go
type DistributionFile struct {
	ModulePlatformID string           `json:"module_platform_id"`
	Distribution     DistributionItem `json:"distribution"`
	ArchX86          *Architecture    `json:"x86_64,omitempty"`
}
```
and in the future could look like
```go
type DistributionFile struct {
	ModulePlatformID string           `json:"module_platform_id"`
	Distribution     DistributionItem `json:"distribution"`
	ArchX86          *Architecture    `json:"x86_64,omitempty"`
        ArchAarch64      *Architecture    `json:"aarch64,omitempty"`
}
```

Not a big deal, I just wanted to point this out.

thozza (approved):
LGTM, thanks for the changes!

internal/distribution/distribution.go
Gundersanne: https://github.com/osbuild/image-builder/issues/339

thozza: 🙇

PR #333, opened 2022-03-04
schutzbot: update slack notificarion

jabia99:
slack notification for load test creates a message with the full link to
the results. Use markdown sintax to shorten the link. Also add the rest
of QE team to the message.

### DIFF
--- a/schutzbot/slack_notification.sh
+++ b/schutzbot/slack_notification.sh
@@ -2,7 +2,7 @@
 
 set -eux
 
-MESSAGE="\"Load testing on image builder finished with status *$1* $2 \n QE: @jabia \n Link to results: $CI_PIPELINE_URL \""
+MESSAGE="\"<$CI_PIPELINE_URL|Load testing on image builder> finished with status *$1* $2 \n QE: @atodorov, @jrusz, @jabia \""
 
 curl \
     -X POST \


schutzbot/slack_notification.sh
```diff
@@ -2,7 +2,7 @@
 
 set -eux
 
-MESSAGE="\"Load testing on image builder finished with status *$1* $2 \n QE: @jabia \n Link to results: $CI_PIPELINE_URL \""
+MESSAGE="\"Load testing on image builder finished with status *$1* $2 \n <$CI_PIPELINE_URL|Results> - QE: @atodorov, @jrusz, @jabia \""
```
jrusz, line 5: Hm. don't you think it would be better to just put the link under the main text? Just to make it shorter, we don't need another line that just says "Results"

```suggestion
MESSAGE="\"<$CI_PIPELINE_URL|Load testing on image builder> finished with status *$1* $2 \n QE: @atodorov, @jrusz, @jabia \""
```

jrusz (commented):
I've put a suggestion, what do you think about it?

schutzbot/slack_notification.sh
jabia99: I actually removed one of the lines. But yeah, that would look much better!

jrusz (approved):
LGTM

PR #335, opened 2022-03-10
internal/config: Get rid of *string/string distinction

Gundersanne:


Gundersanne:
We already ignore this for some config options, so let's just get rid of it.

Gundersanne:
This is blocked on #334 for the spec update

ondrejbudai (approved):
&nice!

PR #337, opened 2022-03-11
Client credentials grant

Gundersanne:


teg:
Please explain a bit :) Otherwise looking good!

Gundersanne:
Added some extra info in the 3rd commit:

```

    This will allow us to use the service accounts which work against
    identity.api.openshift.com, which are much easier to manage and have the
    added benefit of not expiring automatically/expiring instantly when we
    want them to.
```

PR #343, opened 2022-03-21
templates: Stop using offline token

Gundersanne:


Gundersanne:
Will need a followup for CI. But I'd prefer to get image-builder to post jobs to the new tenant first; so those jobs don't get lost in the queue.

ondrejbudai (approved):

PR #344, opened 2022-03-22
Fix rhel 85 appstream pkglist

Gundersanne:


distributions/rhel-85/rhel-85.json
```diff
@@ -11,7 +11,7 @@
       "baseurl": "https://cdn.redhat.com/content/dist/rhel8/8.5/x86_64/baseos/os",
       "rhsm": true
     }, {
-      "id": "baseos",
+      "id": "appstream",
```
teg, line 14: I used this was the bug? For the future, could you explain it in the commit message?

internal/distribution/distribution_test.go
```diff
@@ -56,12 +56,30 @@ func TestAvailableDistributions(t *testing.T) {
...
 
 func TestFindPackages(t *testing.T) {
-	pkgs, err := FindPackages("../../distributions", "centos-8", "x86_64", "ssh")
+	pkgs, err := FindPackages("../../distributions", "centos-8", "x86_64", "vim")
 	require.NoError(t, err)
-	require.Greater(t, len(pkgs), 0)
-	for _, p := range pkgs {
-		require.Contains(t, p.Name, "ssh")
-	}
+	require.ElementsMatch(t, []Package{
```
teg, line 61: Makes sense, but explanation in the commit message would make it perfecter.

teg (approved):
Looks good, minor nits inline not a blocker.

distributions/rhel-85/rhel-85.json
Gundersanne: done!

internal/distribution/distribution_test.go
Gundersanne: done!

PR #348, opened 2022-03-26
org-id as primary tenant identifier

teg:
Move from using `account_number` to `org_id` as tenant identifier. For the time being, we still populate the database with both, for backwards compatibility.

Currently all our tenants have both `account_number` and `org_id` set in the identity header, but that is about to change at the platform level, so this prepares us.

Every `User` account (all of our potential tenants) will continue to have an `org_id` associated with them, but `account_numbers` are only associated with accounts that are associated to an EBS account, which happens when a subscription is created for the organization. Accounts that are not associated with EBS accounts are refer to as "anemic accounts". At the moment console.redhat.com does not support anemic accounts, and all access is gated on EBS accounts, but in the future that requirement will be dropped, allowing users access to console.redhat.com (and image builder) before they have any subscriptions.

This PR does a few things:
 - start relying on redhatinsights/identity identity header validation (still enforcing `account_numbers`, but ready to follow their changes)
 - make `org_id` required in our database and use that to make tenancy decisions (should be a noop)
 - gate access to RHEL content on having an EBS account (a noop until the restriction is lifted)

Follow-ups:
 - drop `account_numbers` entirely when we have redeployed and nothing relies on it
 - update the `identity` library when the policy has been changed to no longer require `account_number`
 - use the `rhel: is_enabled` property from the identity header once https://github.com/RedHatInsights/identity/pull/3 has been merged

Blockers:
- [ ] move quotas from account_numbers to org_ids in app-interface (maybe: just define default quotas?)


teg:
See discussion here: https://github.com/RedHatInsights/identity/pull/2

teg:
@lphiri, could you give this a quick sanity check, just to see if I have understood the consoleDot parts correctly?

lphiri:
@teg the consoleDot parts look ok.  Do you have a plan for filling in org IDs for existing DB data when you do the migration?   We will be providing an a translation API to help with that. There is actually a golang helper package you may be able to use : https://github.com/RedHatInsights/tenant-utils

teg:
> @teg the consoleDot parts look ok.

Thanks!

> Do you have a plan for filling in org IDs for existing DB data when you do the migration? We will be providing an a translation API to help with that. There is actually a golang helper package you may be able to use : https://github.com/RedHatInsights/tenant-utils

Ah, that's great to know! We happened to already record the `org_id` in the database (I should have mentioned that in the PR description), so we should be good (is that right @Gundersanne? I'm assuming any old entries without `org_id` have been garbage collected by now...)

teg:
@Gundersanne I'm unlikely to have time to finish this up, feel free to take it over / consider it an RFE. Happy to discuss it if you'd like.

Gundersanne:
re: quotas

They're currently defined in app-intf. There's 2 accounts with non-standard quota both in stage (CI/QE accounts).

teg:
> re: quotas
> 
> They're currently defined in app-intf. There's 2 accounts with non-standard quota both in stage (CI/QE accounts).

Considering this is in staging, my suggestion was to up the quota there for everyone (I think we support `*`?), so we would not need a flag day when replacing the account numbers with org_ids.

Gundersanne:
I made the default on stage 10k, so should be fine. I think we're ready on this?

internal/v1/server.go
```diff
@@ -870,3 +848,22 @@ func (s *Server) HTTPErrorHandler(err error, c echo.Context) {
...
+			return err
+		}
+
+		if idh.Identity.Type == "Associate" {
+			// Associate account types are not guaranteed to have an associated org_id, these accounts
+			// should not be able to access image-builder as long as we don't explicitly enable turnpike
+			// access, or another such service forwards them to us. Explicitly reject such accounts for
+			// now.
+			return echo.NewHTTPError(http.StatusBadRequest, "unsupported account type: 'Associate'")
+		}
```
Gundersanne, line 865: So I wonder if we can't just introduce a dummy org_id value here for associates?  Like if suddenly a bunch of internal people lose access, that could be bad no?

teg: This would require explicit support beyond just the org_id. These accounts are not used to access image builder today (basically: not what it looks like). Happy to discuss though.

teg: This would require explicit support beyond just the org_id. These accounts are not used to access image builder today (basically: not what it looks like). Happy to discuss though.

PR #352, opened 2022-04-13
v1: limit the number of image requests in the spec itself 

ondrejbudai:
We have always supported only one image request per a compose. This is not
enforced on the openapi spec level though, just on the Go code level.
Let's add the limits also to the spec file to be more explicit.

As passing more/no requests was never possible, this is not a change.


Gundersanne (approved):
Good idea, thank you!

ondrejbudai:
Had to actually modify the test for this. Also, I removed the now redundant check from `server.go` since this is now being caught on the openAPI level.

Gundersanne (approved):

Gundersanne:
GCP is out of resources again. This is fine.

PR #353, opened 2022-04-14
v1: Adapt to new composer spec

jrusz:


jrusz:
@Gundersanne do you have any idea what's happening to the Edge test? It's like it's trying to download the edge commit but it takes forever and the times out...

Gundersanne:
> @Gundersanne do you have any idea what's happening to the Edge test? It's like it's trying to download the edge commit but it takes forever and the times out...

The edge test is extremely flaky and has multiple pain points. Need to get on fixing it or maybe just removing it at some point.

PR #359, opened 2022-05-04
db: add max image name length constraint

jkozol:
Image Names must be 100 characters or less. Existing entries are not changed but new entries must pass this check.


Gundersanne (changes requested):
I think this migration will fail if there's any entry with more than 100 characters (can you add a test for that as well?)

As part of the migration, can you go through the table and updated any entries longer than 100 chars?

jkozol:
> This migration will fail if there's any entry with more than 100 characters (can you add a test for that as well?)
>
> As part of the migration, can you go through the table and update/truncate any entries longer than 100 chars?

Since I set `Not Valid` the migration won't validate any existing entries. So, even if there are names with more than 100 characters the migration should still succeed. I think it would be better to maintain the data users have already entered instead of truncating it.

Gundersanne:
> > This migration will fail if there's any entry with more than 100 characters (can you add a test for that as well?)
> > As part of the migration, can you go through the table and update/truncate any entries longer than 100 chars?
> 
> Since I set `NOT VALID` the migration won't validate any existing entries. So, even if there are names with more than 100 characters the migration should still succeed. I think it would be better to maintain the data users have already entered instead of truncating it.

Ah nice, I missed that! Perfect, then lgtm, thank you.

Gundersanne (approved):

PR #362, opened 2022-05-14
Filter errors

Gundersanne:


teg:
Might want to run the "open a support case" advice by CEE.

Gundersanne:
> Might want to run the "open a support case" advice by CEE.

Yea probably need a different string there, can also say unknown or something.

teg:
I like the idea though. Should just get their ack and make sure we agree on what info to include (and when to give this advice, probably should not be on a catch all)

Gundersanne:
> I like the idea though. Should just get their ack and make sure we agree on what info to include (and when to give this advice, probably should not be on a catch all)

Yea, I'm just wondering what the default error should be.

PR #365, opened 2022-05-24
v1: limit the architectures to x86_64 in the API

ondrejbudai:
We don't support anything else right now so let's make it very explicit.


Gundersanne (dismissed):

Gundersanne (dismissed):

Gundersanne:
Just pushed a prepare-source commit real quick, makes sure the composer spec is up to date.

Gundersanne (approved):

PR #369, opened 2022-05-27
Add kube-linter check

Gundersanne:
Example runs: https://github.com/Gundersanne/image-builder/actions/workflows/tests.yml.

The 'break template' commits are commits where the check was made to fail on purpose.


ondrejbudai (approved):
More linting, yay!

PR #373, opened 2022-06-01
Add allow list

lucasgarfield:
Adds an allow list for distributions that verifies that a compose
request's distribution is valid for the requesting account. Validation
is done using the account's orgId, which is cross referenced against an
allow list file pointed to by the ALLOW_FILE environment variable. If no
ALLOW_FILE environment variable is set, a set of fallback default values
is used.


distributions/fedora-34/fedora-34.json
```diff
@@ -0,0 +1,29 @@
+{
+  "module_platform_id": "platform:el8",
+  "distribution": {
+    "name": "fedora-34",
+    "description": "Fedora 34"
+  },
+  "x86_64": {
+    "image_types": [ "aws", "gcp", "azure", "ami", "vhd" ],
+    "repositories": [{
```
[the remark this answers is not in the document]
lucasgarfield: Will add Fedora in separate PR.

internal/common/allow.go
```diff
@@ -0,0 +1,81 @@
...
+// variable. In the case that the allowFile parameter is an empty string (as would occur if the ALLOW_FILE
+// environment variable was not set) then a fallback default allow list is used instead.
+func CheckAllow(orgId string, distribution string, allowFile string) (bool, error) {
+	if allowFile == "" {
+		return isAllowed(distribution, defaultAllowList)
+	}
+
+	var allowList map[string]map[string]bool
+
+	jsonFile, err := os.Open(filepath.Clean(allowFile))
```
ondrejbudai, line 60: We should read the file only once - ideally during IB's startup. This way, if the allow-list file is malformed, image-builder will refuse to start (which is better than returning an error for each request).

ondrejbudai: Also: Files must be always closed so we should have `defer jsonFile.Close()` after `os.Open()` is called and `err != nil`.

internal/common/allow.go
```diff
@@ -0,0 +1,81 @@
...
+	if _, error := err.(*os.PathError); error {
+		return false, fmt.Errorf("No allow file found at %s", allowFile)
+	} else {
```
ondrejbudai, line 63: Extra `else`, it's actually not needed.

internal/common/allow.go
```diff
@@ -0,0 +1,81 @@
...
+// environment variable was not set) then a fallback default allow list is used instead.
+func CheckAllow(orgId string, distribution string, allowFile string) (bool, error) {
+	if allowFile == "" {
+		return isAllowed(distribution, defaultAllowList)
+	}
+
+	var allowList map[string]map[string]bool
+
+	jsonFile, err := os.Open(filepath.Clean(allowFile))
+	if _, error := err.(*os.PathError); error {
```
ondrejbudai, line 61: The case when `err` isn't `*os.PathError` also needs to be handled.

internal/common/allow.go
```diff
@@ -0,0 +1,81 @@
...
+		return false, fmt.Errorf("No allow file found at %s", allowFile)
+	} else {
+		rawJsonFile, err := ioutil.ReadAll(jsonFile)
```
ondrejbudai, line 64: `io.ReadAll` is preferred, see https://pkg.go.dev/io/ioutil

internal/common/allow.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+// the ALLOW_FILE file must conform to the following schema:
+// {
```
ondrejbudai, line 12: This format is a great start! My only concern is that whenever we add a new distribution, we would have to change the default allow list. Not much work, but surely a bit annoying. I gave this some thought and I wonder if we can add defaults to distro jsons themselves:

Let's imagine adding `restrictedAccess: true` into https://github.com/osbuild/image-builder/blob/main/distributions/rhel-90/rhel-90.json (well, not in RHEL but in Fedora). If a user wants to build a restricted distribution, we would check their allowlist to see if they explicitly have the approval to build such a distribution.

How does it sound? Happy to diccuss more.

internal/common/allow.go
```diff
@@ -0,0 +1,81 @@
...
+// 	"default":{
+//		"rhel-84": true,
+//		"rhel-85": true,
+//		"rhel-86": true,
+//		"rhel-90": true,
+//		"centos-8": true,
+//		"centos-9": true,
+// 	},
+// 	"000000":{
```
ondrejbudai, line 21: I wonder if accepting glob patterns would make this more convenient:
```
{
  "123456": [
    "fedora-*"
  ]
}
```

ondrejbudai (changes requested):
This is awesome! I like that it's documented and fully functional (hopefully, no test coverage, I will get to this later).

I put my observations inline, my main thoughts are: reading the file must be elsewhere and we should make sure that this doesn't add much burden when defining new distros (therefore I proposed restricted distributions and globs in the allow list). Otherwise, these are just minor Go things that you didn't know.

The other big missing thing is testing - we should try to cover everything as much as possible, unit testing is fine.

Anyway, thanks for your work, and I'm happy to discuss everything in more depth if needed. :)

internal/common/allow.go
lucasgarfield: Sounds great, everything mentioned above has been implemented!

internal/common/allow.go
lucasgarfield: Yep, and done!

internal/common/allow.go
lucasgarfield: In fact os.Open can *only* return an *os.PathError, so I changed this to a more general `if err != nil` block with the same error message.

internal/common/allow.go
lucasgarfield: Done!

internal/common/allow.go
lucasgarfield: Done, but now we are failing the gosec check. I suppose the solution is to just close the file ourselves when we can, without defer?
```
[/github/workspace/internal/common/allow.go:36] - G307 (CWE-703): Deferring unsafe method "Close" on type "*os.File" (Confidence: HIGH, Severity: MEDIUM)
    35: 	}
  > 36: 	defer jsonFile.Close()
    37: 
```

internal/common/allow.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to unmarshal allow file %q: %s", allowFile, err.Error())
+	}
+
+	return allowList, nil
+}
+
+func isAllowed(orgId, distro string, allowList AllowList) (bool, error) {
+	for _, allowedDistro := range allowList[orgId] {
+		match, err := path.Match(allowedDistro, distro)
```
ondrejbudai, line 53: Please put a comment here why we are using `path.Match` here.

internal/common/allow.go
lucasgarfield: Will use this workaround: https://github.com/securego/gosec/issues/512#issuecomment-675286833

internal/common/allow.go
```diff
@@ -0,0 +1,84 @@
...
+// 	"000000": ["fedora-*"]
+//  "000001": ["fedora-34", "fedora-35", "fedora-36"]
+//  "000002": []
+// }
+func LoadAllowList(allowFile string, logger *logrus.Logger) (AllowList, error) {
+	if allowFile == "" {
+		return AllowList{}, nil
+	}
+
+	var allowList AllowList
```
ondrejbudai, line 31: We can put this before the line 48.

internal/common/allow_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	t.Run("orgId not in allow list (forbidden)", func(t *testing.T) {
+		actual, _ := mockAllowList.isAllowed("123456", "fedora-36")
+		expected := false
+		require.Equal(t, expected, actual)
+	})
+}
+
+func TestLoadAllowList(t *testing.T) {
+	var mockLogger, err = logger.NewLogger("DEBUG", "", "", "", "")
```
ondrejbudai, line 34: `var mockLogger, err =` => `mockLogger, err :=`

ondrejbudai (commented):
Nice, we are getting really close, this is exciting, thank you! :)

ondrejbudai (approved):

ondrejbudai (approved):
This is great, thank you!

PR #374, opened 2022-06-03
v1: Promotion spec changes

Gundersanne:
I'd introduce this concept under v1 first in a backward compatible way, together with blueprints.

Once we have all 3 concepts (composes, promotions, blueprints) working we can start designing the v2 spec imo.

This PR is just a wip discussion kinda thing to see what things will look like to support promotions.

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -75,6 +75,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Architectures'
+
   /composes:
     get:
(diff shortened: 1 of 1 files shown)


teg:
Do we (eventually) want to expose (re)sharing in a similar way? Would it be a type of promotion or separate to it?

Gundersanne:
> Do we (eventually) want to expose (re)sharing in a similar way? Would it be a type of promotion or separate to it?

Type of promotion I think on the ibcrc level no? And then each promotion translates to a 'put' job in composer.

PR #375, opened 2022-06-15
templates: remedy iqe-trigger template validation errors

jrusz:
AppSRE found some issues with this template which are resolved by this
commit.

### DIFF
--- a/templates/iqe-trigger.yml
+++ b/templates/iqe-trigger.yml
@@ -7,11 +7,15 @@ objects:
   kind: Job
   metadata:
     name: image-builder-iqe-trigger
+    annotations:
+      ignore-check.kube-linter.io/no-liveness-probe
+      ignore-check.kube-linter.io/no-readiness-probe
   spec:
     backoffLimit: 5
     template:
       spec:
         restartPolicy: Never
+        serviceAccountName: image-builder
         containers:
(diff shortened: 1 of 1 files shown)


Gundersanne (dismissed):
thanks!

jrusz:
I had to make one modification, apparently a kubernetes job won't be triggered if a job with the same name exists in a namespace. `generateName` should help with this.

jrusz:
Had to use a random parameter becuase `oc apply` doesn't working with `generateName`, should be good now @Gundersanne

templates/iqe-trigger.yml
```diff
@@ -25,7 +29,16 @@ objects:
...
+    name: image-builder
+
 parameters:
   - name: WEBHOOK_URL
     value: ''
     required: true
+  - name: JOB_SUFFIX
+    description: "Random string used as job suffix"
+    generate: expression
+    from : "[a-zA-Z0-9]{12}"
```
Gundersanne, line 44: This should not be needed, the job should be replaced. Generate seems like it would add an ever increasing amount of jobs.

Gundersanne: Ah misunderstanding on my part, this is a one-off job run, we don't have to do this for the maintenance job because that's a cronjob which confused me.

PR #380, opened 2022-06-24
templates/composer: add fluentd sidecar

diaasami:


templates/image-builder.yml
```diff
@@ -168,13 +168,46 @@ objects:
...
+                secretKeyRef:
+                  name: splunk
+                  key: token
+                  optional: false
+            - name: SPLUNK_HEC_URL
+              valueFrom:
+                secretKeyRef:
+                  name: splunk
+                  key: url
+                  optional: false
```
Gundersanne, line 195: These havent been added to the insights namespaces yet, that needs to happen first.

Gundersanne (commented):
lgtm but we need to put the secrets in place first.

templates/image-builder.yml
diaasami: PR was just merged.

PR #383, opened 2022-07-01
Revert "templates/composer: add fluentd sidecar"

Gundersanne:
Reverts osbuild/image-builder#380


teg:
Out of curiosity: why?

Gundersanne:
> Out of curiosity: why?

This was blocking deployment and we really needed some other bits in quick. There were some issues with the environment still so once Diaa is back we can merge this again.

PR #390, opened 2022-07-08
v1: expose user customization

chloenayon:
Expose user account customization in image builder, and add tests that:
i. ensure that user customization is only permitted for installer image types
ii. image-builder compose requests are being translated correctly into the expected osbuild-composer compose requests (i.e. IB adds wheel group)

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -574,6 +574,24 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Filesystem'
+        users:
+          type: array
+          items:
+            $ref: '#/components/schemas/User'
+          description:
+            "list of users that a customer can add, also specifying their respective groups and SSH keys"
+    User:
+      type: object
+      required:
+        - name
+        - SSHKey
+      properties:
+        name:
+          type: string
+          example: "user1"
+        SSHKey:
+          type: string
+          example: "ssh-rsa AAAAB3NzaC1"
     Filesystem:
       type: object
       required:
--- a/internal/v1/server.go
+++ b/internal/v1/server.go
@@ -554,6 +554,11 @@ func (h *Handlers) ComposeImage(ctx echo.Context) error {
 		return err
 	}
 
+	// check here for installer types
+	if composeRequest.Customizations != nil && composeRequest.Customizations.Users != nil && !strings.Contains(string(imageType), "installer") {
+		return echo.NewHTTPError(http.StatusBadRequest, "User customization only applies to installer image types")
+	}
+
 	cloudCR := composer.ComposeRequest{
 		Distribution:   distroToStr(composeRequest.Distribution),
 		Customizations: buildCustomizations(composeRequest.Customizations),
@@ -758,6 +763,19 @@ func buildCustomizations(cust *Customizations) *composer.Customizations {
 			})
 		}
 		res.Filesystem = &fsc
+	}
+
+	if cust.Users != nil {
+		var users []composer.User
+		for _, u := range *cust.Users {
+			groups := &[]string{"wheel"}
+			users = append(users, composer.User{
+				Name:   u.Name,
+				Key:    &u.SSHKey,
+				Groups: groups,
+			})
+		}
+		res.Users = &users
 	}
 
 	return res
--- a/internal/v1/server_test.go
+++ b/internal/v1/server_test.go
@@ -60,6 +60,17 @@ func initQuotaFile(t *testing.T) (string, error) {
 	return file.Name(), nil
 }
 
+func makeUploadOptions(t *testing.T, uploadOptions interface{}) *composer.UploadOptions {
+	data, err := json.Marshal(uploadOptions)
+	require.NoError(t, err)
+
+	var result composer.UploadOptions
+	err = json.Unmarshal(data, &result)
+	require.NoError(t, err)
+
+	return &result
+}
+
 func startServerWithCustomDB(t *testing.T, url string, dbase db.DB, distsDir string, allowFile string) (*echo.Echo, *httptest.Server) {
 	var log = &logrus.Logger{
 		Out:       os.Stderr,
@@ -619,11 +630,11 @@ func TestComposeImage(t *testing.T) {
 				Users: &[]User{
 					{
 						Name:   "user-name0",
-						SSHKey: nil,
+						SSHKey: "",
 					},
 					{
 						Name:   "user-name1",
-						SSHKey: nil,
+						SSHKey: "",
 					},
 				},
 			},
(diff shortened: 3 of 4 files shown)


internal/v1/api.yaml
```diff
@@ -573,6 +573,26 @@ components:
...
+            $ref: '#/components/schemas/User'
+    User:
+      type: object
+      required:
+        - name
+      properties:
+        name:
+          type: string
+          example: "user1"
+        key:
```
teg, line 588: Should this be called `SSHKey` or something like that? Are we making sure we restrict what image types this applies to? I think this should only apply to installers, not to commits / base images. @gicmo thoughts?

internal/v1/api.yaml
```diff
@@ -573,6 +573,26 @@ components:
...
+          type: string
+          example: "userkey"
+        groups:
```
teg, line 591: Are there any gotchas with groups? What are the requirements for group support? Could the defaults be good enough so we don't need to expose this? I genuinely don't know, but if we look at how for instance amazon does user configuration I don't think it includes group customisation.

internal/v1/api.yaml
Gundersanne: +1 on a 400 for anything other than the installer types.

internal/v1/api.yaml
teg: Should the default be wheel? Is the default wheel? I feel I don't know as much as I should :)

teg: Right, I guess my question was more: should we allow selecting between sudo and non sudo users at all? Does it make sense to configure users without sudo access in this way?

internal/v1/api.yaml
Gundersanne: +1 on naming the sshkey too :)

internal/v1/api.yaml
```diff
@@ -574,6 +574,21 @@ components:
...
+      type: object
+      required:
+        - name
+      properties:
+        name:
+          type: string
+          example: "user1"
+        key:
+          type: string
+          example: "userkey"
```
ondrejbudai, line 591: Let's make this something that resembles an actual ssh key.

```suggestion
          example: "ssh-rsa AAAAB3NzaC1"
```

internal/v1/api.yaml
```diff
@@ -574,6 +574,21 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Filesystem'
+        users:
```
ondrejbudai, line 577: `description` would be sweet. I would love our API to be more documented.

internal/v1/api.yaml
teg: @gicmo thoughts on this?

internal/v1/server.go
```diff
@@ -552,6 +552,13 @@ func (h *Handlers) ComposeImage(ctx echo.Context) error {
 		return err
 	}
 
+	// check here for installer types
+	if composeRequest.Customizations != nil {
+		if composeRequest.Customizations.Users != nil && !strings.Contains(string(imageType), "installer") {
```
ondrejbudai, line 557: I've just realized that this can be just one `if` statement with 3 conditions joined by `&&`.

internal/v1/server.go
```diff
@@ -758,6 +765,19 @@ func buildCustomizations(cust *Customizations) *composer.Customizations {
 		res.Filesystem = &fsc
 	}
 
+	if cust.Users != nil {
+		users := make([]composer.User, len(*cust.Users))
```
ondrejbudai, line 769: I don't think this is correct, it will create a slice with the same length as `cust.Users` and then append to it. It means that the slice will have empty members at the beginning. See https://go.dev/tour/moretypes/13

It would be also great if we can cover this conversion process with a test, it would have actually uncovered this issue I think.

ondrejbudai (changes requested):
Looking good but I found one small issue in the conversion to the composer API.

internal/v1/server_test.go
```diff
@@ -911,101 +954,197 @@ func TestComposeCustomizations(t *testing.T) {
...
-					UploadRequest: UploadRequest{
-						Type:    UploadTypes_aws_s3,
-						Options: AWSS3UploadRequestOptions{},
-					},
+					Repositories:  nil,
+					UploadOptions: &composer.UploadOptions{},
 				},
 			},
 		},
 	}
```
Gundersanne, line 1131: Can we add a case for user customization? And verify that IB adds the wheel group?

internal/v1/server_test.go
```diff
@@ -911,101 +954,197 @@ func TestComposeCustomizations(t *testing.T) {
...
 		require.Equal(t, http.StatusCreated, response.StatusCode)
 
 		var result ComposeResponse
 		err := json.Unmarshal([]byte(body), &result)
 		require.NoError(t, err)
 		require.Equal(t, "fe93fb55-ae04-4e21-a8b4-25ba95c3fa64", result.Id)
+
+		fmt.Printf("ACTUAL COMPOSE REQUEST:\n")
+		fmt.Print(&composerRequest.ImageRequest.UploadOptions)
+		fmt.Printf("\n\n")
```
Gundersanne, line 1145: Leftover debug?

internal/v1/server_test.go
Gundersanne, line 1131: Can we add a case for the user customization, and verify that IB adds the wheel group?

internal/v1/api.yaml
```diff
@@ -574,6 +574,23 @@ components:
...
+            "list of users that a customer can add, also specifying their respective groups and SSH keys"
+    User:
+      type: object
+      required:
+        - name
+      properties:
+        name:
+          type: string
+          example: "user1"
+        SSHKey:
```
Gundersanne, line 591: Is there a usecase to create a user without an ssh key? I notice composer hasn't marked it as required either but wonder if there's any point to it.

Gundersanne (changes requested):
Thank you!

A test case for user customizations and verify that the group gets added by image builder is useful i think.

internal/v1/api.yaml
teg: I think in IB we should require the SSH key (maybe composer too, but at least in IB it feels definitely right to be quite opinionated in what people can do).

lavocatt:
Can you add some context in the pull request description please ? I would like to know what this is about to review the PR.

ondrejbudai (changes requested):
Changes look fine but the commits are slightly messy. I think you should move all the API-related changes from the 2nd commit to the 1st one, and make the 2nd commit only adding the test. Also, the second commit needs a better commit message.

If I could be super-picky, I would rather see the second commit being authored by you (`git commit --amend --reset-author`). I can be mentioned as a co-author if you want to give me credit. :)

ondrejbudai (approved):
Splendid!

PR #391, opened 2022-07-12
Add rhel-8 and rhel-9 alias

Gundersanne:
Composer requires that users specify the specific RHEL minor                                                                                                   
version. Image builder exposes this too. This can be misleading as older                                                                                       
minor versions aren't actually supported officially (they're not EUS                                                                                           
releases).                                                                                                                                                     
                                                                                                                                                               
Long-term image builder should only expose major versions or specific                                                                                          
EUS releases. This would follow how RHEL is actually supported and                                                                                             
released.


chloenayon (approved):
looks good to me!

PR #393, opened 2022-07-20
Refactor the distribution package

ondrejbudai:
When I was doing some work in the `distribution` package, I realized that it would benefit from a small clean-up and I decided to do it.

A big advantage of the code after the clean-up is that the distribution (and package lists) are read only once when `image-builder` is started. Otherwise, it should be just easier to read (I hope).

There is one breaking change, `RepositoriesForArch` returned `users organization is not entitled for distribution` before. Now, the generic `Distribution not available` is returned instead. I don't see much of a difference but happy to bring this behaviour back if needed.


internal/v1/server_test.go
```diff
@@ -96,8 +97,13 @@ func startServerWithCustomDB(t *testing.T, url string, dbase db.DB, distsDir str
 	quotaFile, err := initQuotaFile(t)
 	require.NoError(t, err)
 
+	adr, err := distribution.LoadDistroRegistry(distsDir)
+	if err != nil {
+		panic(err)
+	}
```
Gundersanne, line 103: require.NoError

Gundersanne (changes requested):
Nice! one nitpick

internal/v1/server_test.go
ondrejbudai: Fixed, good catch, thanks! :)

Gundersanne (approved):
Thanks!

PR #398, opened 2022-08-02
treat composer service code 10 as http.status 400

### LINKED ISSUE #385: IMAGE-BUILDER-COMPOSER-10 errors should be treated as 400
The error passed from composer should be parsed, and passed along as a 400 error. As this error most likely means user input error ("Error resolving OSTree repo").

miyunari:
closes #385

here is my first draft. Would be nice to get already some feedback on it :)

### DIFF
--- a/internal/v1/handler.go
+++ b/internal/v1/handler.go
@@ -486,6 +486,14 @@ func (h *Handlers) ComposeImage(ctx echo.Context) error {
 			logrus.Errorf("Unable to parse composer's compose response: %v", err)
 		} else {
 			_ = httpError.SetInternal(fmt.Errorf("%s", body))
+			var serviceStat composer.Error
+			if err := json.Unmarshal(body, &serviceStat); err != nil {
+				return httpError
+			}
+			if serviceStat.Id == "10" {
+				httpError.Message = "Error resolving OSTree repo"
+				httpError.Code = http.StatusBadRequest
+			}
 		}
 		return httpError
 	}
--- a/internal/v1/server_test.go
+++ b/internal/v1/server_test.go
@@ -728,6 +728,58 @@ func TestComposeImageErrorsWhenStatusCodeIsNotStatusCreated(t *testing.T) {
 	require.Contains(t, body, "Failed posting compose request to osbuild-composer")
 }
 
+func TestComposeImageErrorResolvingOSTree(t *testing.T) {
+	// simulate osbuild-composer API
+	apiSrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if "Bearer" == r.Header.Get("Authorization") {
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+		require.Equal(t, "Bearer accesstoken", r.Header.Get("Authorization"))
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		serviceStat := &composer.Error{
+			Id:     "10",
+			Reason: "not ok",
+		}
+		err := json.NewEncoder(w).Encode(serviceStat)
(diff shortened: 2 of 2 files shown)


internal/v1/server.go
```diff
@@ -576,6 +576,14 @@ func (h *Handlers) ComposeImage(ctx echo.Context) error {
 			logrus.Errorf("Unable to parse composer's compose response: %v", err)
 		} else {
 			_ = httpError.SetInternal(fmt.Errorf("%s", body))
+			var serviceStat composer.ComposeStatus
```
Gundersanne, line 579: The error returned here would be of `composer.Error` type.

internal/v1/server_test.go
```diff
@@ -677,6 +677,57 @@ func TestComposeImageErrorsWhenStatusCodeIsNotStatusCreated(t *testing.T) {
...
+		require.Equal(t, "Bearer accesstoken", r.Header.Get("Authorization"))
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		serviceStat := composer.ComposeStatus{
+			ImageStatus: composer.ImageStatus{
+				Error: &composer.ComposeStatusError{
+					Id:     10,
+					Reason: "not ok",
+				},
+			},
```
Gundersanne, line 696: Similar here, composer returns composer.Error, not ComposeStatus.

Gundersanne (changes requested):
Sorry for the late review, the idea is correct but mostly using the wrong types here.

Composer returns an error object in case http calls fail, composer.ComposeStatus is an object that is returned when querying the status of a compose. See https://api.openshift.com/?urls.primaryName=Image%20Builder%20service.

miyunari:
hi @Gundersanne, thanks for your feedback. Could you please have a look again? :)

internal/v1/server_test.go
```diff
@@ -677,6 +677,55 @@ func TestComposeImageErrorsWhenStatusCodeIsNotStatusCreated(t *testing.T) {
...
+				ImageType:    ImageTypes_aws,
+				UploadRequest: UploadRequest{
+					Type: UploadTypes_aws,
+					Options: AWSUploadRequestOptions{
+						ShareWithAccounts: []string{"test-account"},
+					},
+				},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
miyunari: Thanks, I added the missing parts :)

internal/v1/handler.go
```diff
@@ -486,6 +487,20 @@ func (h *Handlers) ComposeImage(ctx echo.Context) error {
...
+
+			id, err := strconv.Atoi(serviceStat.Id)
+			if err != nil {
+				return httpError
+			}
+
+			if id == 10 {
+				httpError.Message = "Error resolving OSTree repo"
+				httpError.Code = http.StatusBadRequest
+			}
```
Gundersanne, line 503: I don't think it's necessary to convert the id to an integer first. Just check if serviceStat.Id == "10"

Gundersanne (commented):
One final comment, otherwise lgtm

internal/v1/handler.go
miyunari: ok, done :)

Gundersanne (dismissed):
Thank you!

miyunari:
Hey @Gundersanne, is there anything else I need to do? :)

miyunari:
Just called `./tools/prepare-source.sh` again, to fix the broken test.

cc: @achilleas-k @Gundersanne

achilleas-k (approved):

croissanne:
Egh, seems like a variable got marked as unrequired. Will send a separate PR.

croissanne:
Also opened https://github.com/osbuild/image-builder/issues/467

croissanne (approved):

PR #402, opened 2022-09-02
.github: Fix dependabot schedule and bump open PR limit

Gundersanne:
GH threw the following error:
```
The property '#/updates/0/schedule/time' value "4:00" did not match the regex '^\d\d:\d\d$'
The property '#/updates/1/schedule/time' value "4:00" did not match the regex '^\d\d:\d\d$'
```


jrusz (commented):
Seems like this needs `tools/prepare-source.sh` ?

Gundersanne:
> Seems like this needs `tools/prepare-source.sh` ?

Indeed, done!

teg (approved):

PR #403, opened 2022-09-02
Clone aws images across regions

Gundersanne:
Adds 3 new endpoints. To clone images across regions:

1. post to /composes/{id}/clone
2. poll status at /clones/{id} until complete
3. list clones of a specific compose at /composes/{id}/clones

---

TODO 
- add  clone to aws integration test?
- what if it's a pure sharing thing (so without a new target region?); do we still treat it as any other clone and deduplicate the actual AMIs in the FE?


internal/db/db.go
```diff
@@ -104,6 +109,29 @@ func (db *dB) GetCompose(jobId string, orgId string) (*ComposeEntry, error) {
...
+		return "", err
+	}
+	defer conn.Release()
+
+	result := conn.QueryRow(ctx, sqlGetComposeImageType, orgId, jobId)
+
+	var imageType string
+	err = result.Scan(&imageType)
+	if err != nil {
+		if errors.As(err, &pgx.ErrNoRows) {
```
kingsleyzissou, line 166: Should this be `errors.Is` like it is here:
https://github.com/osbuild/osbuild-composer/commit/9def5455708f7e247fe4103d47e00a7beb52169b

internal/db/db.go
```diff
@@ -193,3 +234,79 @@ func (db *dB) CountComposesSince(orgId string, duration time.Duration) (int, err
...
+	conn, err := db.Pool.Acquire(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Release()
+
+	var clone CloneEntry
+	err = conn.QueryRow(ctx, sqlGetClone, id, orgId).Scan(&clone.Id, &clone.Request, &clone.CreatedAt)
+	if err != nil {
+		if errors.As(err, &pgx.ErrNoRows) {
```
kingsleyzissou, line 304: Same here

internal/v1/handler.go
```diff
@@ -711,3 +711,195 @@ func buildCustomizations(cust *Customizations) *composer.Customizations {
...
+	if err != nil {
+		return err
+	}
+
+	idHeader, err := getIdentityHeader(ctx)
+	if err != nil {
+		return err
+	}
+
+	imageType, err := h.server.db.GetComposeImageType(composeId, idHeader.Identity.OrgID)
```
ondrejbudai, line 843: I wonder if this check is even needed. Querying a compose other than an aws one does not harm. :shrug:

internal/v1/handler.go
```diff
@@ -711,3 +711,195 @@ func buildCustomizations(cust *Customizations) *composer.Customizations {
...
+	}
+
+	offset := 0
+	if params.Offset != nil {
+		offset = *params.Offset
+	}
+
+	cloneEntries, count, err := h.server.db.GetClonesForCompose(composeId, idHeader.Identity.OrgID, limit, offset)
+	if err != nil {
+		return err
```
ondrejbudai, line 866: I wonder whether it's valuable to translate this error. We do it for other DB queries but I'm not sure what's your rule of thumb.

Gundersanne: ib is pretty bad at this yea, makes sense to always translate.

internal/v1/handler.go
Gundersanne: true

internal/db/migrations/007_alter_table_composes_alter_colum_request_json_to_jsonb.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE composes ALTER COLUMN request SET DATA TYPE jsonb;
```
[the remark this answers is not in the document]
Gundersanne: I added it to a different commit instead. The one that actually this column as jsonb.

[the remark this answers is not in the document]
ondrejbudai: Makes sense, thanks! :)

internal/db/db.go
ondrejbudai: I think this is still unresolved.

ondrejbudai (changes requested):
One small mistake, otherwise, this is awesome! Thanks for fixing the commits and their messages. :)

internal/db/db.go
kingsleyzissou: Whoops I also just noticed that now too

Gundersanne: Woops :facepalm:

kingsleyzissou (approved):
lgtm

ondrejbudai (approved):
:heart:

PR #414, opened 2022-09-04
Update to Go 1.17 and oapi-codegen to 1.11

ondrejbudai:
See individual commits, I don't think I have much to add.


Gundersanne (approved):
lgtm, ty!

deepmap also has `--old-style-config` or something, but might as well.

ondrejbudai:
I know but I think it's worth switching long-term. :)

PR #420, opened 2022-09-07
v1: remove upload request

### LINKED ISSUE #368: API v1 can be simplified for s3 images
Currently, if I want to create a guest-image, I must specify also the upload request:

```
{
  "distribution": "rhel-90",
  "image_requests": [
    {
      "architecture": "x86_64",
      "image_type": "guest-image",
      "upload_request": {
        "type": "aws.s3",
        "options": {}
      }
    }
  ]
}
```

The upload request is redundant though, the upload request type can be inferred from the image type (we always upload guest images into s3) and the options are always... empty.

Ideally, I would like to request the image with only:

```
{
  "distribution": "rhel-90",
  "image_requests": [
    {
      "architecture": "x86_64",
      "image_type": "guest-image"
    }
  ]
}
```

Much less typing, same result.

chloenayon:
Fixes #368


internal/v1/api.yaml
```diff
@@ -395,7 +395,6 @@ components:
       required:
         - architecture
         - image_type
-        - upload_request
```
Gundersanne, line 398: Upload request is required, no?

internal/v1/server.go
```diff
@@ -507,7 +507,7 @@ func (h *Handlers) ComposeImage(ctx echo.Context) error {
 		return err
 	}
 
-	if (composeRequest.ImageRequests[0].UploadRequest == UploadRequest{}) {
+	if (composeRequest.ImageRequests[0].UploadRequest == &UploadRequest{}) {
```
Gundersanne, line 510: This check will fail if there's an empty upload request object, but not on nil?

Gundersanne (commented):
Why mark upload_request as not required? I think this PR is missing some context?

internal/v1/api.yaml
chloenayon: I should've added it to the description, this is intended to address #368 , there are cases in which upload request is not necessary and can be inferred from the image type

ondrejbudai:
Sorry for ignoring this PR for so long, I wasn't sure what to do about it until today.

While I'm very much for any simplification, this has the disadvantage that it's no longer obvious from the spec that `upload_options` are still required for EC2, GCP and Azure. I think that I prefer the spec being obvious rather than the API calls being simple.

I actually proposed a change to the cloud API in composer today that makes `upload_options` required for just some image types - https://github.com/osbuild/osbuild-composer/pull/3018. If the reviewers like this change, I suppose we can make the same changes here and make all upload options required except for the S3 one, which indeed isn't needed.

PR #421, opened 2022-09-14
v1: add support for an image name when uploading to azure

ondrejbudai:
I think that custom names of the uploaded artifacts are definitely the way to go. For now, I care only about the backend, since this is what my story needs to use. Therefore, I figured out that mimicking the composer's API is good enough for now.

If we ever wanted to support this in the frontend, I'm not sure about the form. The issue is that we already have a "global name"  for the compose exposed in the frontend, but this might not be usable for all clouds since all of them have some requirements for its format.

### The actual commit message:

Composer already supports that, so let's just add the name into our IB's API.

The validation pattern is taken from Azure's GUI, it shows

> The name must begin with a letter or number, end with a letter, number or
> underscore, and may contain only letters, numbers, underscores, periods, or
> hyphens.

when a bad name is inserted. The max length of 60 was found experimentally.


Gundersanne (approved):

ondrejbudai:
No worries, it took just a minute. ;)

Gundersanne (approved):
thank you!

PR #424, opened 2022-09-14
templates: Grab access tokens from RH SSO

Gundersanne:
Newer service accounts only work with RH SSO. Older accounts work with both Openshift and RH SSO. Let's just switch to using RH SSO in case the service accounts ever need to be reprovisioned.


ondrejbudai (approved):
Thanks!

PR #428, opened 2022-09-19
v1: Spec fixes and validation in github workflows

Gundersanne:

PR #429, opened 2022-09-19
tools/prepare-source.sh: Run go mod generate before fmt and mod

Gundersanne:
`go generate` somehow touches go.sum, which causes issues for dependabot PRs (osbuild/image-builder#418 for example). Run mod tidy command last.


ondrejbudai (approved):
Also makes sense to run `generate` before `fmt` I think. :)

Also: my god :/

Gundersanne:
I force pushed the #428 emoji fix here :facepalm:

ondrejbudai (approved):
Hehe, no worries. :)

PR #430, opened 2022-09-19
distributions: add Fedora 35-38

ondrejbudai:
This PR brings these new features:

- a distribution can opt out from having a list of packages. This is useful for distros not exposed in the frontend and that has a huge number of packages (Fedora)
- `restrictedAccess` is now `restricted_access`
- `metalink` is now supported
- Fedora 35-38 is supported for allow listed orgs (will be done in app-intf)

There's one thing that I'm not sure about: The `GET /distributions` route returns Fedora also for not allow-listed accounts. This feels wrong, but on the other hand, if people find Fedora in the list but cannot build it, they might reach out to us, which would give us the idea if there are some interested customers. Not sure if that's a legitimate use-case though...


Gundersanne (approved):
lgtm!

PR #433, opened 2022-09-20
Add support for building Fedora 37 images

achilleas-k:
Enumerate all supported F37 image types in the distribution configuration, but only allow iot-raw-image in the API for now.

@Gundersanne can you have a look at the Fedora repository URLs?  I used metalinks which I don't see being used for any other distro.


ondrejbudai:
I think this is duplicating https://github.com/osbuild/image-builder/pull/430

achilleas-k:
> I think this is duplicating #430

It is.  My bad :)  Closing

PR #435, opened 2022-09-21
templates: Add clowder template

jrusz:
This is an initial template to enable deployment to ephemeral environments. It contains the minimum necessary to deploy a working image-builder backend instance.

Connection to composer instance is done via a separate service account and the credentials are stored as secrets in the ephemeral-base environment. The secrets are not deployed automatically and have to be setup manually in the environment before deploying image-builder.

There are also some minor code changes to get database connection info from the ephemeral environment.

There will be a followup which will also enable running IQE test suite as part of pr_check.sh.


templates/image-builder-clowder.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: image-builder-8080
+        protocol: TCP
+        port: 8080
+        targetPort: 8000
+    selector:
+      pod: image-builder-service
+
+parameters:
+  - name: IMAGE
+    value: quay.io/jrusz/image-builder
```
Gundersanne, line 92: quay.io/cloudservices/image-builder

Gundersanne (commented):
had preliminary look, looks good! some comments, i'll try it out as well

templates/image-builder-clowder.yml
jrusz: Thanks!

jrusz:
I think the parameters will need some update later, but since this is a separate template we can iterate the changes without worrying that we'll break something in stage or prod. I'm also creating a separate saas deploy file for this.

jrusz:
I'm trying to trigger https://ci.ext.devshift.net/job/osbuild-image-builder-gh-pr-check/ but idk why it's not triggering on this PR :/

teg:
For a follow up, probably:
Would be nice to add instructions for using clowder to the README.
Should we use clowder unconditionally, also in stage/production?

Gundersanne:
> For a follow up, probably: Would be nice to add instructions for using clowder to the README. Should we use clowder unconditionally, also in stage/production?

Yea, if this works fine in ephemeral we should replace the other deployments with it.

jrusz:
> For a follow up, probably: Would be nice to add instructions for using clowder to the README. Should we use clowder unconditionally, also in stage/production?

Yeah we should be able to replace it once we're confident enough that it works as expected. 

Also I am going to create a Jenkins job which will provision Image-builder in ephemeral environment so I'll put a link to that in README once it's ready and will probably also create a docs page in our internal docs.

Gundersanne (approved):
thank you!

PR #437, opened 2022-09-21
templates: Clean up the configuration of IB and add an allow list

ondrejbudai:
See individual commits.


Gundersanne (approved):
lgtm ty, feel free to mark as ready/merge as soon as the needed config map is deployed

PR #441, opened 2022-09-22
 templates/image-builder.yml: update splunk port for splunk cloud

diaasami:
and use an openshift template variable for flexibility while deploying


Gundersanne (approved):
lgtm, ty!

ondrejbudai (approved):

PR #444, opened 2022-09-22
DEBUGGING - DO NOT MERGE templates: Add clowder template

jrusz:
This is an initial template to enable deployment to ephemeral environments. It contains the minimum necessary to deploy a working image-builder backend instance.

Connection to composer instance is done via a separate service account and the credentials are stored as secrets in the ephemeral-base environment. The secrets are not deployed automatically and have to be setup manually in the environment before deploying image-builder.

There are also some minor code changes to get database connection info from the ephemeral environment.

There will be a followup which will also enable running IQE test suite as part of pr_check.sh.


croissanne:
@jrusz This can be closed now, right?

PR #451, opened 2022-09-28
Clowder initcontainer

Gundersanne:


jrusz (changes requested):
Looks good, thanks! I had something similar locally, but didn't know about https://github.com/osbuild/image-builder/commit/e21832ae1fee9cd3ed9f207a094344617f3b28d1 I just put it in the migration `main.go` file similar to how it was in the image-builder `main.go` file  :sweat_smile:  

I've tried deploying it and running the IQE test suite against this and the `test_endpoints_acessible` is reporting PASS, so I'd say it's good. I just had to change the hostname and port. I'm not sure about how will this work with the frontend though, but we'll come to that later.

I have two questions. 

1) Is the  `COMPOSER_TOKEN_URL` correct actually? You told me it should be `https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token` but tbh I don't know how to tell, what's the difference...
1) Are the resource limits actually used? We don't specify them as parameters here...

jrusz:
Nice, so the trigger is fixed.

Gundersanne:
> Looks good, thanks! I had something similar locally, but didn't know about [e21832a](https://github.com/osbuild/image-builder/commit/e21832ae1fee9cd3ed9f207a094344617f3b28d1) I just put it in the migration `main.go` file similar to how it was in the image-builder `main.go` file sweat_smile

Yea that would work too; but let's just put the entire thing in the config package no? Might as well since both entrypoints need to check if clowder is enabled.
 
> I've tried deploying it and running the IQE test suite against this and the `test_endpoints_acessible` is reporting PASS, so I'd say it's good. I just had to change the hostname and port. I'm not sure about how will this work with the frontend though, but we'll come to that later.

Awesome, thank you :)

> 
> I have two questions.
> 
>     1. Is the  `COMPOSER_TOKEN_URL` correct actually? You told me it should be `https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token` but tbh I don't know how to tell, what's the difference...

Either will work for the current deployments; but you're right, let's switch to rh sso by default as that's where it's heading in future.

> 
>     2. Are the resource limits actually used? We don't specify them as parameters here...

Good point! We should though, I'll add them. Resource limit + pod affinity will need to be set in order to pass kubelinter. On that note, I'll add this to our own kubelinter check :D

jrusz (commented):
> > Looks good, thanks! I had something similar locally, but didn't know about [e21832a](https://github.com/osbuild/image-builder/commit/e21832ae1fee9cd3ed9f207a094344617f3b28d1) I just put it in the migration main.go file similar to how it was in the image-builder main.go file sweat_smile

> Yea that would work too; but let's just put the entire thing in the config package no? Might as well since both entrypoints need to check if clowder is enabled.

Yes that's great, I didn't think of this option. 

I actually have two more questions

templates/image-builder-clowder.yml
```diff
@@ -55,38 +58,28 @@ objects:
...
+              cpu: "${CPU_LIMIT}"
+              memory: "${MEMORY_LIMIT}"
+          env:
+          - name: MIGRATIONS_DIR
+            value: "${MIGRATIONS_DIR}"
       webServices:
-        public: 
+        public:
           enabled: true
-          apiPath: image-builder/v1
```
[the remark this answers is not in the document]
Gundersanne: Not required. Anything that comes in on the service this provides will be routed to image-builder. Happy to re-add it but without the `/v1`, so just `image-builder`. Either way it doesn't matter hugely.

templates/image-builder-clowder.yml
```diff
@@ -34,11 +34,14 @@ objects:
           periodSeconds: 30
           successThreshold: 1
           timeoutSeconds: 10
+        ports:
+        - containerPort: 8000
+          protocol: TCP
```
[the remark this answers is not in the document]
jrusz: Huh, I see... Well this is a bit confusing to me now cause by default services in ephemeral environment are epxected to listen on port 8000 if possible which is what the public webservices expose actually... So I don't understand what's the expose in Dockerfile actually doing :/

templates/image-builder-clowder.yml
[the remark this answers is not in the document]
jrusz: Yeah ok, I was mostly curious, the IQE test suite seems to be fine without this, it defines the path by itself, so if we don't need it we can remove it.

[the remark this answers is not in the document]
Gundersanne: scratch that, it's actually required :facepalm:  but i'll go with "image-builder"

templates/image-builder-clowder.yml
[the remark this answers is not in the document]
Gundersanne: In this case, nothing. In the old case, it's actually the default port we listen on :) Once we've migrated to clowder we can drop the EXPOSE from the dockerfile.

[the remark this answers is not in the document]
Gundersanne: Actually, you're right again. It automatically  exposes port 8000, so this isn't required :facepalm:

templates/image-builder-clowder.yml
```diff
@@ -38,13 +46,15 @@ objects:
...
-            value: "${COMPOSER_TOKEN_URL}" 
+            value: "${COMPOSER_TOKEN_URL}"
           - name: DISTRIBUTIONS_DIR
             value: '/app/distributions'
           - name: CLOWDER_ENABLED
             value: ${CLOWDER_ENABLED}
           - name: COMPOSER_URL
             value: "${COMPOSER_URL}"
+          - name: OSBUILD_AWS_REGION
+            valud: "${OSBUILD_AWS_REGION}"
```
jrusz, line 57: Typo :)
```suggestion
            value: "${OSBUILD_AWS_REGION}"
```

Gundersanne: :facepalm:

templates/image-builder-clowder.yml
```diff
@@ -14,8 +14,16 @@ objects:
       iqePlugin: image-builder
     deployments:
     - name: service
+      minReplicas: ${{MIN_REPLICAS}}
```
jrusz, line 17: This doesn't seem to do anything, it does work if I change it to `replicas` though, can you double check?

templates/image-builder-clowder.yml
```diff
@@ -14,8 +14,16 @@ objects:
...
       podSpec:
         image: ${IMAGE}:${IMAGE_TAG}
+        resources:
+        requests:
+          cpu: "${CPU_REQUEST}"
+          memory: "${MEMORY_REQUEST}"
+        limits:
+          cpu: "${CPU_LIMIT}"
+          memory: "${MEMORY_LIMIT}"
```
jrusz, line 26: Shouldn't this be indented?

jrusz: Also, how is this supposed to work? Even when I indent it, looking at the yaml of the pod in OC I see:
```
      containers:
        - resources:
            limits:
              cpu: 300m
              memory: 256Mi
            requests:
              cpu: 30m
              memory: 128Mi
```

jrusz (changes requested):
Got some comments inline, but otherwise looks good :)

templates/image-builder-clowder.yml
Gundersanne: woops, yea this should be indented. resources should be set on podspec? so unsure why it doesn't get the right values.

templates/image-builder-clowder.yml
Gundersanne: According to the api spec https://consoledot.pages.redhat.com/clowder/dev/api_reference.html#k8s-api-github-com-redhatinsights-clowder-apis-cloud-redhat-com-v1alpha1-deployment, minReplicas is indeed deprecated.

templates/image-builder-clowder.yml
Gundersanne: this is weird, it's clearly in the spec; but it seems like it's not being passed on at all

PR #452, opened 2022-09-28
middleware: return 400 on route error

### LINKED ISSUE #426: Return 400 if route note found or bad method
https://github.com/osbuild/image-builder/blob/7bf6468c6e1f2c4518283ba0712a34d00eb012c7/internal/v1/middleware.go#L10-L31

`FindRoute` returns ErrPathNotFound or ErrMethodNotAllowed, both should result in a 400. Currently ib returns a 500.

chloenayon:
Previously returned a 500, change to return a 400
when FindRoute fails.
Fixes #426

### DIFF
--- a/internal/v1/middleware.go
+++ b/internal/v1/middleware.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 
 	"github.com/getkin/kin-openapi/openapi3filter"
+	"github.com/getkin/kin-openapi/routers"
 	"github.com/labstack/echo/v4"
 )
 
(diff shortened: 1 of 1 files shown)


Gundersanne (commented):
Rereading the issue, I think 404 and 405 are much more appropriate. A 404 on ErrPathNotFound and 405 on ErrMethodNotAllowed. Ideally something obvious in the error as well, unsure what the error returned by findroute looks like.

croissanne (approved):
Perfect, thank you!

PR #459, opened 2022-10-06
db: Switch to jackc/tern for migrations

### LINKED ISSUE #197: Use jackc/tern for database migration
Currently we depend on `lib/pq` for migrations and `jackc` for the actual db operations. It makes sense to move the migrations over to `jackc/tern`. Furthermore `lib/pq` doesn't support sslmode `prefer`, which should be the sensible default.

croissanne:
Tern uses transactions for each migration, so a failed migration can just be resolved by pushing a new version. Migrations with `golang-migrate` might result in a broken database which needs to be manually resolved.

This doesn't remove golang-migrate yet, as it runs the db-tests against a database migrated with golang-migrate and tern, and against a database migrated with just tern.

Fixes #197

### DIFF
--- a/templates/image-builder.yml
+++ b/templates/image-builder.yml
@@ -214,7 +214,7 @@ objects:
         initContainers:
         - name: image-builder-migrate
           image: "${IMAGE_NAME}:${IMAGE_TAG}"
-          command: [ "/app/image-builder-migrate-db" ]
+          command: [ "/app/image-builder-migrate-db-tern" ]
           resources:
             requests:
               cpu: "${CPU_REQUEST}"
@@ -223,8 +223,6 @@ objects:
               cpu: "${CPU_LIMIT}"
               memory: "${MEMORY_LIMIT}"
           env:
-          - name: MIGRATIONS_DIR
-            value: "${MIGRATIONS_DIR}"
           # Credentials/configuration for AWS RDS.
           - name: PGHOST
             valueFrom:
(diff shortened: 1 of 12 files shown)


templates/image-builder.yml
```diff
@@ -214,7 +214,7 @@ objects:
         initContainers:
         - name: image-builder-migrate
           image: "${IMAGE_NAME}:${IMAGE_TAG}"
-          command: [ "/app/image-builder-migrate-db" ]
+          command: [ "/opt/migrate/tern", "migrate", "-m", "${MIGRATIONS_DIR}" ]
```
ondrejbudai, line 217: Didn't you delete `MIGRATIONS_DIR` later in this file?

ondrejbudai (commented):
Generally looks good, I have just one concern in the template.

templates/image-builder.yml
croissanne: It's still a template variable; but idk if it makes much sense to have it as a variable in the first place tbh, since we always copy the migrations to /app/migrations in the container.

croissanne: I just hardcoded the path

ondrejbudai (dismissed):

croissanne:
egh we can use the tern cli with a database supplied through clowder :facepalm:

ondrejbudai (approved):
Thanks!

PR #468, opened 2022-10-13
internal/composer: spec update

croissanne:
The GCP bucket is no longer a required variable.


lucasgarfield:
Is this due to an upstream change from Google or...?

croissanne:
> Is this due to an upstream change from Google or...?

OH no, upstream spec change of composer's cloudapi; but i want to just start doing this in scheduled PRs (see #469)

PR #472, opened 2022-10-14
templates: add FloorPlan

achilleas-k:
FloorPlan for data extraction based on sample [1] and Compliance's FloorPlan [2].

The query already runs weekly and is defined in app-interface.  With this FloorPlan, we will configure the Floorist operator to store the data in an S3 bucket in the Parquet format.

[1] https://github.com/RedHatInsights/compliance-backend/blob/822e2d9823d88a2ac01b2184c22e3a77827d411c/clowdapp.yaml#L272-L298
[2] https://github.com/RedHatInsights/compliance-backend/blob/822e2d9823d88a2ac01b2184c22e3a77827d411c/clowdapp.yaml#L272-L298

### DIFF
--- a/templates/image-builder.yml
+++ b/templates/image-builder.yml
@@ -322,6 +322,30 @@ objects:
         hec_token "#{ENV['SPLUNK_HEC_TOKEN']}"
       </match>
 
+- apiVersion: metrics.console.redhat.com/v1alpha1
+  kind: FloorPlan
+  metadata:
+    name: imagebuilder
+  spec:
+    database:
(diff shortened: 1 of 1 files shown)


achilleas-k:
Converting to draft because we need to set the secrets and variables before merging.

croissanne:
> Converting to draft because we need to set the secrets and variables before merging.

Can you add those variables as template variables; so we can set them per namespace in app-intf?

croissanne (approved):
:rocket:

PR #485, opened 2022-10-25
.github/codeql: add codeql static analysis

croissanne:
LGTM.com is being deprecated (see
https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/). The codeql analysis engine is available on GH however.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,29 @@
+name: "CodeQL"
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


jrusz (dismissed):
Checked it according to https://github.com/github/codeql-action#usage and all seems good, it just says that the permissions for `actions` and `contents` are not necessary for public repos, but can't hurt to have it I guess, so LGTM

jrusz (approved):

PR #488, opened 2022-10-27
internal/v1: correctly cap offset in GetPackages

croissanne:
internal/v1: correctly deal with offsets in GetPackages
    
If the offset was 0, which is the default, and if no packages were
found, it panicked due to an out of bounds index.
    
The reasoning by subtracting one was that even if you chose too large of
an offset, it would still give you back something. Simply cap the offset
in case it's larger than the amount of packages. The data returned will
be empty but that's ok.
    
Also cap the last offset at 0 if no results were found.

### DIFF
--- a/internal/v1/handler.go
+++ b/internal/v1/handler.go
@@ -113,8 +113,8 @@ func (h *Handlers) GetPackages(ctx echo.Context, params GetPackagesParams) error
 
 	offset := 0
 	if params.Offset != nil {
-		if *params.Offset >= len(packages) {
-			offset = len(packages) - 1
+		if *params.Offset > len(packages) {
+			offset = len(packages)
 		} else if *params.Offset > 0 {
 			offset = *params.Offset
 		}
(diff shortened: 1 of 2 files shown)


kingsleyzissou (approved):
lgtm

jkozol (approved):
Looks good. Just to clarify, our desired behavior is to have an offset of 0 if we have either 0 or 1 packages?

croissanne:
> Looks good. Just to clarify, our desired behavior is to have an offset of 0 if we have either 0 or 1 packages?

first is supposed to point at the first package in the entire collection; last is supposed to point at the last package in the whole collection. In case there's 0 packages found it's fine if both are 0 I think. If there's 1 found both should be 0. And then more it's 0 and len(x)-1.

Come to think of it, offset in first should probably always be 0. repushing that.

jkozol (approved):

kingsleyzissou (approved):

Issue #457, opened 2022-10-05
ISO-8601 compliant datetime strings


lucasgarfield:
https://en.wikipedia.org/wiki/ISO_8601

Currently the API returns a human readable string for the time a compose was created, something like `"2022-10-01 08:31:13.23555 +0000 UTC"`.

Is there any issue with returning an ISO-8601 compliant string instead? The example given would in that case be: `"2022-10-01T08:31:13.23555+00:00"`. This is slightly less human readable, but would be much easier to parse if working with the API programmatically.

For instance, I am currently implementing a feature in the front end that requires me to find the most recently created image. Javascript's Date() object is able to parse ISO-8601 compliant datetime strings and I would be able to simply write code like this:
```
const date1 = new Date('2022-10-01T08:31:13.23555+00:00"');
const date2 = new Date('2022-10-01T08:32:13.23555+00:00"');
if (date1 > date2) {
  ...
}
```

(Note that Javascript's Date() object is not able to parse what the API currently returns.)

PR #493, opened 2022-10-28
internal/v1: ISO8601 datetime

kingsleyzissou:
Return an ISO8601 compliant datetime to facilitate easier parsing and comparison in the frontend.

Addresses #457

### DIFF
--- a/internal/v1/handler.go
+++ b/internal/v1/handler.go
@@ -363,7 +363,7 @@ func (h *Handlers) GetComposes(ctx echo.Context, params GetComposesParams) error
 	var data []ComposesResponseItem
 	for _, c := range composes {
 		data = append(data, ComposesResponseItem{
-			CreatedAt: c.CreatedAt.String(),
+			CreatedAt: c.CreatedAt.Format(time.RFC3339),
 			Id:        c.Id.String(),
 			ImageName: c.ImageName,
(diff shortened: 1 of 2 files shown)


kingsleyzissou:
I used the `time` package's `RFC339` constant. Fore reference, [here](https://ijmacd.github.io/rfc3339-iso8601/) are the differences with ISO 8601.

The result is the following format:
"2022-10-28T13:37:38Z"

But this can be parsed and compared with the JavaScript `Date` object

image

ondrejbudai (approved):
Thanks!

ondrejbudai:
@croissanne This breaks backward compatibility a bit. On the other hand, we've never specified the format of the string before, so this is not that bad, actually. WDYT?

croissanne:
> @croissanne This breaks backward compatibility a bit. On the other hand, we've never specified the format of the string before, so this is not that bad, actually. WDYT?

I think that's fine yea, I don't think anyone relies on it and the frontend was adapted to deal with both formats.

lucasgarfield:
With respect to `image-builder-frontend` this will be fine, we only use only the date which we obtain by slicing the first 10 characters of the date string we receive from `image-builder`. The first 10 characters are YYYY-MM-DD before and after this PR.

PR #504, opened 2022-11-08
distributions: add rhel-87

croissanne:


ondrejbudai (approved):

ondrejbudai:
Thanks, that was quick! :)

PR #523, opened 2022-11-23
logger: logrus hook for splunk forwarding

croissanne:
To make the switch to ephemeral/Clowder easier, drop the existing sidecar and rely on the splunk forwarder hook.

---

Clowder doesn't support sidecars in the same way as openshift Deployments do. Sidecars are predefined by Clowder, and the one we'd need isn't available.


internal/logger/splunk_logger.go
```diff
@@ -0,0 +1,135 @@
...
+
+	req, err := http.NewRequest("POST", sl.url, bytes.NewReader(buf.Bytes()))
+	if err != nil {
+		return err
+	}
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("Authorization", fmt.Sprintf("Splunk %s", sl.token))
+
+	res, err := sl.client.Do(req)
+	if err != nil {
```
diaasami, line 109: I think it makes sense to retry once or twice for failed requests, may be this can be added later using retryablehttp.
It is not common but I believe it used  to happen occasionally, probably once every week or two.

diaasami (approved):

internal/logger/splunk_logger.go
croissanne: makes sense, opened https://github.com/osbuild/image-builder/issues/525 as a followup.

PR #534, opened 2022-12-05
templates/image-builder-clowder: prepare template for stage

croissanne:


croissanne:
> Well, I'm not sure what is it all that we need in stage/prod but this still works fine in epehemeral, I've tested it manually as well to be sure.
> 
> To test it in stage we'll have to change the `saas` file in `app-interface` and in case there are issues we'll revert and do a followup, right?

yup!

PR #541, opened 2022-12-12
templates: move to ClowdApp for main deployment

croissanne:
Keep the image-builder-clowder template for now as a copy, so the
    ephemeral environment doesn't break (bonfire doesn't support symlinks).


jrusz (approved):
I've tried deploying this template in ephemeral and everything seems to work fine. Ran the IQE ui test suite and it passed so it has to be working fine. Let's see if it's the same in stage :)

PR #561, opened 2022-12-21
Add aarch64

diaasami:


test/cases/api.sh
```diff
@@ -134,6 +134,17 @@ ARCH=$(uname -m)
 
 DISTRO="rhel-8"
 SSH_USER="cloud-user"
+
+if [[ "$ARCH" == "x86_64" ]]; then
+    INSTANCE_TYPE="t2.micro"
+elif [[ "$ARCH" == "aarch64" ]]; then
+    INSTANCE_TYPE="t4g.micro"
```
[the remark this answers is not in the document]
ondrejbudai: FTR, Fedora requires 2GB of memory: https://docs.fedoraproject.org/en-US/fedora/latest/release-notes/welcome/Hardware_Overview/#hardware_overview-specs

There was recently a thread on `fedora-devel` about `dnf` being memory hungry. :/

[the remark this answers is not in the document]
diaasami: In that case may be also update `t2.micro` since it has 1 GB of mem only.

schutzbot/deploy.sh
```diff
@@ -14,7 +14,7 @@ sudo dnf install -y podman \
      qemu-kvm \
      jq
 
-sudo podman pull docker://quay.io/osbuild/postgres:latest
+sudo podman pull docker://quay.io/osbuild/postgres:13-alpine
```
ondrejbudai, line 17: Just curious, what's going on here with `13-alpine` instead of `latest`?

schutzbot/deploy.sh
```diff
@@ -34,16 +34,17 @@ for RETRY in {1..10}; do
 done
 
 # Pull image-builder image
+ARCH=$(uname -m)
 QUAY_REPO_URL="quay.io/osbuild/image-builder-test"
 QUAY_REPO_TAG="${CI_PIPELINE_ID}"
-sudo podman pull --creds "${V2_QUAY_USERNAME}":"${V2_QUAY_PASSWORD}" "${QUAY_REPO_URL}":"${QUAY_REPO_TAG}"
+sudo podman pull --creds "${V2_QUAY_USERNAME}":"${V2_QUAY_PASSWORD}" "${QUAY_REPO_URL}":"$ARCH"-"${QUAY_REPO_TAG}"
```
ondrejbudai, line 40: I just wonder if we want to put the `$ARCH` as a postfix instead of a prefix. Sounds a bit more natural... What do you think?

schutzbot/deploy.sh
```diff
@@ -34,16 +34,17 @@ for RETRY in {1..10}; do
 done
 
 # Pull image-builder image
+ARCH=$(uname -m)
 QUAY_REPO_URL="quay.io/osbuild/image-builder-test"
 QUAY_REPO_TAG="${CI_PIPELINE_ID}"
```
ondrejbudai, line 39: I would just put the arch into this variable to avoid repetition.

schutzbot/build.sh
```diff
@@ -7,11 +7,12 @@ function greenprint {
 }
 
 # Build the container and push to quay
+ARCH=$(uname -m)
 QUAY_REPO_URL="quay.io/osbuild/image-builder-test"
 QUAY_REPO_TAG="${CI_PIPELINE_ID}"
```
ondrejbudai, line 12: I would just put the arch into this variable to avoid repetition.

ondrejbudai (commented):
Great work! I have just minor nitpicks inline. :)

schutzbot/deploy.sh
diaasami: latest doesn't have multi-arch builds so it was just failing in this PR, and in addition, it is outdated and not built regularly.

schutzbot/deploy.sh
diaasami: This is better, I agree.

schutzbot/deploy.sh
ondrejbudai: Ah, thanks!

ondrejbudai (approved):
Let's go aarch64! Thanks a lot for this. ;)

PR #576, opened 2023-01-09
internal/v1: return empty array instead of null for collections

croissanne:
HMSIB-173


ondrejbudai (commented):
Maybe cover it with a tiny test? Otherwise LGTM. ;)

croissanne:
> Maybe cover it with a tiny test? Otherwise LGTM. ;)

indeed, I was testing out the jira integration

ondrejbudai:
hehe, sorry, I was too quick

ondrejbudai (approved):
lgtm, thanks!

PR #578, opened 2023-01-10
.github/workflows: use go 1.18

croissanne:
tools/prepare-source uses go 1.18 as well, and the ubi9 container has go 1.18.


kingsleyzissou (approved):
Cooool!! 🚀🚀

PR #581, opened 2023-01-11
internal/v1: Add distributuion repositories to an endpoint

regexowl:
This adds list of repositories to an `architectures/{distribution}` endpoint for the purpose of future deduplication of repository definitions between api and front-end.


internal/v1/api.yaml
```diff
@@ -680,6 +685,30 @@ components:
...
+        metalink:
+          type: string
+          format: uri
+        rhsm:
+          type: boolean
+          example: true
+        image_type_tags:
+          type: array
+          items:
+            type: string
```
croissanne, line 711: I wonder if we can just reuse the existing Repository definition. The id isn't strictly required for anything, it's just for the package list script. And I don't think we need to return image_type_tags at this time.

regexowl: If we don't need id and tags, then reusing the existing Repository definition should be doable. I'll take a look at it

regexowl: Done!

internal/v1/server_test.go
```diff
@@ -202,10 +202,40 @@ func TestWithoutOsbuildComposerBackend(t *testing.T) {
...
+					}, {
+						Baseurl: common.StringToPtr("http://mirror.centos.org/centos/8-stream/extras/x86_64/os/"),
+						Rhsm:    false,
+					}, {
+						Baseurl: common.StringToPtr("https://packages.cloud.google.com/yum/repos/google-compute-engine-el8-x86_64-stable"),
+						Rhsm:    false,
+					}, {
+						Baseurl: common.StringToPtr("https://packages.cloud.google.com/yum/repos/cloud-sdk-el8-x86_64"),
+						Rhsm:    false,
+					},
```
croissanne, line 221: I think it's best that we only return repositories which are valid for all image types (so which don't have `image_type_tags`), we can potentially add a query parameter in future which allows you to query the repos for a specific image type.

croissanne (changes requested):
One more change, otherwise lgtm.

internal/v1/api.yaml
```diff
@@ -356,6 +357,10 @@ components:
           items:
             type: string
             example: 'qcow2'
+        repositories:
```
ondrejbudai, line 360: May I ask for a short `description` for the new property? Since the API is public for the customers and the spec is the only documentation we have, it would be great to document it as much as possible. Thanks! :)

regexowl: Absolutely, short description added :)

internal/v1/server_test.go
regexowl: Got it! Updated, now it returns only repositories that have no image type tags :)

croissanne (approved):

PR #588, opened 2023-01-16
Accept source for aws

croissanne:


internal/v1/api.yaml
```diff
@@ -588,15 +588,19 @@ components:
...
           example: ['123456789012']
           items:
             type: string
           uniqueItems: true
+        share_with_sources:
+          type: array
+          example: ['12345']
+          items:
+            type: string
+          uniqueItems: true
```
atodorov, line 603: So none of these 2 options are going to be required from now on. Is that intentional?

atodorov (commented):
Added a question.

internal/v1/api.yaml
croissanne: Yes, but I need to add a check to have at least 1 of them specified.

atodorov (dismissed):
LGTM

internal/provisioning/client.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	return pc.client.Do(req)
+}
+
+func (pc *ProvisioningClient) GetAccountID(sourceID string, identity string) (*http.Response, error) {
+	return pc.request("GET", fmt.Sprintf("%s/sources/%s/account_identity", pc.url, sourceID), map[string]string{
+		"x-rh-identity": identity,
+	}, nil)
+}
```
lzap, line 46: Is there a reason why you don't pass context.Context into this function so you do not need to care about identity argument. Assuming you put identity into the context at the beginning of the request. Then you can simply do:

    req.Header.Set("X-RH-Identity", identity.GetIdentityHeader(ctx))

In fact, you don't need to do this in every method of this client, you can actually implement a `Doer` single-method interface and decorate the `ProvisioningClient` in the `NewClient` function. After that, you don't need to worry about identity ever again for our app.

We use `Doer` for trace context, however, we still add individual headers in our own codebase, I would like to change that :-)

internal/v1/server.go
```diff
@@ -25,7 +26,8 @@ import (
 
 type Server struct {
 	echo       *echo.Echo
-	client     *composer.ComposerClient
+	cclient    *composer.ComposerClient
+	pClient    *provisioning.ProvisioningClient
```
lzap, line 30: Inconsistent naming scheme.

lzap (approved):
Small nitpicking.

By the way, our app supports open telemetry tracing, if you want to enjoy a nice inter-app logging searching, all you need to do is:

* Generate and store trace id in the context [for each request](https://github.com/RHEnVision/provisioning-backend/blob/main/internal/middleware/trace_id.go).
* Put the trace id from context to [all logging messages](https://github.com/RHEnVision/provisioning-backend/blob/main/internal/middleware/logger.go#L25).
* Implement `Doer` interface that will return open telemetry [HTTP transport](https://github.com/RHEnVision/provisioning-backend/blob/main/internal/clients/http/platform_client.go#L31).
* Use the transport in the provisioning client for all calls.
* Trace id will be passed via HTTP headers into our app where it will be part of all logs and also passed further down (sources).

I know this is a bit off, however, this is the initial provisioning client implementation and I thought it would be nice to have this awesome feature.

internal/provisioning/client.go
croissanne: Good point, I tried to just copy what we do for the composer client. But that's much cleaner.

internal/v1/server.go
croissanne: I'll change that :flushed:

lzap (approved):
LGTM TY

croissanne:
Fixed the templates, and dependency name :facepalm: https://github.com/osbuild/image-builder/compare/558a7019a95d9fec751fe70104e9791a6dbe6788..f6a1fa19d0cd0490d29045044d072f611a47f55f

regexowl (dismissed):
LGTM

jrusz:
I'm sorry I had to push a change in `pr_check` to use a different environment, however the tests should be fixed now.

atodorov (approved):

lucasgarfield (approved):

PR #596, opened 2023-01-23
internal/v1: return empty array instead of null

jabia99:
the endpoint /composes/{id}/clones returned null instead of empty array if there were no clones.
HMS-163


croissanne (approved):
lgtm, ty!

PR #625, opened 2023-02-15
test/README.md: remove random paths

croissanne:
These paths included my own home directory, leftover from long ago.
 Setting some composer secrets is also required now, even if it's just
    bogus values.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -12,14 +12,9 @@ Run `make build` and migrate the db (this only needs to be done once) to the lat
 
 ```
 PGHOST=localhost PGPORT=5432 PGUSER=postgres PGPASSWORD=foobar PGDATABASE=imagebuilder
-MIGRATIONS_DIR=/home/sanne/workstuff/image-builder/internal/db/migrations/ ./image-builder-migrate-db
+MIGRATIONS_DIR=$HOME/image-builder/internal/db/migrations/ ./image-builder-migrate-db
 ```
 
-Start Image Builder:
-```
-PGHOST=localhost PGPORT=5432 PGUSER=postgres PGPASSWORD=foobar PGDATABASE=imagebuilder ALLOWED_ORG_IDS="*"
-DISTRIBUTIONS_DIR="/home/sanne/workstuff/image-builder/distributions" ./image-builder
-```
 
 To compose images locally, and if you have Composer running locally, add these to the environment:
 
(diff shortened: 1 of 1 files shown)


test/README.md
```diff
@@ -12,14 +12,9 @@ Run `make build` and migrate the db (this only needs to be done once) to the lat
 
 ```
 PGHOST=localhost PGPORT=5432 PGUSER=postgres PGPASSWORD=foobar PGDATABASE=imagebuilder
-MIGRATIONS_DIR=/home/sanne/workstuff/image-builder/internal/db/migrations/ ./image-builder-migrate-db
+MIGRATIONS_DIR=/$HOME/image-builder/internal/db/migrations/ ./image-builder-migrate-db
```
regexowl, line 15: It seems `$HOME` already contains the first slash, so I'd remove the one after the equal sign.

regexowl (changes requested):

test/README.md
```diff
@@ -28,10 +23,21 @@ OSBUILD_URL="https://localhost:8085" OSBUILD_CA_PATH="/etc/osbuild-composer/ca-c
...
 
+Otherwise set some bogus values:
+```
+COMPOSER_TOKEN_URL=bogus COMPOSER_CLIENT_ID=bogus COMPOSER_CLIENT_SECRET=bogus
+```
+
+Start Image Builder:
+```
+PGHOST=localhost PGPORT=5432 PGUSER=postgres PGPASSWORD=foobar PGDATABASE=imagebuilder
+DISTRIBUTIONS_DIR="/$HOME/image-builder/distributions" ./image-builder
```
regexowl, line 34: Is it possible to set the bogus value in a command separate from the one to start Image Builder?

croissanne: Well  it's implied that you should add those when invoking image-builder. So You'd need to just add them, or export them.

regexowl: I'm asking because I probably wouldn't get that if I haven't tried it before. Maybe `Start Image Builder with the paths or bogus values` could help to be more explicit? Not entirely sure so just adding that as a thought.

regexowl: But otherwise looks good to me :)

regexowl (approved):

test/README.md
croissanne: hm, it's just a bash thing tho, idk

PR #635, opened 2023-02-23
internal/v1: Allow specifying sources when cloning an ami

croissanne:
When cloning an image that was shared with a source, the clone should also be shared with the source.


lucasgarfield (approved):
Great work! I tested this using the ephemeral environment with the necessary changes to image-builder-frontend and can confirm that everything is working.

PR #666, opened 2023-03-30
image-builder: Limit total image size for AWS and Azure

croissanne:
This limits the size of AWS and Azure images to 10GiB total per image. The other image types are thinly/dynamicly provisioned, meaning it's fine to not check the limit for those.

HMS-1554


supakeen (approved):
I'd have prefered this to be a single change and the move to `validateCustomizations` to be separate. 

However, the *filesystem* checks look correct.

diaasami (approved):

PR #672, opened 2023-04-06
internal/v1: accept source to share an Azure image with

lucasgarfield:
This commit follows 019d2c8 which added support for sharing AWS images with sources.

The /sources/{sourceId}/account_identity endpoint, which was used to resolve the account id for AWS images shared with sources, is deprecated.

This commit removes the deprecated endpoint, replacing it with the new upload_info endpoint which is used for both AWS and Azure images.


lucasgarfield:
I am waiting until I have manually verified this is working on the ephemeral environment to open it for review...

lucasgarfield:
Ready for review... I have verified that the 'old' method of creating an Azure still works on the ephemeral environment. Given that and that the tests, I'm confident this PR doesn't break anything. It is harder to test whether or not creating an Azure image using sources works, because the services running on the ephemeral environment are not authorized to query our Azure account... I think we will just have to verify that it works once it hits stage.

PR #694, opened 2023-04-25
templates: Fix template for stage post-deploy end-to-end HMS testing

tpapaioa:
This PR makes a couple fixes the e2e test template:
- Only run tests once, in case of failure
- Pull the latest testing image each time
- Run tests against stage-beta environment


croissanne (dismissed):
Thanks!

croissanne (approved):

PR #696, opened 2023-04-26
Add custom repository support

kingsleyzissou:
This depends on https://github.com/osbuild/osbuild-composer/pull/3405# getting merged.


internal/composer/openapi.v2.yml
```diff
@@ -822,6 +822,50 @@ components:
...
+      properties:
+        id:
+          type: string
+        name:
+          type: string
+        filename:
+          type: string
+        baseurl:
+          type: array
+          example: [ 'https://cdn.redhat.com/content/dist/rhel8/8/x86_64/baseos/os/' ]
```
kingsleyzissou, line 843: I was in two minds about making this an array (since we changed it to an array in composer).
Maybe we could keep it as a string here and then convert it to an array internally so that this
and `payload_repositories` have a similar structure?

croissanne (approved):

PR #707, opened 2023-05-04
internal/v1: add compose request to status

lzap:
In order to be able to launch Azure, we need to know which resource group (Azure term) the image was shared or uploaded into. This is stored in the database along with the compose request. After discussion with @croissanne it turns out that it should be possible to add this information to the existing endpoint.

From the performance standpoint, no extra external REST calls are performed and in the initial version of my patch, not even an extra DB roundtrip is made since the account check function actually pulls all data (including the required JSON field) to only check for existence. I am taking advantage and updating the function to actually return the data. If you ask me, I would probably prefer having the check function to only check for existence which should be always faster in expense of doing two calls for this particular case.

I have to say we are short on time with this, I would appreciate quick review and possible merge with stage push as this is really sub-par user experience for Azure. Monday is holiday here, so I think the only reasonable window to get this in is tomorrow. Ping me for faster feedback loop, thank you!


.gitignore
```diff
@@ -4,3 +4,5 @@ rpmbuild
 /image-builder-migrate-db
 /image-builder-migrate-db-tern
 dnf-json
+/.fleet
+/.idea
```
lzap, line 8: Not too relevant but I am on JetBrains tools.

Makefile
```diff
@@ -6,6 +6,18 @@ build:
...
+	go test ./...
+
+.PHONY: generate
+generate:
+	go generate ./...
+
+.PHONY: fmt
+fmt:
+	go fmt ./...
+
```
lzap, line 20: Two of these are in the `tools/prepare-source.sh` tho I really was not searching in there.

internal/v1/handler.go
```diff
@@ -224,6 +224,13 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
 		return httpError
 	}
 
+	// TODO GetCompose
+	var composeRequest ComposeRequest
+	err = json.Unmarshal(composeEntry.Request, &composeRequest)
+	if err != nil {
+		return err
+	}
+
```
lzap, line 233: The TODO is a left over, will remove.

internal/v1/server_test.go
```diff
@@ -119,6 +119,7 @@ func startServerWithCustomDB(t *testing.T, url, provURL string, dbase db.DB, dis
 	require.NoError(t, err)
 
 	echoServer := echo.New()
+	echoServer.HideBanner = true
```
lzap, line 122: Spent just an hour testing the code and I am already sick of that banner taking all the vertical space :-)

internal/v1/handler.go
```diff
@@ -300,7 +308,7 @@ func parseComposeStatusError(composeErr *composer.ComposeStatusError) *ComposeSt
 }
 
 func (h *Handlers) GetComposeMetadata(ctx echo.Context, composeId uuid.UUID) error {
-	err := h.canUserAccessComposeId(ctx, composeId)
+	_, err := h.canUserAccessComposeId(ctx, composeId)
```
lzap, line 311: Granted this is sub-ideal, I offer refactoring into special SQL later as a separate PR. I think the fasted way would be something like `SELECT ID FROM COMPOSES WHERE XXX LIMIT 1` I guess. Maybe `COUNT`.

go.sum
```diff
@@ -405,6 +406,7 @@ golang.org/x/tools v0.0.0-20191119224855-298f0cb1881e/go.mod h1:b+2E5dAYhXwXZwtn
 golang.org/x/tools v0.0.0-20200103221440-774c71fcf114/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
 golang.org/x/tools v0.0.0-20201224043029-2b0845dc783e/go.mod h1:emZCQorbCU4vsT4fOWvOPXz4eW1wZW4PmDk9uLelYpA=
 golang.org/x/tools v0.1.10/go.mod h1:Uh6Zz+xoGYZom868N8YTex3t7RhtHDBrE8Gzo9bV56E=
+golang.org/x/tools v0.1.12 h1:VveCTK38A2rkS8ZqFY25HIDFscX5X9OoEhJd3quQmXU=
 golang.org/x/tools v0.1.12/go.mod h1:hNGJHUnrk76NpqgfD5Aqm5Crs+Hm0VOH/i9J2+nxYbc=
 golang.org/x/xerrors v0.0.0-20190410155217-1f06c39b4373/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
 golang.org/x/xerrors v0.0.0-20190513163551-3ee3066db522/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
```
lzap, line 412: Hmmm is this caused by different Go version? I am on 1.18.0.

On our project we do not enforce specific Go version as long as tests pass on CI and containers are built via ubi-golang build image it really does not matter much. Sure, couple of go sums here and there but that is it. I am eager to hear more why this is enforced.

ondrejbudai: Running `tools/prepare-source.sh` will fix this. This script is meant to align the project's code to our standards.

Makefile
ondrejbudai: If we want these targets in the `Makefile`, they should be running in a similar way as `tools/prepare-source.sh` does: Enforcing a specific version of Go. Otherwise, they might not actually format the code according to the same standards that our CI enforces.

I'm actually against adding these, `tools/prepare-source.sh` already does everything we need. One potential improvement might to be to add a `prep` target to the `Makefile` calling `tools/prepare-source.sh`, but I don't see that much value in it.

internal/v1/handler.go
```diff
@@ -359,21 +367,21 @@ func (h *Handlers) GetComposeMetadata(ctx echo.Context, composeId uuid.UUID) err
 }
 
 // return an error if the user does not have the composeId associated to its OrgID in the DB, nil otherwise
-func (h *Handlers) canUserAccessComposeId(ctx echo.Context, composeId uuid.UUID) error {
+func (h *Handlers) canUserAccessComposeId(ctx echo.Context, composeId uuid.UUID) (*db.ComposeEntry, error) {
```
ondrejbudai, line 370: `canUserAccessComposeId` no longer describes well what the method does - it now returns the compose, instead of just checking its access rights. Also, the comment on the line above must be changed.

internal/v1/handler.go
ondrejbudai: Looks correct to me.

internal/v1/handler.go
ondrejbudai: Sounds good!

internal/v1/server_test.go
ondrejbudai: Might be better to do this in a separate commit. I suggest putting the github comment into the commit message. Git is eternal, github might not be.

.gitignore
ondrejbudai: Again, this might be better as a separate commit, but since it's pretty obvious what this does, I don't insist on it. :)

ondrejbudai (changes requested):
Looks good. If we want this to merge this quickly, I suggest dropping all unneeded changes (mainly the ones in `Makefile`) and focusing just on what's important. For me, the only blocker is the incorrectly named method `canUserAccessComposeId`.

internal/v1/api.yaml
```diff
@@ -382,6 +382,8 @@ components:
       properties:
         image_status:
           $ref: '#/components/schemas/ImageStatus'
+        compose_request:
```
croissanne, line 385: In `ComposesResponseItem:` we just call it `request`, so let's call it the same thing?

Also it can be a required property. Adding required properties in responses is fine.

croissanne:
> Oh no I just pushed with a small changes (`go.mod` changes reverted and gitignore newline). Sorry I noticed you started with review!

that's ok, I only have the 1 comment in addition to what Ondřej said. And force push as many times as you want :)

internal/v1/api.yaml
lzap: Problem is, the JSON Go tag is actually set to `compose_request`:

```
type ComposeStatus struct {
	ComposeRequest *ComposeRequest `json:"compose_request,omitempty"`
	ImageStatus    ImageStatus     `json:"image_status"`
}
```

Is it worth creating a new struct?

lzap:
Force pushed:

* Both Makefile/gitignore and banner hide changes are in separate commits.
* Introduced new function `getComposeByIdAndOrgId` and updated the existing `canUserAccessComposeId` to make use of it. This way both semantics are kept clear yet we take the advantage of just a single DB query when we can.
* Added `compose_request` into `required` statement in OpenAPI, but I am keeping the name for cleaner code. Can rename to `request` if needed, new type needs to be introduced unless I want to do some raw JSON processing. Spoiler alert: I don’t want. :-D

internal/v1/api.yaml
croissanne: the JSON go tag is generated based on the name of the field in the openapi spec though.

internal/v1/api.go
```diff
@@ -212,7 +212,8 @@ type ComposeResponse struct {
 
 // ComposeStatus defines model for ComposeStatus.
 type ComposeStatus struct {
-	ImageStatus ImageStatus `json:"image_status"`
+	ComposeRequest *ComposeRequest `json:"compose_request,omitempty"`
```
croissanne, line 215: after marking it as required, it should no longer be a pointer. `tools/prepare-source.sh` needs to be called again.

lzap: Oh right I haven’t noticed that. We generate OpenAPI from Go. Rebased.

croissanne: Can we still rename it to `request` though, I really don't think it's a big change.
```
diff --git a/internal/v1/api.yaml b/internal/v1/api.yaml
index 7f6a812..0e16f2b 100644
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -379,11 +379,11 @@ components:
     ComposeStatus:
       required:
         - image_status
-        - compose_request
+        - request
       properties:
         image_status:
           $ref: '#/components/schemas/ImageStatus'
-        compose_request:
+        request:
           $ref: "#/components/schemas/ComposeRequest"
     ImageStatus:
       required:
diff --git a/internal/v1/handler.go b/internal/v1/handler.go
index 5f09de1..7ca9d27 100644
--- a/internal/v1/handler.go
+++ b/internal/v1/handler.go
@@ -241,7 +241,7 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
                        Status:       ImageStatusStatus(cloudStat.ImageStatus.Status),
                        UploadStatus: nil,
                },
-               ComposeRequest: &composeRequest,
+               Request: composeRequest,
        }
 
        if cloudStat.ImageStatus.UploadStatus != nil {

```

lzap: Oh right it works, I thought I am unmarshaling it as a whole but actually this will work. Updated.

Makefile
croissanne: I agree here, either we could have a `make prepare-source` target, but we need to enforce the go version each time.

lzap:
All passing!

croissanne:
Just the makefile thing, we have to enforce go versions.

lzap:
Ah missed that one, dropped from the first commit.

croissanne (approved):
Thanks!

ondrejbudai (approved):
Lovely, thanks!

PR #711, opened 2023-05-09
mark a compose as deleted

croissanne:


kingsleyzissou:
Oh neat! I verified it locally and it all works, this is great. Did you leave it in draft because you're going to add more to it or so it doesn't get merged just yet?

kingsleyzissou (approved):

PR #712, opened 2023-05-09
Rhel 92

croissanne:


achilleas-k:
Do the tests need to be adjusted for the expected values for the distro name/alias or is the behaviour in composer an issue?

achilleas-k:
> Do the tests need to be adjusted for the expected values for the distro name/alias or is the behaviour change an issue?

Never mind.  Commit answered my question :)

mshriver:
The IQE tests do not need to be updated for 9.2, and have passed in pr-checks.

achilleas-k (approved):
LGTM

PR #713, opened 2023-05-09
run server unit tests against sqlite DB

croissanne:
followup to #711


croissanne:
The alternative is that we depend on psql also during unit tests, and spin up a containter during the test itself.

PR #714, opened 2023-05-09
distribution/Dockerfile-ubi: pin go toolset to go 1.18.10

croissanne:
The latest 1.19 toolset breaks with:
```
RUN go install ./...
    error obtaining VCS status: exit status 128
        Use -buildvcs=false to disable VCS stamping.
    error obtaining VCS status: exit status 128
    	Use -buildvcs=false to disable VCS stamping.
```


bcl (approved):
Sounds like a good idea :)

PR #766, opened 2023-06-26
templates/dashboards: show the last 6 hours by default

croissanne:
The SLO section always shows the last 28 days, and the operational section is only interesting over a shorter period of time, so reflect that in the default time and internal settings.


kingsleyzissou (approved):
Yeah this is a sensible change. lgtm

PR #768, opened 2023-06-27
Enable OpenSCAP customizations

kingsleyzissou:
Depends on https://github.com/osbuild/osbuild-composer/pull/3522


internal/v1/api.yaml
```diff
@@ -967,6 +969,18 @@ components:
...
+    OpenSCAP:
+      type: object
+      required:
+        - profile_id
+        - datastream
+      properties:
+        profile_id:
+          type: string
+          example: "xccdf_org.ssgproject.content_profile_cis"
+        datastream:
```
teg, line 981: Are you sure this should be exposed? Ref our discussion in Brno?

croissanne:
took the liberty of rebasing

PR #782, opened 2023-07-07
Use psql db in unit tests

croissanne:


achilleas-k (dismissed):
This is good.  I like this.

achilleas-k (approved):
Still great.

kingsleyzissou (dismissed):
Oh, very nice!!

internal/tutils/psql.go
```diff
@@ -0,0 +1,136 @@
...
+func containerRuntime() (string, error) {
+	out, err := exec.Command("which", "podman").Output()
+	if err == nil {
+		return strings.TrimSpace(string(out)), nil
+	}
+	out, err = exec.Command("which", "docker").Output()
+	if err == nil {
+		return strings.TrimSpace(string(out)), nil
+	}
+	return "", err
```
ondrejbudai, line 35: nitpick: I would add a message that no container runtime was found and that the user needs to have one.

internal/tutils/psql.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil {
+			time.Sleep(time.Second * 1)
+			continue
+		}
+		break
+	}
+	return p, nil
+}
+
+func (p *PSQLContainer) execCommand(args []string) (string, error) {
```
ondrejbudai, line 87: nitpick: `args ...string` would make the API similar to the one `exec.Command` has.

internal/tutils/psql.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	p := &PSQLContainer{
+		name: name,
+		id:   strings.TrimSpace(string(out)),
+		port: port,
+	}
+
+	tries := 0
+	for tries < 10 {
```
ondrejbudai, line 72: Maybe return an error explicitly if the container failed to start? Might be overcautious, though...

internal/tutils/psql.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+type PSQLContainer struct {
+	name string
+	id   string
+	port int
+}
+
+const (
+	image string = "postgres:13"
```
ondrejbudai, line 22: `podman` prefers fully-qualified names, so I suggest making this `docker.io/library/postgres:13`.

internal/v1/server_test.go
```diff
@@ -26,7 +26,23 @@ import (
...
+func TestMain(m *testing.M) {
+	d, err := tutils.NewPSQLContainer()
+	if err != nil {
+		panic(err)
+	}
+
+	dbc = d
+	code := m.Run()
+
+	err = dbc.Stop()
```
ondrejbudai, line 40: I wonder whether we can somehow put this into a `defer` block. I would love making this as error-proof as possible, as I don't like the sound of dangling database containers on my system.

ondrejbudai (commented):
Lovely! Fewer code paths is always amazing. :)

ondrejbudai (approved):
Amazing! :)

croissanne:
going to force merge, dependencies on eph are broken :(

PR #786, opened 2023-07-12
distributions: add rhel 8 and 9 nightlies

croissanne:
These have restricted access and no package lists. They will only be buildable on stage.


ondrejbudai (approved):
Thanks! :heart_eyes:

PR #792, opened 2023-07-24
add description field to api

mgold1234:
This commit add a description field to the backend side to give us the ability to add it in the frontend side as well.
related to issue on the frontend side - 
https://github.com/RedHatInsights/image-builder-frontend/issues/1244


kingsleyzissou (changes requested):
Nice, thanks for this. I think it might make sense to make the DB changes in the same PR

kingsleyzissou (approved):

kingsleyzissou (approved):

PR #799, opened 2023-08-11
Add image size to api

bcl:
and a couple other small code cleanups.

But this needs #791 (or something like it that pulls in the cloudapi yaml that includes the new size option). This has a WIP commit that copies it over temporarily to see if the tests will run. Do not merge it, I'll remove it once #791 has been merged.


internal/common/pointers.go
```diff
@@ -1,9 +1,5 @@
...
 
-func StringToPtr(s string) *string {
-	return &s
-}
-
-func BoolToPtr(b bool) *bool {
-	return &b
+func ToPtr[T any](x T) *T {
+	return &x
 }
```
croissanne, line 5: Excellent, been wanting to do this for a while ^_^

croissanne (approved):

croissanne (changes requested):
Actually can we also add a check in `func validateCustomizations(cr *ComposeRequest) error`  to make sure the requested size is less than `FSMaxSize` ? AWS images are raw, so we should avoid exorbitant sizes.

croissanne (approved):
thanks!

PR #802, opened 2023-08-15
add name field to repository structure in api

mgold1234:
This commit add a name field to the backend side to give us the ability to add it in the frontend side as well.
related to issue on the frontend side -
https://github.com/RedHatInsights/image-builder-frontend/issues/1305


kingsleyzissou:
@lucasgarfield I'm not too familiar how the repositories are saved on the frontend. Is this a similar case to #792 where we could get the info from the image request? Or will we need to update the database too

croissanne:
Don't we get the name from the content-sources service?

mgold1234:
@croissanne thank you very much, you are correct I can get the name from the content -source, I will close this pr.

PR #810, opened 2023-08-30
api: embed the parent compose ID in clone response

lavocatt:
The frontend uses RTKQ to perform its requests to the IB's backend, and we're using programmatically generated requests to fetch the data over. RTKQ allows us to enhance these automatically generated endpoints and add a few behavioral customizations like cache tagging and cache invalidation.

We are planning on using cache tagging and invalidation to enhance the user experience by allowing clones to get refreshed only in certain circumstances.

However, in the current state of the API we do not let have access to the parent composeID of a clone at the CloneResponseItem level, meaning that our tag invalidation will invalidat too broadly.

Adding the parent composeID to the clone data that is return will allow us to only refetch the clone data associated to a given composeID and only regenerate parts of the screen associated with that data.


internal/v1/api.yaml
```diff
@@ -1023,12 +1023,17 @@ components:
     ClonesResponseItem:
       required:
         - id
+        - parent_compose_id
```
croissanne, line 1026: Can we just call it compose_id?

croissanne (commented):
small suggestion re: naming, otherwise lgtm

internal/v1/api.yaml
lavocatt: works for me!

croissanne (approved):

PR #814, opened 2023-09-01
Detail restrictions in the api spec

croissanne:
Makes the /distributions, /packages, and /architecture/{distribution} check the restrictions on distributions.


ondrejbudai (approved):
Love it!

PR #821, opened 2023-09-13
internal: Record email addresses

regexowl:
This adds an email column to the `composes` table.

### DIFF
--- a/cmd/image-builder-db-test/main_test.go
+++ b/cmd/image-builder-db-test/main_test.go
@@ -26,6 +26,8 @@ const (
 	ORGID2 = "100001"
 	ORGID3 = "100002"
 
+	EMAIL1 = "user1@test.test"
+
 	fortnight = time.Hour * 24 * 14
 )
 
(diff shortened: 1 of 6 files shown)


cmd/image-builder-db-test/main_test.go
```diff
@@ -26,6 +26,10 @@ const (
 	ORGID2 = "100001"
 	ORGID3 = "100002"
 
+	EMAIL1 = "user1@test.test"
+	EMAIL2 = "user2@test.test"
```
croissanne, line 30: EMAIL 2 and 3 aren't actually used :sweat_smile:

croissanne (changes requested):
one nitpick, lgtm otherwise

cmd/image-builder-db-test/main_test.go
regexowl: But three is a lucky number :shrug: 
I think I've added the constants first and then just used the same number as ANR and ORGID, forgot to check if I actually use all three. It's fixed now :sweat_smile:

internal/db/migrations-tern/001_create_table_composes.sql
```diff
@@ -3,6 +3,7 @@ CREATE TABLE IF NOT EXISTS composes(
        request jsonb NOT NULL,
        created_at timestamp NOT NULL,
        account_number varchar NOT NULL,
+       email varchar,
```
croissanne, line 6: Oh wait, this won't work actually since table already exists, this will never get executed during the migration.

You need to add a new file 005_add_email_column.sql with
```
ALTER TABLE composes ADD email varchar;
```

croissanne (changes requested):
woops, noticed something about the migration

croissanne (approved):

croissanne:
Thank you!

croissanne (approved):

croissanne (approved):

PR #834, opened 2023-09-21
OpenSCAP autotailoring

kingsleyzissou:
Add the ability to specify selected and and selected rules in the
OpenSCAP customizations. This allows users to create a tailoring file
which extends the base profile.


internal/v1/api.yaml
```diff
@@ -1019,6 +1019,28 @@ components:
...
+          items:
+            type: string
+        unselected:
+          type: array
+          example: [ 'xccdf_org.ssgproject.content_rule_rpm_verify_hashes' ]
+          description: |
+            Rules to be excluded from the scan and remediation
+            The rule ID can be either full, or just the suffix
+          items:
+            type: string
```
croissanne, line 1043: will this be ok with the tailoring profiles we'll get from the compliance service?

kingsleyzissou: It's kind of complicated, they can generate a tailoring xml file only. Getting that into the image would be the challenge

kingsleyzissou:
Waiting for the compliance `v2` api

internal/v1/api.yaml
kingsleyzissou: Circling back to this, I think compliance and the UX team don't want the ability to tailor compliance policies in the service (which is fair).

I was chatting to @lucasgarfield and I think maybe we could make a clear distinction between compliance and OpenSCAP. Then we could enable this for OpenSCAP but hide it for Insights Compliance - in the frontend of course.

lucasgarfield: As I understand it, OpenSCAP adoption in the service by customers will probably be limited until we expose tailoring.

Let's try to get tailoring exposed in the service via API as a first step. Once it is in place, we can figure out how we want to expose it through the UI... and I think that @kingsleyzissou's idea to enable it for OpenSCAP and hide it for Compliance is probably a great idea.

If I am right and we capture that customer usage of OpenSCAP + tailoring greatly exceeds that of the Compliance service, we can share with the Compliance team and let them re-evaluate. But for now, let's not let anything block us from shipping highly demanded features to customers.

kingsleyzissou:
I'll close this for now and when the compliance api is ready, we can re-think how to approach things.

PR #839, opened 2023-09-26
v1: cleanup OpenAPI documentation

lzap:
Since I was reviewing the API, I noticed many endpoints are missing either "summary", or "description" or even "id". In general, summary should be really short, ideally a title because it is used by many documentation generators as the title for the endpoint.

So I shortened summaries and made sure they match the operation id (e.g. "getComposes" is "get composes"), also added missing elements and description are now sentences.

Finally, I added tags which are super useful not only for browsing generated documentation, but also in client generators (you can easily skip unwanted endpoints).

Also adding a link to ReDoc: an online OpenAPI generation tool.

This is how the API looks before the patch:

https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/osbuild/image-builder/main/internal/v1/api.yaml#tag/Source/operation/getSourceUploadInfo

And this after the patch:

https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/lzap/image-builder/84780c9ad101413bf9dc314575103f6c0eb49c22/internal/v1/api.yaml

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -15,9 +15,11 @@ servers:
 paths:
   /version:
     get:
-      summary: get the service version
-      description: "get the service version"
+      summary: get version
+      description: "Get the service version."
       operationId: getVersion
+      tags:
+        - meta
       responses:
         '200':
           description: a service version
@@ -27,8 +29,11 @@ paths:
                 $ref: '#/components/schemas/Version'
   /ready:
     get:
-      summary: return the readiness
+      summary: return readiness
+      description: "Return the service readiness."
       operationId: getReadiness
+      tags:
+        - meta
       responses:
         '200':
           description: readiness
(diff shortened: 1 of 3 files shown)


internal/v1/api.yaml
```diff
@@ -98,14 +115,14 @@ paths:
             default: 100
             minimum: 1
             maximum: 100
-          description: max amount of composes, default 100
+          description: "Max amount of composes, default 100."
```
bcl, line 118: Does the description really need to also include the default? I see that the ReDoc service also shows the default so it's not needed there -- but can other API viewers be expected to do the same thing?

bcl (approved):
Looks good to me, more docs are always a good thing!

lzap:
Resolved conflicts, removed "default" remarks from descriptions.

lzap:
Rebased, can we get this in then? Thanks.

PR #843, opened 2023-09-27
v1: break handlers into individual files

lzap:
Because soon we will be working on bluprints, I would like to cleanup the api package a bit.

Initially I thought I would break this package into separate packages like middleware, server and handlers but everything is heavily tied together that I think more reasonable approach is to break down `handlers.go` into individual files a bit so we can add `handler_blueprint.go` later on when we start.

I am following the pattern from OpenAPI cleanup PR where I introduced few new OpenAPI tags:

* meta
* package
* architecture
* compose

Therefore I am following this naming in this proposal:

* handler_common.go (common functions)
* handler_compose.go (still quite big but better now)
* handler_package.go
* handler_distribution.go (and architecture together)

Other than moving function among files within the same package, no changes.


lzap:
Unlucky. I will resolve conflicts once we get to an agreement, there are currently no PRs touching these files so it is probably good time to do this now.

PR #848, opened 2023-09-29
api: expose oscap customizations in the API

lavocatt:
As part of the effort to create a new step in the wizard for hardened images we need to have the back end able to communicate what are the filesystem and packages customization an image needs to have.
This PR does exactly this:
* add an endpoint to get the list of available policies to apply on a distribution
* add an endpoint to get the Customizations for that policy on that distribution
* add a script to generate these Customizations

Not in the scope of this PR, but a nice followup:
* a bot to propose automated PRs when the Customizations are needing an update

Open questions:
* How do we make sure to avoid forgetting to update the list of valid policies for a given distribution ?


internal/v1/handler.go
```diff
@@ -1215,3 +1215,69 @@ func closeBody(body io.Closer) {
...
+		fallthrough
+	case Rhel84:
+		fallthrough
+	case Rhel85:
+		fallthrough
+	case Rhel86:
+		fallthrough
+	case Rhel87:
+		fallthrough
+	case Rhel88:
```
kingsleyzissou: So technically, I think anything past  RHEL 8.6 and RHEL 9.1 can take on OpenSCAP customizations.

internal/v1/handler.go
```diff
@@ -1215,3 +1215,69 @@ func closeBody(body io.Closer) {
...
+			XccdfOrgSsgprojectContentProfileE8,
+			XccdfOrgSsgprojectContentProfileHipaa,
+			XccdfOrgSsgprojectContentProfileIsmO,
+			XccdfOrgSsgprojectContentProfileOspp,
+			XccdfOrgSsgprojectContentProfilePciDss,
+			XccdfOrgSsgprojectContentProfileStig,
+			XccdfOrgSsgprojectContentProfileStigGui,
+		})
+	default:
+		return echo.NewHTTPError(http.StatusBadRequest, "No profile for the specified distribution")
```
kingsleyzissou: It would be nice to have `centos-8` and `centos-9` too

oscap/rhel8/xccdf_org.ssgproject.content_profile_anssi_bp28_enhanced/customizations.json
```diff
@@ -0,0 +1 @@
+{"filesystem":[{"min_size":1073741824,"mountpoint":"/boot"},{"min_size":1073741824,"mountpoint":"/home"},{"min_size":1073741824,"mountpoint":"/opt"},{"min_size":1073741824,"mountpoint":"/srv"},{"min_size":1073741824,"mountpoint":"/tmp"},{"min_size":5368709120,"mountpoint":"/usr"},{"min_size":3221225472,"mountpoint":"/var"},{"min_size":5368709120,"mountpoint":"/var/log"},{"min_size":10737418240,"mountpoint":"/var/log/audit"},{"min_size":1073741824,"mountpoint":"/var/tmp"}],"packages":["aide","sudo","dnf-automatic","rsyslog-gnutls","chrony"]}
```
kingsleyzissou, line 1: Nitpick: it can we try add a new line at the end of the file?

kingsleyzissou:
Thanks for this, just for the record from our call: in an ideal world we would be able to get some of this information from the compliance team and so this could be a solution in the interim. Will try set up a meeting with @croissanne and them next week when the person I know there is back from PTO.

internal/v1/api.yaml
```diff
@@ -1111,3 +1129,28 @@ components:
...
+        xccdf_org.ssgproject.content_profile_cis_server_l1,
+        xccdf_org.ssgproject.content_profile_cis_workstation_l1,
+        xccdf_org.ssgproject.content_profile_cis_workstation_l2,
+        xccdf_org.ssgproject.content_profile_cui,
+        xccdf_org.ssgproject.content_profile_e8,
+        xccdf_org.ssgproject.content_profile_hipaa,
+        xccdf_org.ssgproject.content_profile_ism_o,
+        xccdf_org.ssgproject.content_profile_ospp,
+        xccdf_org.ssgproject.content_profile_pci-dss,
+        xccdf_org.ssgproject.content_profile_standard,
```
kingsleyzissou, line 1178: Standard is only supported by `fedora` so maybe we can remove it.
For reference: https://www.osbuild.org/guides/image-builder-on-premises/oscap-remediation.html

oscap/rhel8/xccdf_org.ssgproject.content_profile_anssi_bp28_enhanced/customizations.json
lavocatt: I'll have to check, because it's `json.Marshall` that is doing the writing. Is that possible with the tool ?

internal/v1/handler.go
lavocatt: Do you think it's worth changing the switch case for something else ?

internal/v1/handler.go
lavocatt: They should be considered as `rhel8` and `rhel9` right ?

internal/v1/handler.go
croissanne: Since the deployment has the most recent version of composer, wouldn't rhel84/5 with openscap work as well?

kingsleyzissou: Yeah it will work too

internal/v1/handler.go
kingsleyzissou: There is weirdly a difference between `centos-8` and `rhel-8`. Same for `centos-9` and `rhel-9`

oscap/rhel8/xccdf_org.ssgproject.content_profile_anssi_bp28_enhanced/customizations.json
kingsleyzissou: Good question, you might be able to append `\n` to the buffer before writing to the file

internal/v1/handler.go
lavocatt: All done :)

internal/v1/handler.go
lavocatt: All done :)

internal/v1/api.yaml
lavocatt: Ah, maybe we can keep it there ? I mean it's only in the possible profiles, true we can't use it yet, but maybe it's worth it being exhaustive ?

oscap/rhel8/xccdf_org.ssgproject.content_profile_anssi_bp28_enhanced/customizations.json
lavocatt: I've done that. Good idea thanks.

internal/v1/oscap.go
```diff
@@ -0,0 +1,109 @@
...
+			XccdfOrgSsgprojectContentProfileStigGui,
+		}, nil
+	case Rhel90:
+		fallthrough
+	default:
+		return nil, errors.New("No profile for the specified distribution")
+	}
+}
+
+func OscapDistributionGeneraization(distribution Distributions) (string, error) {
```
kingsleyzissou, line 76: Nitpick: OscapDistributionGeneraization -> OscapDistributionGeneralization

internal/v1/handler.go
```diff
@@ -1223,3 +1226,27 @@ func (h *Handlers) GetOscapProfiles(ctx echo.Context, distribution Distributions
...
 	return ctx.JSON(http.StatusOK, profiles)
 }
+
+func (h *Handlers) GetOscapCustomizations(ctx echo.Context, distribution Distributions, profile DistributionProfileItem) error {
+	globalDist, err := OscapDistributionGeneraization(distribution)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, err)
+	}
+	//Load the json file with the customizations
+	jsonFile, err := os.Open(path.Join(h.server.OscapDir, filepath.Base(globalDist), filepath.Base(string(profile)), "customizations.json"))
```
kingsleyzissou, line 1236: Maybe we could delegate the file reading and parsing to the `oscap` package? Kind of like we do with `distributions`,
this would tidy this up a bit

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/jackc/pgx/v4 v4.18.1
 	github.com/labstack/echo/v4 v4.10.2
 	github.com/labstack/gommon v0.4.0
+	github.com/pelletier/go-toml v1.9.5
```
ondrejbudai, line 14: I wonder if we should use github.com/BurntSushi/toml instead. We use this one in composer. The API should be very similar.

lavocatt:
> Adding a new top-level directory seems like an overkill for me, because it adds more visual clutter to the repository, and we need to pass yet another variable through the whole stack (`OSCAP_DIR` and its variations).

I can see why having another variable might be slightly annoying, on the other hand, can we consider that changing that at this stage might require a redesign of a bunch of elements in this PR and that since there are time constraints associated to the global functionality being delivered, it might not be optimal ?

Also, another point to consider is that we don't want to keep these endpoints anyway. The short term goal is to provide a PoC that IB can be used to generate hardened images via the frontend. Then as a mid-term goal to get the compliance team to serve these endpoints instead of us.

This entire PR, to call their tool to generate customizations to serve, doesn't really make sense in IB anyway, but we need that shortcut to deliver fast.

> Since the customizations are always tied to a distribution, cannot we just put them under `distributions/$DISTRO/oscap`? I see that some customizations are shared amongst multiple distributions (currently solved with the concept of "distro generalization"). We can use symlinks instead (or not care at all, the jsons are small).

It's not the storage that is an issue but the generation that takes a very long time. So when you're developing the utility and running it a bunch of times in a row, you want that to be rather quick. That's why calling it only for the necessary distribution is a plus.

With that said, if you want me to spend some time changing where/how these oscap are stored I can take care of it.

kingsleyzissou:
> Adding a new top-level directory seems like an overkill for me, because it adds more visual clutter to the repository, and we need to pass yet another variable through the whole stack (OSCAP_DIR and its variations).

I think since this is a temporary fix @ondrejbudai's feedback is actually more applicable. Passing a whole variable through the stack to then have remove it when we're able to query the compliance API feels like a lot. Whereas, if we implement this suggestion and put the configurations in the distro, it would be easier to remove at a later stage. It does mean generating more files, but I think the trade-off makes sense.

lavocatt:
> > Adding a new top-level directory seems like an overkill for me, because it adds more visual clutter to the repository, and we need to pass yet another variable through the whole stack (OSCAP_DIR and its variations).
> 
> I think since this is a temporary fix @ondrejbudai's feedback is actually more applicable. Passing a whole variable through the stack to then have remove it when we're able to query the compliance API feels like a lot. Whereas, if we implement this suggestion and put the configurations in the distro, it would be easier to remove at a later stage. It does mean generating more files, but I think the trade-off makes sense.

Alright guys, let's implement that.

cmd/oscap/main.go
```diff
@@ -0,0 +1,162 @@
...
+		panic(err)
+	}
+	fmt.Println("✓")
+}
+
+// This program needs as an argument the directory to the distributions root file
+func main() {
+	var distributionsFolder = os.Args[1]
+	distros, err := distribution.LoadDistroRegistry(distributionsFolder)
+	distros.Available(true).List()
```
ondrejbudai, line 130: I think this can be removed.

internal/v1/oscap.go
```diff
@@ -73,37 +78,29 @@ func OscapProfiles(distribution Distributions) (DistributionProfileResponse, err
...
-		fallthrough
-	case Rhel9Nightly:
-		return "rhel9", nil
-	case Centos8:
-		return "centos8", nil
-	case Centos9:
-		return "cs9", nil
-	default:
-		return "", errors.New("No customizations for the specified distribution")
+func loadOscapCustomizations(distributionDir string, distribution Distributions, profile DistributionProfileItem) (*Customizations, error) {
```
ondrejbudai, line 81: My suggestion is to load the oscap customizations when we load [the distributions](https://github.com/osbuild/image-builder/blob/3e09617ce00040a5d4c61fc4948c963719f5c6e2/internal/distribution/distribution.go#L165) on startup and just save them as a map in [`DistributionItem`](https://github.com/osbuild/image-builder/blob/3e09617ce00040a5d4c61fc4948c963719f5c6e2/internal/distribution/distribution.go#L18).

This has several advantages:
- No I/O in the handler.
- No passing `distributionDir` to the server.
- The customization's existence and validity (if it's a valid JSON) is checked during launch. This means that if there's something up with them, the container will crash immediately instead of behaving weirdly during runtime.

The disadvantage is higher memory usage, but they are tiny, this is fine.

internal/v1/oscap.go
```diff
@@ -0,0 +1,109 @@
...
+			XccdfOrgSsgprojectContentProfileStigGui,
+		}, nil
+	case Rhel90:
+		fallthrough
+	default:
+		return nil, errors.New("No profile for the specified distribution")
+	}
+}
+
+func OscapDistributionGeneralization(distribution Distributions) (string, error) {
```
ondrejbudai: This method is added in one commit and removed in the next one. While this has no effect on the resulting merged code, it makes the history harder to read.

internal/v1/oscap.go
```diff
@@ -0,0 +1,109 @@
+package v1
+
+import (
+	"errors"
+)
+
+func OscapProfiles(distribution Distributions) (DistributionProfileResponse, error) {
```
ondrejbudai, line 12: This method has a similar issue: One commit adds it, and the following one moves it to a separate file. Once again, this has no effect on the result, but makes the history harder to read.

ondrejbudai (approved):
Amazing, thanks for taking extra time to make this more maintainable long-term, I'm very happy with the result.

I put some suggestions inline, but since none of them affects the resulting experience, I'm comfortable merging this PR as it is. :)

PR #851, opened 2023-10-04
templates: Include image name in FloorPlan

avitova:
Hello,
I am working on statistics of image builder vs launch.
if I understand from your database files correctly, this should match our [image_id](https://github.com/RHEnVision/provisioning-backend/blob/259c6f32122d1623d1e1591d1429fa5fb421c479/deploy/clowdapp.yaml#L461) column. :)
I hope this is what we are going for. :)

### DIFF
--- a/templates/image-builder.yml
+++ b/templates/image-builder.yml
@@ -143,7 +143,7 @@ objects:
     - prefix: ${FLOORIST_QUERY_PREFIX}/builds
       query: >-
         select
-          job_id,created_at,org_id,account_number,email,request->>'distribution' as distribution,
+          job_id,created_at,org_id,account_number,email,image_name,request->>'distribution' as distribution,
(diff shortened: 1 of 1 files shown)


templates/image-builder.yml
```diff
@@ -143,7 +143,7 @@ objects:
     - prefix: ${FLOORIST_QUERY_PREFIX}/builds
       query: >-
         select
-          job_id,created_at,org_id,account_number,email,request->>'distribution' as distribution,
+          job_id,created_at,org_id,account_number,email,request,image_name->>'distribution' as distribution,
```
croissanne, line 146:
```suggestion
          job_id,created_at,org_id,account_number,email,image_name,request->>'distribution' as distribution,
```
'distribution' needs to be queried on the request, not on image_name.

croissanne (changes requested):

templates/image-builder.yml
avitova: True, thank you! Fixed. :)

croissanne (approved):

croissanne:
lgtm! thank you

PR #860, opened 2023-10-09
Expose oci image type

croissanne:


lavocatt:
@croissanne how can I test the PR?

kingsleyzissou (approved):
LGTM

croissanne:
> @croissanne how can I test the PR?

It's available in stage, so should be possible to run it.

lavocatt:
Are there needs for special credentials ?

internal/v1/api.yaml
```diff
@@ -711,7 +720,12 @@ components:
             - $ref: '#/components/schemas/AzureUploadRequestOptions'
     UploadTypes:
       type: string
-      enum: ['aws', 'gcp', 'azure', 'aws.s3']
+      enum:
+      - aws
+      - gcp
+      - azure
+      - aws.s3
+      - oci.objectstorage
```
lavocatt, line 728: The `UploadType` can be ` oci.objectstorage` but there's no corresponding options for it, if I want to create a request to test the functionality against stage, what should I have in the options ?

```
{
    "customizations": {
    "subscription": {
      "activation-key": "my-secret-key",
      "base-url": "http://cdn.redhat.com/",
      "insights": true,
      "organization": 2040324,
      "server-url": "subscription.rhsm.redhat.com"
    }
  },
  "distribution": "centos-8",
  "image_requests": [
    {
      "architecture": "x86_64",
      "image_type": "oci",
      "upload_request": {
        "type": "oci.objectstorage",
        "options":{
        }
      }
    }
  ]
}

```

Even in the case where the `options` object should be empty for the oci object storage, shouldn't there, for coherency, be a corresponding schema anyway ?

lavocatt: After giving it a couple of thoughts, the request above validates because `AWSS3UploadRequestOptions` is an empty object. So there should be a `OciObjectStorageOptions` for the request validation to be more coherent.

lavocatt (changes requested):

kingsleyzissou (approved):

PR #861, opened 2023-10-10
api: create and delete blueprint endpoint

lzap:
This is my proposal for the create blueprint endpoint. It is 1:1 copy of the payload from the create compose request. Let’s discuss the API first, then I will start working on the code and DDL.

### DIFF
--- a/cmd/image-builder-db-test/main_test.go
+++ b/cmd/image-builder-db-test/main_test.go
@@ -4,7 +4,9 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"github.com/osbuild/image-builder/internal/v1"
 	"os/exec"
 	"testing"
 	"time"
@@ -64,8 +66,8 @@ func migrateTern(t *testing.T) {
 		"--host", c.PGHost, "--port", c.PGPort,
 		"--user", c.PGUser, "--password", c.PGPassword,
 		"--sslmode", c.PGSSLMode)
-	err := cmd.Run()
-	require.NoError(t, err)
+	output, err := cmd.CombinedOutput()
+	require.NoErrorf(t, err, "tern command failed with non-zero code, combined output: %s", string(output))
 }
 
 func connect(t *testing.T) *pgx.Conn {
@@ -77,10 +79,14 @@ func connect(t *testing.T) *pgx.Conn {
 func tearDown(t *testing.T) {
 	conn := connect(t)
 	defer conn.Close(context.Background())
-	conn.Exec(context.Background(), "drop table clones")
-	conn.Exec(context.Background(), "drop table composes")
-	conn.Exec(context.Background(), "drop table if exists schema_migrations")
-	conn.Exec(context.Background(), "drop table if exists schema_version")
+	_, err := conn.Exec(context.Background(), "drop schema public cascade")
+	require.NoError(t, err)
+	_, err = conn.Exec(context.Background(), "create schema public")
+	require.NoError(t, err)
+	_, err = conn.Exec(context.Background(), "grant all on schema public to postgres")
+	require.NoError(t, err)
+	_, err = conn.Exec(context.Background(), "grant all on schema public to public")
+	require.NoError(t, err)
 }
 
 func testInsertCompose(t *testing.T) {
@@ -342,7 +348,41 @@ func testClones(t *testing.T) {
 	require.Equal(t, clones[1], *entry)
 }
 
-func TestMain(t *testing.T) {
+func testBlueprints(t *testing.T) {
+	d, err := db.InitDBConnectionPool(connStr(t))
+	require.NoError(t, err)
+	conn := connect(t)
+	defer conn.Close(context.Background())
+
+	b1 := v1.BlueprintV1{
+		Version:     1,
+		Name:        "name",
+		Description: "desc",
+	}
+	body, err := json.Marshal(b1)
+	require.NoError(t, err)
+
+	id := uuid.New()
+	err = d.InsertBlueprint(id, ORGID1, ANR1, "name", "desc", 1, body)
+	require.NoError(t, err)
+
+	entry, err := d.GetBlueprint(id, ORGID1, ANR1)
+	require.NoError(t, err)
+	b2, err := v1.BlueprintFromEntry(entry)
+	require.NoError(t, err)
+	require.Equal(t, b1, b2)
+
+	err = d.DeleteBlueprint(id, ORGID1, ANR1)
+	require.NoError(t, err)
+}
+
+func runTest(t *testing.T, f func(*testing.T)) {
+	migrateTern(t)
+	defer tearDown(t)
+	f(t)
+}
+
+func TestAll(t *testing.T) {
 	fns := []func(*testing.T){
 		testInsertCompose,
 		testGetCompose,
@@ -350,11 +390,10 @@ func TestMain(t *testing.T) {
 		testGetComposeImageType,
 		testDeleteCompose,
 		testClones,
+		testBlueprints,
 	}
 
 	for _, f := range fns {
-		migrateTern(t)
-		f(t)
-		tearDown(t)
+		runTest(t, f)
 	}
 }
--- a/cmd/image-builder-db-test/main_test.go
+++ b/cmd/image-builder-db-test/main_test.go
@@ -6,10 +6,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/osbuild/image-builder/internal/v1"
 	"os/exec"
 	"testing"
 	"time"
+
+	v1 "github.com/osbuild/image-builder/internal/v1"
 
 	"github.com/google/uuid"
 	"github.com/jackc/pgx/v4"
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -4,9 +4,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/labstack/gommon/random"
-	"github.com/sirupsen/logrus"
-
 	"github.com/osbuild/image-builder/internal/common"
 	"github.com/osbuild/image-builder/internal/composer"
 	"github.com/osbuild/image-builder/internal/config"
@@ -16,11 +13,13 @@ import (
 	"github.com/osbuild/image-builder/internal/provisioning"
 	v1 "github.com/osbuild/image-builder/internal/v1"
 
-	"github.com/labstack/echo/v4"
-	"github.com/labstack/echo/v4/middleware"
-
 	"github.com/getsentry/sentry-go"
 	sentryecho "github.com/getsentry/sentry-go/echo"
+	"github.com/labstack/echo/v4"
+	"github.com/labstack/echo/v4/middleware"
+	"github.com/labstack/gommon/log"
+	"github.com/labstack/gommon/random"
+	"github.com/sirupsen/logrus"
 )
 
 func main() {
@@ -134,6 +133,10 @@ func main() {
 	if conf.GlitchTipDSN != "" {
 		echoServer.Use(sentryecho.New(sentryecho.Options{}))
 	}
+	// log stack traces into standard logger as error (instead of stdout)
+	echoServer.Use(middleware.RecoverWithConfig(middleware.RecoverConfig{
+		LogLevel: log.ERROR,
+	}))
 	if conf.IsDebug() {
 		echoServer.Debug = true
 	}
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -5,9 +5,9 @@ import (
 	"strings"
 
 	"github.com/labstack/gommon/random"
-	"github.com/osbuild/image-builder/internal/common"
 	"github.com/sirupsen/logrus"
 
+	"github.com/osbuild/image-builder/internal/common"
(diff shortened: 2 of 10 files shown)


lzap:
Ping @croissanne any comments before I start?

croissanne:
> Ping @croissanne any comments before I start?

Go for it :)

lzap:
Rebased the code logic for inserting the record into the database. As part of the PR, I am introducing new type named `BlueprintV1` and a version flag that is stored in a separate column. This allows taking advantage of types in Go as I expect us to modify the schema a bit and unstructured data in SQL database is always a problem: existing records on stage will carry old versions which can be easily "upgraded".

Other than that, it is just a copy-paste.

internal/db/migrations-tern/007_blueprint_composes.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE IF NOT EXISTS blueprint_composes
+(
+    blueprint_id uuid REFERENCES blueprints(id) ON DELETE CASCADE,
+    compose_id uuid REFERENCES composes(job_id) ON DELETE CASCADE
```
croissanne, line 4: Does it make sense to enforce the compose_id to be unique in this table?

croissanne (commented):
Overall this lgtm

internal/db/migrations-tern/007_blueprint_composes.sql
lzap: Good point, it does. Now that I am thinking about this table, I am a bit ahead as creating composes through the new experimental API is not part of this PR, but if you don’t mind I will go ahead, add unique constraint and leave it here.

lzap:
Ok rebased and added UNIQUE constraint.

internal/v1/api.yaml
```diff
@@ -383,6 +383,37 @@ paths:
...
+          application/json:
+            schema:
+              $ref: "#/components/schemas/CreateBlueprintRequest"
+      responses:
+        '201':
+          description: blueprint was saved
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ComposeResponse'
```
lzap, line 404: Ha I should be returning blueprint id instead. Do not merge.

lzap:
So I have also created new endpoint `/blueprints/UUID/compose` which forwards the creation to the existing code. I had to slightly refactor one handler, I introduced `composeImage` which is called from `ComposeImage` as well as `ComposeBlueprint` except the latter uses input data from the database.

cmd/image-builder-db-test/main_test.go
```diff
@@ -342,19 +348,52 @@ func testClones(t *testing.T) {
...
+		testBlueprints,
 	}
 
 	for _, f := range fns {
-		migrateTern(t)
-		f(t)
-		tearDown(t)
+		runTest(t, f)
 	}
 }
```
lzap, line 399: Renamed because `TestMain` is reserved for test suites, it does work but Goland is marking this as an error and it is pesky. Also refactored it to function so `defer` can be used to ensure all data is deleted after even an error.

lzap, line 399: Renamed because TestMain is reserved for test suites, it does work but Goland is marking this as an error and it is pesky. Also refactored it to function so defer can be used to ensure all data is deleted after even an error.

internal/tutils/psql.go
```diff
@@ -49,6 +49,8 @@ func NewPSQLContainer() (*PSQLContainer, error) {
 	out, err := exec.Command(
 		rt,
 		"run",
+		"--mount=type=tmpfs,destination=/var/lib/postgresql/data",
+		"--mount=type=tmpfs,destination=/dev/shm",
```
lzap, line 53: This significantly speeds up on one of my systems (old Xeon with HDD). From 336s to just 36s.

internal/tutils/psql.go
```diff
@@ -87,6 +89,9 @@ func (p *PSQLContainer) execCommand(args ...string) (string, error) {
 	}
 	/* #nosec G204 */
 	out, err := exec.Command(rt, args...).CombinedOutput()
+	if err != nil {
+		err = fmt.Errorf("command %s %s error: %w, output: %s", rt, args, err, out)
+	}
```
lzap, line 94: The PR contains couple of error printing, during initial setup I had problems (missing `podman` or misconfiguration there) and it was hard to troubleshoot without these.

internal/tutils/psql.go
```diff
@@ -124,9 +129,9 @@ func (p *PSQLContainer) NewDB() (db.DB, error) {
 		"--user", user,
 	)
 
-	_, err = cmd.Output()
+	out, err := cmd.CombinedOutput()
```
lzap, line 132: This was the culprint - `Output()` is missing standard error completely, so the error was actually hidden.

lzap:
This is now ready for re-review, I removed "compose blueprint" endpoint, this is only create operation (with get/delete DB operations). The rest will be followed in separate PRs. I am leaving few small changes in tests and db tests because I stuggled hard to set it up and with performance. @croissanne

lzap:
Resolved conflicts.

internal/v1/api.yaml
```diff
@@ -581,6 +612,44 @@ components:
...
+          $ref: '#/components/schemas/Distributions'
+        image_requests:
+          type: array
+          minItems: 1
+          maxItems: 1
+          items:
+            $ref: '#/components/schemas/ImageRequest'
+          uniqueItems: true
+          description: |
+            Array of exactly one image request. Having more image requests in a single blueprint is currently not supported.
```
lucasgarfield, line 643: This is probably leftover from a copy/paste... shouldn't there be a request per target?

internal/v1/api.yaml
```diff
@@ -581,6 +612,44 @@ components:
...
+        description:
+          type: string
+          example: "My blueprint description"
+          maxLength: 250
+        distribution:
+          $ref: '#/components/schemas/Distributions'
+        image_requests:
+          type: array
+          minItems: 1
+          maxItems: 1
```
lucasgarfield, line 638: We need to support all target types, and that number will probably increase... maybe no max needed?

internal/db/migrations-tern/006_create_table_blueprints.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS blueprints
+(
+    id uuid PRIMARY KEY,
+    org_id varchar NOT NULL CHECK (coalesce(trim(org_id), '') != ''),
+    account_number varchar NOT NULL CHECK (coalesce(trim(account_number), '') != ''),
+    name VARCHAR(100) NOT NULL,
+    description VARCHAR(250),
+    body_version INTEGER NOT NULL,
+    body JSONB NOT NULL,
```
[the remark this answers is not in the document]
lzap: This is not blueprint version from the users perspective, this is body (payload) version which will be increased as soon as incompatible JSON change will be done. This must be stored outside of the JSON itself because only then you can effectively pick the proper type, `BlueprintV1` in this case, and unmarshal data into specific type.

Blueprint versioning will be done in a separate PR and that field will be likely called simply `version`, or we might end up only having `parent uuid` column. We talked about this with @avitova today on techdiss, it needs more research on what will be more appropriate for this case.

internal/v1/api.yaml
lzap: This is copy-pasted from the "create compose" endpoint without changes, the way I understand this is that this only accepts array with exactly one element and this is for the future (e.g. when multiple requests can be send). @croissanne ?

internal/v1/api.yaml
lzap: Ditto.

internal/db/migrations-tern/006_create_table_blueprints.sql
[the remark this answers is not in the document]
ezr-ondrej: Ah! Gotcha, than ignore my comment, this makes sense!

internal/v1/api.yaml
croissanne: So we'd generate multiple compose requests from the blueprint, so this will have to change in future for sure (but we can change stuff under /experimental so).

And then we also might arrive at a point where a single compose requests can have multiple upload targets, which will be the case for edge (s3 + pulp).

lzap: So if you ask me @lucasgarfield I'd say out of scope for this PR :-D

internal/v1/api.yaml
lzap: I suggest to do this separately after this is merged in both components.

ezr-ondrej: This should be different in Blueprint and Compose, here it should be 1:n, but compose stay at 1:1, we'd just make N composes when making composes out of blueprint (rebuilding blueprint images) or do I misunderstand something?

I'm not opposed to do it in follow up, I just don't get why 🤔 but sure, I can propose it myself 👍 I can have contribution!! 😛

lzap: I don’t understand, this is a datatype from "create compose" API request without any change. I believe that blueprint is just a "template" of what we have today. Therefore, I designed this endpoint in a way that blueprint can be resubmitted "as is" without any changes.

Maybe let’s have a chat on Slack to speed things up, I might be missing something.

lzap: Oh now I understand, multiple image requests can be in a single blueprint. This is different from what is now. Fixed.

lzap:
Rebased, this is now ready for re-review, thanks.

ezr-ondrej:
The tests seem to not agree with the swagger spec, I guess it needs to get updated? :thinking:

lzap:
Oh right, rebased.

ezr-ondrej:
We are :green_apple: :tada:

lzap:
Let's go @croissanne or @lucasgarfield we have more than one PRs to work on. Thanks.

cmd/image-builder/main.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/labstack/gommon/log"
```
croissanne, line 5: external imports tend to go in the 2nd paragraph, it's a bit of a mess in this file though.

internal/db/migrations-tern/006_create_table_blueprints.sql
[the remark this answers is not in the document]
croissanne: JSON unmarshaling is very forgiving though, so you could just grab the version and then unmarshal it again into the appropriate struct.

Once we build versioning on top we need a way to group the blueprints indeed.

internal/v1/api.yaml
croissanne: also I think it's fine to keep this as an array, for koji we ended up accepting multiple image requests per compose reqs so who knows where we need to go with this.

internal/db/db_blueprints.go
```diff
@@ -0,0 +1,68 @@
+package db
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/google/uuid"
```
croissanne, line 7: nitpick, space between built-in imports and external imports.

internal/tutils/psql.go
croissanne: would be nice so split out test improvements to a separate commit.

croissanne (commented):
Thank you! overal lgtm, some nitpicks

cmd/image-builder/main.go
lzap: I noticed, we use an automated tool to fix them all `goimports` it does this kind of sorting for us. I guess I will propose a new PR soon.

internal/db/db_blueprints.go
lzap: Fixing too.

lzap:
Rebased, I will do a followup on those nasty imports scattered around.

The test refactoring is unfortunately buried deep in the patch, I can do hunk by hunk split if you insist.

lzap:
All green.

lzap:
Here is the followup which I can rebase after this is merged: https://github.com/osbuild/image-builder/pull/895

cmd/image-builder/main.go
croissanne: right, can you move the gommon/log down though?

cmd/image-builder/main.go
```diff
@@ -134,6 +135,10 @@ func main() {
 	if conf.GlitchTipDSN != "" {
 		echoServer.Use(sentryecho.New(sentryecho.Options{}))
 	}
+	// log stack traces into standard logger as error (instead of stdout)
+	echoServer.Use(middleware.RecoverWithConfig(middleware.RecoverConfig{
+		LogLevel: log.ERROR,
+	}))
```
croissanne, line 141: Can you please split this into a separate commit?

lzap: Sure, done.

cmd/image-builder/main.go
lzap: Done.

lzap:
Some unrelated changes I guess?

/retest

ezr-ondrej:
/retest

> Some unrelated changes I guess?

[quai](https://quay.io/) was  down for the whole day ;)

lzap:
ALL TESTS GREEN.

lzap:
All green.

croissanne (approved):

PR #895, opened 2023-11-09
Formatting

lzap:
I noticed messy imports, so I propose `goimports` as well as `gofumpt` (a bonus). We find both tools not being intrusive. The patch introduces new script in the ./tools directory and moves formatting from prepare-source.sh.

### DIFF
--- a/tools/prepare-source.sh
+++ b/tools/prepare-source.sh
@@ -9,11 +9,15 @@ GO_BINARY=$(go env GOPATH)/bin/go$GO_VERSION
 go install golang.org/dl/go$GO_VERSION@latest
 $GO_BINARY download
 
+# Ensure dev tools are installed
+which goimports || $GO_BINARY install golang.org/x/tools/cmd/goimports@latest
+
 # Ensure that all code has been regenerated from its sources
 $GO_BINARY generate ./...
 
 # ... the code is formatted correctly, ...
-$GO_BINARY fmt ./...
+goimports -w internal cmd
+$GO_BINARY fmt ./internal/... ./cmd/...
 
 # ... and that go.mod and go.sum are up to date.
 $GO_BINARY mod tidy


tools/format-source.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+set -eux
+. "$(dirname "$0")/env.sh"
+
+goimports -w internal cmd
+gofumpt -l -w internal cmd
```
ondrejbudai, line 6: I wonder what's the support for this formatter in editors/IDEs. It's subpar compared to `gofmt`, this might affect developer experience negatively.

.github/workflows/tests.yml
```diff
@@ -23,9 +23,10 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v3.0.2
 
-      - name: Check that source has been prepared
+      - name: Check that source has been prepared and formatted
         run: |
           ./tools/prepare-source.sh
+          ./tools/format-source.sh
```
ondrejbudai, line 29: I don't want a separate script. It's nice that `tools/prepare-source.sh` does everything needed pre-commit related to formatting, modules and vendoring. We have this script everywhere, so creating a separate workflow in `image-builder` is confusing.

lzap: Thing is, if you want to reformat (fix) all the code in case CI complains, you would need to run the prepare-source.sh which is quite slow.

lzap:
So how about just `goimports`.

.github/workflows/tests.yml
ondrejbudai: 3 seconds cold start, 2 seconds hot start on my laptop. Doesn't feel that bad. :shrug: But I agree that this is very subjective.

ondrejbudai:
Yep, `goimports` is fairly reasonable. :)

lzap:
All green, fyi there is no code change related to `goimports` the code is clean at the moment :-)

ondrejbudai (approved):

PR #897, opened 2023-11-10
Add a new column `ClientId` to the database

regexowl:
This adds a new column with an id of a client via which a compose request was sent.

The logic is similar to the insertion of an image name. Frontend sends a request which contains a new value of `ClientId` set to "ui". Next the value is parsed to appear in a new column of the database.


internal/v1/api.yaml
```diff
@@ -697,6 +700,8 @@ components:
           type: string
         image_name:
           type: string
+        client_id:
+          type: string
```
croissanne, line 704: Can we make this an enum that allows on a select set of values?

regexowl: Most probably yes. It would contain only "ui" for now, but would be expandable. Unlike boolean which I was considering originally. :thinking: :+1:

regexowl:
This should hopefully do the trick. Got tangled in the pointers and types a bit.

croissanne:
The integration test needs a tweak, since the client_id has a default value now.

croissanne (changes requested):
lgtm, just a test tweak needed.

jrusz:
I fixed it in iqe plugin but we the MR/PR sync doesn't work without the env var I'm adding and I'm also trying to fix the api.sh test that's running in Schutzbot but I'm not sure if that's going to work, if not I'll do some manual testing to figure it out.

jrusz:
Ok so Schutzbot is fixed an I'm releasing a new iqe plugin, should be green in < 1h ... The sync is broken because of some cert issues in the Gitlab pipeline, I'll have to investigate later but the api update doesn't break testing on stage/prod so it was fine to merge.

croissanne (approved):
lgtm! thank you

PR #915, opened 2023-11-23
config: enable loading dotenv file

ezr-ondrej:
For local developement, enable having local.env file with config variables.

This is just for ease of local developement.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ dnf-json
 /.fleet
 /.idea
 /gen-oscap
+local.env
--- a/internal/config/utils.go
+++ b/internal/config/utils.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"reflect"
@@ -8,6 +9,9 @@ import (
 	"github.com/joho/godotenv"
 	clowder "github.com/redhatinsights/app-common-go/pkg/api/v1"
 )
+
+var ErrMissingEnvTag = errors.New("missing 'env' tag in config field")
+var ErrUnsupportedFieldType = errors.New("unsupported config field type")
 
 func LoadConfigFromEnv(conf *ImageBuilderConfig) error {
 	err := godotenv.Load("local.env")
@@ -23,7 +27,7 @@ func LoadConfigFromEnv(conf *ImageBuilderConfig) error {
 		fieldV := v.Field(i)
 		key, ok := fieldT.Tag.Lookup("env")
 		if !ok {
-			return fmt.Errorf("No env tag in config field")
+			return ErrMissingEnvTag
 		}
 
 		confV, ok := os.LookupEnv(key)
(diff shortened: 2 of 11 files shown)


internal/config/utils.go
```diff
@@ -1,14 +1,26 @@
...
 
+	"github.com/joho/godotenv"
 	clowder "github.com/redhatinsights/app-common-go/pkg/api/v1"
 )
 
+var ErrMissingEnvTag = errors.New("missing 'env' tag in config field")
+var ErrUnsupportedFieldType = errors.New("unsupported config field type")
+
 func LoadConfigFromEnv(conf *ImageBuilderConfig) error {
+	if _, err := os.Stat("local.env"); err == nil {
```
mvo5, line 17: (nitpick). I would just `if err := godotenv.Load(); err != nil && !os.IsNotExist(err) { return err}` here instead of first stat/open because doing it in two steps is a TOCTOU issue. Not a big deal here of course but I would avoid it out of the general principle.

mvo5 (commented):
Thanks for this PR! A quick drive-by comment inline and I would prefer if we had a small test for this new feature in config_test.go. But maybe the test can wait until we have a comment about the general feature from someone more experienced :)

.gitignore
```diff
@@ -8,3 +8,4 @@ dnf-json
 /.fleet
 /.idea
 /gen-oscap
+local.env
```
croissanne, line 11: nitpick, 1 newline at the end of each file

internal/config/utils.go
ezr-ondrej: I like your codesnipet, I've come up  with the same (similar), tested it and works fine :)

.gitignore
ezr-ondrej: ugh, new editor,  I've changed settings and fixed it :)

ezr-ondrej:
> I would prefer if we had a small test for this new feature in config_test.go

Generally I'd agree, but IMO in this case when it's not easy to test and the feature is not intended for prod, and current tests do test the prod path.
I'd say it's not necessary, but let me know if you insist, I'm happy to figure out how to test it :)

croissanne (commented):
Not sure if the named errors are necessary, since that would mean the `ImageBuilderConfig` struct is actually badly defined (no env tags and non-string values). Plus we only ever panic on these errors.

ezr-ondrej:
> Not sure if the named errors are necessary, since that would mean the `ImageBuilderConfig` struct is actually badly defined (no env tags and non-string values). Plus we only ever panic on these errors.

Necesary for sure not, I've just been trained by strict linter to always do it xD
But I can revert that, that's no biggie and should even be a separate commit at least :+1:
Let me put it in separate commit and once we decide I can keep it or remove it ;)

ezr-ondrej:
> Let me put it in separate commit and once we decide I can keep it or remove it ;)

Done, let me know what you prefer ;)

croissanne (approved):

PR #919, opened 2023-11-28
Add OpenSCAP profile name & description

kingsleyzissou:
Update the `oscap` tool to generate the name and description fields
for the OpenSCAP customisations. Some pain points to note:

- the blueprint description field contains the OpenSCAP
  profile's name
- the blueprint name field contains the OpenSCAP profile's
  id (or reference id)
- policy type is an alias for the profile name
- the reference id is an alias for the profile id


lucasgarfield:
This is looking great! As discussed on Slack, there's a minor issue now in that the "profile_id" is an empty string in the generated code. I'm just adding a blocked tag for now so we don't accidentally merge before the fix gets pushed.

PR #921, opened 2023-11-29
api: add compose blueprint endpoint

amirfefer:
This adds a new endpoint for composing a blueprint.
path:`/experimental/blueprint/{id}/compose`
Added a FK for reference a blueprint in compose table

Versioning association if needed will be added as a follow-up
The compose's name needs to be generated, for now I used just blueprint's name (unify the blueprint's name with a sequence number?)

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -37,6 +37,7 @@ type CloneEntry struct {
 
 type BlueprintEntry struct {
 	Id          uuid.UUID
+	VersionId   uuid.UUID
 	Version     int
 	Body        json.RawMessage
 	Name        string
@@ -44,7 +45,7 @@ type BlueprintEntry struct {
 }
 
 type DB interface {
-	InsertCompose(jobId uuid.UUID, accountNumber, email, orgId string, imageName *string, request json.RawMessage, clientId *string) error
+	InsertCompose(jobId uuid.UUID, accountNumber, email, orgId string, imageName *string, request json.RawMessage, clientId *string, blueprintVersionId *uuid.UUID) error
 	GetComposes(orgId string, since time.Duration, limit, offset int, ignoreImageTypes []string) ([]ComposeEntry, int, error)
 	GetCompose(jobId uuid.UUID, orgId string) (*ComposeEntry, error)
 	GetComposeImageType(jobId uuid.UUID, orgId string) (string, error)
@@ -62,8 +63,8 @@ type DB interface {
 
 const (
 	sqlInsertCompose = `
-		INSERT INTO composes(job_id, request, created_at, account_number, email, org_id, image_name, client_id)
-		VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4, $5, $6, $7)`
+		INSERT INTO composes(job_id, request, created_at, account_number, email, org_id, image_name, client_id, blueprint_version_id)
+		VALUES ($1, $2, CURRENT_TIMESTAMP, $3, $4, $5, $6, $7, $8)`
 
 	sqlGetComposes = `
 	        SELECT job_id, request, created_at, image_name, client_id
@@ -147,7 +148,7 @@ func InitDBConnectionPool(connStr string) (DB, error) {
 	return &dB{pool}, nil
 }
 
-func (db *dB) InsertCompose(jobId uuid.UUID, accountNumber, email, orgId string, imageName *string, request json.RawMessage, clientId *string) error {
+func (db *dB) InsertCompose(jobId uuid.UUID, accountNumber, email, orgId string, imageName *string, request json.RawMessage, clientId *string, blueprintVersionId *uuid.UUID) error {
 	ctx := context.Background()
 	conn, err := db.Pool.Acquire(ctx)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


ezr-ondrej (commented):
We are going through the code and we've identified two major issues:
- In the database compose is referencing Blueprint id, but not specific version, but compose is build for version of blueprint
- We are building just one compose, but we should build one compose per provider
- + The return should return all of the new composes

Some minor findings
- We could refactor the handler to not copy the common funcionality, but use some common function instead
Nitpicks:
- the migration is missing an trailing line

That said, it would be nice to split the DB change so we can get that in first.
WDYT? :)

lzap:
FYI please rebase on top of https://github.com/osbuild/image-builder/pull/923 thanks.

ezr-ondrej:
Also I've justnoticed, there is a typo in the commit message /bluepreint/blueprint/

amirfefer:
@lzap , @ezr-ondrej 

rebased
used a common handler 
used uuid.nil for insertCompose

lzap (changes requested):
Just a small nitpick, great work. Thanks.

ezr-ondrej (dismissed):
Code looks good to me.
But few changes are needed:
* there is a typo in commit message
* there is a conflict

internal/v1/handler_blueprints.go
```diff
@@ -71,3 +72,38 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
...
+	for _, imageRequest := range blueprint.ImageRequests {
+		composeRequest := ComposeRequest{
+			Customizations:   &blueprint.Customizations,
+			Distribution:     blueprint.Distribution,
+			ImageRequests:    []ImageRequest{imageRequest},
+			ImageName:        &blueprint.Name,
+			ImageDescription: &blueprint.Description,
+			ClientId:         &clientId,
+		}
+		composesResponse, err := h.handleCommonCompose(ctx, composeRequest, blueprintEntry.Id)
```
ezr-ondrej, line 101: We're passing the blueprint id here, should we pass version id instead?

Also adding a test for this handler could help.

amirfefer: indeed,  but we don't have the versioning implemented, right?

amirfefer:
@lzap, @ezr-ondrej  I found out that the `uuid.nil` fails on local DB tests:
```
violates foreign key constraint \"composes_blueprint_version_id_fkey\" (SQLSTATE 23503)
``` 
therefore, I changed the field to a pointer of UUID, and passed a `nil` instead.
Tests passed locally.

Also I rebased and fixed the typo in the commit message

internal/v1/handler_blueprints.go
ezr-ondrej: I'm not sure what you mean, but having corrupted data is definitelly wrong.
We do have versioning implemented to the level of having the first version of blueprint inserted.
We can easily adjust the BlueprintEntry to fetch that id as well.

amirfefer: I though that we want for wait a  bit until the versioning/editing will be implemented and than do a rebase, I didn't notice that the DB is already implemented. 

I added a `VersionId` field to `BlueprintEntry`, hopefully it answers this concern.

ezr-ondrej (approved):
Looks great to me, test verifying the pointer works as expected when provided with a value would be awesome, but I think there should not be an issue and I'm happy to add the test in a follow-up.

internal/db/db.go
```diff
@@ -37,14 +37,15 @@ type CloneEntry struct {
 
 type BlueprintEntry struct {
 	Id          uuid.UUID
+	VersionId   uuid.UUID
```
lzap, line 40: This column is NULLable, therefore you have to use pointer here:

https://github.com/osbuild/image-builder/blob/main/internal/db/migrations-tern/009_create_table_blueprint_versions.sql#L18C1-L18C117

Scanning a NULL value into non-pointer will lead to panic.

lzap (changes requested):
There is a bug, consider writing a test for this one.

This is what I expressed as a concern during our scrum last week and now I believe I understand the issue.

The `github.com/google/uuid` package is widely popular UUID package and it also implements a nullable type which is compatible with `database/sql` package from the standard library. The way it works in this SQL package is that each type must implement its own way of NULL values, typically via a `Valid` boolean field. This is pretty flexible and can be used for other values not just NULLs, but it sucks to work with.

Thing is, `image-builder` does not use `database/sql` but `pgx` and this driver handles NULL values in a different manner - by just using Go `nil`. This is more practical approach. You figured out that you have to use `nil` for NULLs and that is correct.

You can still use `uuid.Nil` with `pgx` like other code does in this repo, however, you are not storing a NULL value in a table, you are storing zero bytes (which is interpreted as not `Valid` value). Problem is, you cannot store this in a column that has UNIQUE constraint.

Hope this helps, please fix the bug and we are good.

internal/db/db.go
ezr-ondrej: I don't think the version can ever be nil :thinking: or can it?

ezr-ondrej: Oh, you're pointing to a Compose table, but this is a BlueprintEntry.
We do not  scan for the Compose blueprint_version_id column here, if I read correctly.

lzap (approved):
Correction: no problem there, that is a different table. Thanks @ezr-ondrej for pointing that out. I am good.

internal/v1/handler_blueprints.go
ezr-ondrej: Yup, that's perfect! :)

croissanne (approved):
neat, thank you!

PR #925, opened 2023-12-01
db: add composes org id index

lzap:
I noticed most queries filter records by org_id, however, there is no index on the column. With the current strategy of not deleting records from composes table, this might be a problem in the future causing too many table scans. This fixes it.

Warning: We now have three PRs with a migration, make sure we rename the numbers accordingly. Cheers. @croissanne


croissanne (approved):
lgtm!

lzap:
Just rebased, sorry. I accidentally created this index as unique, meaning a single customer account could only have a single compose :-) Fixed.

PR #928, opened 2023-12-05
HMS-3230 - v1: Add module_hotfixes flag

ezr-ondrej:
Allow passing module_hotfixes flag to both Custom and Payload repositories. This will allow passing the flag to composer, where it will disable modularity filtering on given repository.

Refs HMS-3230

Needs https://github.com/osbuild/osbuild-composer/pull/3831 to go in first


lzap (dismissed):
I don’t have the full context, but the patch looks sane.

ezr-ondrej:
Composer is in and I've added this to the integration test on AWS, so we are even testing this.
Looks like it's passing, @croissanne care taking a look? :)
Would be awesome to have it in stage already :)

ondrejbudai (approved):
:hot_pepper: :hammer: 

Thanks!

PR #930, opened 2023-12-06
v1: add all customizations which are available  in composer

croissanne:


internal/v1/api.yaml
```diff
@@ -1302,3 +1112,460 @@ components:
...
+          type: string
+          enum:
+            - raw
+            - lvm
+            - auto-lvm
+          description: |
+            Select how the disk image will be partitioned. 'auto-lvm' will use raw unless
+            there are one or more mountpoints in which case it will use LVM. 'lvm' always
+            uses LVM, even when there are no extra mountpoints. 'raw' uses raw partitions
+            even when there are one or more mountpoints.
```
kingsleyzissou, line 1202: Do we want to add `fips` now too? Or do that separately?

kingsleyzissou:
It looks good to me, I've tested some of the OpenSCAP stuff on top of this PR and it was okay. Guess it needs to be rebased and sort out the conflicts.

internal/v1/api.yaml
croissanne: oh that's new!

kingsleyzissou (dismissed):
LGTM

kingsleyzissou:
Huh:
 ```Problem: conflicting requests\n  - package nginx-module-njs-1:1.18.0.0.3.9-1.el8.ngx.aarch64 from 5f1ae8ef976a975d930d62623ffd8660e75dd01b70e79a9c9b75f95ccb2d7b0a requires nginx = 1:1.18.0-1.el8.ngx, but none of the providers can be installed```

lucasgarfield:
Thank you for the great work, @croissanne and @kingsleyzissou!!!

kingsleyzissou (approved):

PR #937, opened 2023-12-11
db: add function to GetBlueprintComposes

ezr-ondrej:
Refactors common funcionality to get generic composes and composes filtered by Blueprint.
It uses two functions, as the sql queries are quite different.

Refs HMS-3068

Currently depends on https://github.com/osbuild/image-builder/pull/921

PR #938, opened 2023-12-11
v1: Allow filtering composes by Blueprint

ezr-ondrej:
Add blueprint id parameter to /composes for filtering by Compose.
Also adds version parameter to filter composes for a specific blueprint version.

Refs HMS-3068

Depends on https://github.com/osbuild/image-builder/pull/937


internal/db/db.go
```diff
@@ -21,11 +21,13 @@ type dB struct {
...
-	ImageName *string
-	ClientId  *string
+	Id               uuid.UUID
+	BlueprintId      uuid.UUID
+	BlueprintVersion int
+	Request          json.RawMessage
+	CreatedAt        time.Time
+	ImageName        *string
+	ClientId         *string
 }
```
lzap, line 31: I am not a huge fan of adding fields from joined tables into the "entry" types. That being said, I am not that experienced yet with how this should be handled in Go. In my other project, I use embedding approach for result-only types:

```
type ComposeVersion struct {
	Compose
	Version
}
```

Then you simply:

```
cv := ComposeVersion{}
err = result.Scan(&cv.Compose.field1, &cv.Compose.field2, ...)
```

Embedded fields can be actually accessed directly: `&cv.field1` as long as they are unique. For `id` you will have to use the "long" form.

It also maybe makes sense to figure out a naming convention which would indicate this type is not an "entry" by not having `Entry` suffix. Meaning: you cannot insert it, this is just a result (of a join).

In fact, in my other project I use the Scanny library where you define Go struct tags and the library does the mapping automatically:

```
type ComposeVersion struct {
	Compose    `db:"c"`
	Version `db:"v"`
}
```

Then you do this:

```
result := &model.ComposeVersion{}
pgxscan.Get(ctx, Pool, result, "SELECT ...")

Unfortunately, postgres never returns column names with prefixes so these must be entered explicitly one by one:

```
SELECT
 c.id as "c.id",
 c.name as "c.name"
 FROM ... join c and v
```

This is only ugly for these joins, other than that Scanny is great.

File: internal/db/db_blueprints.go [hunk 2]
```diff
@@ -18,6 +19,23 @@ const (
...
+	sqlGetBlueprintComposes = `
+		SELECT composes.job_id, blueprint_versions.version, composes.request, composes.created_at, composes.image_name, composes.client_id
+		FROM composes INNER JOIN blueprint_versions ON composes.blueprint_version_id = blueprint_versions.id
+		WHERE composes.org_id = $1
+		  	AND blueprint_versions.blueprint_id = $2
+			AND CURRENT_TIMESTAMP - composes.created_at <= $3
+			AND ($4::text[] is NULL OR composes.request->'image_requests'->0->>'image_type' <> ALL($4))
+			AND composes.deleted = FALSE
+		ORDER BY composes.created_at DESC
+		LIMIT $5 OFFSET $6`
```
lzap, line 31: Not a huge fan of digging JSON via SQL but you know what, let’s give it a go and this can be always rewritten to client-side filtering OR a new field can be introduced if performance issues arises. Might not be a big deal, SQL optimizer will very likely schedule the scan as the last step.

internal/db/db_blueprints.go
```diff
@@ -18,6 +19,23 @@ const (
...
+
+	sqlCountActiveBlueprintComposesSince = `
+		SELECT COUNT(*)
+		FROM composes INNER JOIN blueprint_versions ON composes.blueprint_version_id = blueprint_versions.id
+		WHERE org_id=$1 AND blueprint_versions.blueprint_id = $2 AND CURRENT_TIMESTAMP - created_at <= $3 AND deleted = FALSE
+		AND ($4::text[] is NULL OR request->'image_requests'->0->>'image_type' <> ALL($4))`
```
lzap, line 37: Generic suggestion for all these long query strings with joins, you can rename tables to single char names:

```
SELECT c.job_id, c.version, c.request, c.created_at, c.image_name, c.client_id, ...
FROM composes AS c INNER JOIN blueprint_versions AS v
ON c.blueprint_version_id = v.id
```

internal/db/db.go
```diff
@@ -133,6 +136,10 @@ const (
 			WHERE composes.org_id=$2)`
 )
 
+func newComposeEntry(jobId uuid.UUID, blueprintId uuid.UUID, version int, request json.RawMessage, createdAt time.Time, imageName *string, clientId *string) ComposeEntry {
+	return ComposeEntry{Id: jobId, BlueprintId: blueprintId, BlueprintVersion: version, Request: request, CreatedAt: createdAt, ImageName: imageName, ClientId: clientId}
+}
```
lzap, line 141: Alert, alert. This is ugly. Why would you do this? :-D

Just use struct literal syntax, this is very anti Go idiomatic.

internal/db/db_blueprints.go
```diff
@@ -27,6 +45,46 @@ const (
...
+		var request json.RawMessage
+		var createdAt time.Time
+		var imageName *string
+		var clientId *string
+		err = result.Scan(&jobId, &version, &request, &createdAt, &imageName, &clientId)
+		if err != nil {
+			return nil, 0, err
+		}
+		composes = append(composes, newComposeEntry(jobId, blueprintId, version, request, createdAt, imageName, clientId))
+	}
```
lzap, line 74: Please scan into the result value directly, not via a temporary variables and the unnecessary function `newComposeEntry`.

```
var ce []ComposeEntry
...
err = result.Scan(
  &ce.Id,
  &ce.BlueprintId,
)
```

I intentionally created a multi-line form because I believe it is ultimately better.

internal/v1/handler.go
```diff
@@ -27,6 +27,21 @@ const (
...
+	}
+	return ComposesResponseItem{
+		CreatedAt: c.CreatedAt.Format(time.RFC3339),
+		Id:        c.Id,
+		ImageName: c.ImageName,
+		Request:   cmpr,
+		ClientId:  (*ClientId)(c.ClientId),
+	}, nil
+}
+
```
lzap, line 44: Don’t like it here either and I know we are doing this in our codebase in the payload package but that is a different use case: the function is exported (like here) and used in a different package (service) which is not the case here.

Unless you plan to use it in a different package, I suggest to stick with a struct literal.

internal/v1/handler.go
```diff
@@ -513,6 +513,32 @@ func convertIgnoreImageTypeToSlice(ignoreImageTypes *[]ImageTypes) []string {
 	return stringSlice
 }
 
+func (h *Handlers) fetchComposes(orgId string, blueprintId *uuid.UUID, blueprintVersion *int, since time.Duration, limit, offset int, ignoreImageTypeStrings []string) ([]db.ComposeEntry, int, error) {
+	if blueprintId != nil {
+		return h.server.db.GetBlueprintComposes(orgId, *blueprintId, since, limit, offset, ignoreImageTypeStrings)
+	}
+	return h.server.db.GetComposes(orgId, since, limit, offset, ignoreImageTypeStrings)
+}
```
lzap, line 521: Ah now I understand why you just sticked the new field into the "entry" struct :-)

A new struct will work here, you just return the Compose portion of the embedded type.

internal/v1/handler.go
```diff
@@ -513,6 +513,32 @@ func convertIgnoreImageTypeToSlice(ignoreImageTypes *[]ImageTypes) []string {
...
+		if blueprintVersion != nil {
+			path = fmt.Sprintf("%vblueprint_version=%d&", path, *blueprintVersion)
+		}
+	} else {
+		path = path + "?"
+	}
+	first := fmt.Sprintf("%voffset=0&limit=%v", path, limit)
+	last := fmt.Sprintf("%voffset=%v&limit=%v", path, lastOffset, limit)
+	return first, last
+}
```
lzap, line 540: Is it possible that I don’t see a unit test for this function? :-D

lzap (changes requested):
Few concerns about setting a new standard on how to represent joined data. I wonder what @croissanne thinks about it.

I don’t particularly like passing all the fields individually around in many functions, I wonder why the whole struct type was not used for this. But let’s keep it this way, refactoring could be done later.

internal/db/db_blueprints.go
ezr-ondrej: It's copy pasted from fetching composes, might be cood to use some common constant maybe, so once someone gets to refactor it, they'd refactor both :thinking:

internal/db/db_blueprints.go
ezr-ondrej: It's an array, so it might not be easey, but we could append first and then Scan into that appended var :shrug:

internal/v1/handler.go
ezr-ondrej: Nooo ! Busted ! :fearful: 
I can add one I guess :+1:

ezr-ondrej: I've added it in the handler test, but I've got rid of this fucntion now

internal/db/db_blueprints.go
ezr-ondrej: I find it more confusing then helpful, unless the short version is very clear.
Here where we actually work with the SQL, I prefer longer names for readability.
It's just my preference tho, if you think we should alias it, I'm happy to ;)

internal/db/db_blueprints.go
lzap: Yeah that slice was a typo in my example, what I meant was:

```
var ce ComposeEntry
...
err = result.Scan(
  &ce.Id,
  &ce.BlueprintId,
)
...
result = append(result, ce)
```

internal/v1/handler.go
lzap: I do not insist on anything here, just couple of ideas here and there.

internal/db/db_blueprints.go
lzap: Sure it is just a random thought.

internal/db/db_blueprints.go
```diff
@@ -4,11 +4,18 @@ import (
...
 
 	"github.com/google/uuid"
 	"github.com/jackc/pgx/v4"
 )
 
+type ComposeEntryWithVersion struct {
+	ComposeEntry
+	BlueprintId      uuid.UUID
+	BlueprintVersion int
+}
```
lzap, line 17: This is fine, slight disadvantage is what happens when you add another (few) fields from the other table. It will no longer be `ComposeEntryWithVersion` :-) Maybe can you remove `Entry` from the type name? To me "entry" means it is a literal record (entry) while this is not.

internal/db/db_blueprints.go
```diff
@@ -27,6 +51,44 @@ const (
...
+	if err != nil {
+		return nil, 0, err
+	}
+	defer conn.Release()
+	result, err := conn.Query(ctx, sqlGetBlueprintComposes, orgId, blueprintId, since, ignoreImageTypes, limit, offset)
+	defer result.Close()
+
+	if err != nil {
+		return nil, 0, err
+	}
```
lzap, line 66: You have to defer close call after `nil` check, because in case of an error `result` will actually be `nil` = panic.

internal/db/db_blueprints.go
```diff
@@ -27,6 +51,44 @@ const (
...
+
+	var count int
+	err = conn.QueryRow(ctx, sqlCountActiveBlueprintComposesSince, orgId, blueprintId, since, ignoreImageTypes).Scan(&count)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	return composes, count, nil
+}
```
lzap, line 90: This function is responsible for two things, list and count. I suggest to split it into two db functions which can be called separately and possibly reused, there is no performance benefit in doing it this way.

internal/v1/handler.go
```diff
@@ -498,6 +498,89 @@ func convertIgnoreImageTypeToSlice(ignoreImageTypes *[]ImageTypes) []string {
...
+		})
+	}
+
+	lastOffset := count - 1
+	if lastOffset < 0 {
+		lastOffset = 0
+	}
+	first := fmt.Sprintf("%v/v%v/composes?offset=0&limit=%v", RoutePrefix(), specVersion, limit)
+	last := fmt.Sprintf("%v/v%v/composes?offset=%v&limit=%v", RoutePrefix(), specVersion, lastOffset, limit)
+	return newComposeResponse(data, count, first, last), nil
```
lzap, line 544: I wonder if the string building should be part of the helper function, it is gonna be always the same.

internal/v1/handler.go
```diff
@@ -498,6 +498,89 @@ func convertIgnoreImageTypeToSlice(ignoreImageTypes *[]ImageTypes) []string {
...
+			Count int `json:"count"`
+		}{
+			count,
+		},
+		Links: struct {
+			First string `json:"first"`
+			Last  string `json:"last"`
+		}{first, last},
+	}
+}
```
lzap, line 514: I would prefer two functions and types for meta and links so one could use struct literal to build the response but insert the metadata in easy way:

```
result := ComposesResponse{
  Data: // ...
  Meta: newMetaData(count, first, last),
  Links: newMetaLinks(count, first, last),
}
```

That seems more reusable than one function per response type.

internal/v1/handler.go
```diff
@@ -498,6 +498,89 @@ func convertIgnoreImageTypeToSlice(ignoreImageTypes *[]ImageTypes) []string {
...
+	lastOffset := count - 1
+	if lastOffset < 0 {
+		lastOffset = 0
+	}
+	path := fmt.Sprintf("%v/v%v/composes?blueprint_id=%v", RoutePrefix(), specVersion, blueprintId.String())
+	if blueprintVersion != nil {
+		path += fmt.Sprintf("&blueprint_version=%d", *blueprintVersion)
+	}
+	first := path + fmt.Sprintf("&offset=0&limit=%v", limit)
+	last := path + fmt.Sprintf("&offset=%v&limit=%v", lastOffset, limit)
```
lzap, line 580: I really think that path construction should be part of the helper function(s), this is too much to ask for a user who will be copy-pasting a new handler 8 months from now and not messing this up. Just pass a path portion into a function and it should do all the magic.

lzap (changes requested):
Almost there, tests do fail.

internal/v1/handler.go
```diff
@@ -27,6 +29,30 @@ const (
...
+	params.Add("limit", strconv.Itoa(limit))
+	fullPath.RawQuery = params.Encode()
+	first := fullPath.String()
+
+	params.Set("offset", strconv.Itoa(lastOffset))
+	fullPath.RawQuery = params.Encode()
+	last := fullPath.String()
+
+	return links{first, last}
+}
```
lzap, line 54: Can you add a test for this one? I believe it is not correct (does not carry `offset` correctly for the 2nd page).

lzap (changes requested):
Pagination concerns.

Now that think about it, pushing you to these pagination changes was a mistake as I see that others have the same code. I am opening an issue where I want to track all possible refactoring that can be done later. Feel free to keep it as is or just take a simple approach which is in align with what is currently in the existing codebase, we can change it later: https://issues.redhat.com/browse/HMS-3308

internal/v1/handler.go
ezr-ondrej: There is a test for it... I'm testing it in the handler test :)

lzap: Ok, my concern is that offset is not being added in some cases. Is that intentional?

ezr-ondrej: What do you mean? It's added to both first and last URL .... to one it's added (as it's first time its added to the params map and second time it's Set, as it's second time so we need to overriede the first value).
This is tested in the test...
I'm happy to back off with all these changes tho and just do a simple `fmt.Sprintf`

lzap: Ah yeah now I see it, disregard.

cmd/image-builder-db-test/main_test.go
```diff
@@ -380,6 +380,55 @@ func testBlueprints(t *testing.T) {
...
+
+	// get composes for specific version
+	count, err = d.CountBlueprintComposesSince(ORGID1, id, common.ToPtr(1), (time.Hour * 24 * 14), nil)
+	require.NoError(t, err)
+	require.Equal(t, 3, count)
+	entries, err = d.GetBlueprintComposes(ORGID1, id, common.ToPtr(1), (time.Hour * 24 * 14), 10, 0, nil)
+	require.NoError(t, err)
+	require.Len(t, entries, 3)
+
+	// TODO update this test to have date, after we have versioning
```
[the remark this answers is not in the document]
ezr-ondrej: I've added test for versioning :+1:

ezr-ondrej:
I've added DB test for versioning and updated the path building to use the helper method in both cases.

cmd/image-builder-db-test/main_test.go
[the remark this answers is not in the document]
lzap: Did you forget to drop the comment?

lzap (approved):
LGTM

cmd/image-builder-db-test/main_test.go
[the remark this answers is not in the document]
ezr-ondrej: Oh yeah I did... fixed :+1:

internal/v1/handler_blueprints.go
```diff
@@ -199,3 +201,68 @@ func (h *Handlers) GetBlueprints(ctx echo.Context, params GetBlueprintsParams) e
...
+	count, err := h.server.db.CountBlueprintComposesSince(idHeader.Identity.OrgID, blueprintId, params.BlueprintVersion, since, ignoreImageTypeStrings)
+	if err != nil {
+		return err
+	}
+
+	data := make([]ComposesResponseItem, 0, len(composes))
+	for _, c := range composes {
+		bId := c.BlueprintId
+		version := c.BlueprintVersion
+		clientId := (ClientId)(*c.ClientId)
```
ezr-ondrej, line 239: I've learned today. If I don't do this, the values for all entries are overriden for pointer values, apparently the memory in the cycle is reused and the struct values are overriden in the same memory.
Interesting.

lzap: I already did hit this several times myself, yep. A nasty one.

croissanne (approved):

ezr-ondrej:
I've rebased, squashed and improved the commit message :)

lzap (approved):

croissanne (approved):

ezr-ondrej:
Ah right, ephemeral is down, so I just made it less green xD I wanted to restart shutzbot, not the SRE pipeline :face_exhaling: xD

croissanne (approved):

lzap (approved):

ezr-ondrej:
/retest
They say the cluster is up :)

ezr-ondrej:
I think that the old Ephemeral cluster had some hacks for pulp deployment to work, now the new cluster does not have these hacks and fails to deploy content-sources (and that's IB dependency) , so we might be stuck with the tests for quite a while.

ezr-ondrej:
We are :green_apple: ! :tada:

croissanne (approved):

PR #944, opened 2023-12-13
HMS-2973 Update blueprints

avitova:


internal/v1/handler_blueprints_test.go
```diff
@@ -0,0 +1,46 @@
...
+	blueprintEntry, err := dbase.GetBlueprint(id, "000000", "500000")
+	require.NoError(t, err)
+
+	respStatusCode, body := tutils.GetResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/experimental/blueprint/%s", blueprintEntry.Id), &tutils.AuthString0)
+	require.Equal(t, 200, respStatusCode)
+	err = json.Unmarshal([]byte(body), &result)
+	require.NoError(t, err)
+	require.Equal(t, blueprintEntry.Name, blueprintName)
+	require.Equal(t, blueprintEntry.Description, description)
+}
```
lzap, line 46: This test does not test the UpdateBlueprint function. Is this some missing test?

lzap (commented):
Looks good, please continue with OpenAPI changes.

internal/v1/handler_blueprints.go
```diff
@@ -42,6 +43,37 @@ func BlueprintFromEntry(be *db.BlueprintEntry) (BlueprintV1, error) {
...
+		return err
+	}
+
+	var blueprintRequest CreateBlueprintRequest
+	err = ctx.Bind(&blueprintRequest)
+	if err != nil {
+		return err
+	}
+
+	if err != nil {
```
bcl, line 58: duplicate err test :)

internal/v1/handler_blueprints.go
```diff
@@ -73,6 +105,32 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
...
+	body, err := json.Marshal(blueprint)
+	if err != nil {
+		return err
+	}
+
+	id := uuid.New()
+	versionId := uuid.New()
+	err = h.server.db.UpdateBlueprint(versionId, id, body)
+	if err != nil {
+		logrus.Error("Error updating blueprint in db", err)
```
bcl, line 125: What's the difference between using logrus and ctx.Logger for logging things? And should some of these other error paths have logging?

avitova: To be honest, I do not have an idea, I just saw this in the repository few times. Is there a convention on when do you log things in certain way?

lzap: The `ctx` variable is `echo.Context` type, that is a web request context from the Echo framework. It has its own logging facade. Ultimately, it is configured to use `logrus` logging library which also can be used directly, however, there is no context handler when used this way.

Therefore, if Echo context is available one should use `ctx.Logger()` which carries over request id and org id as well as other fields which is important for correlation in the logging database. This should be preferred and the context should be passed into as many functions as possible. It also supports conversion to standard library `context.Context` type which can be passed into other libraries like `pgx` (Postgres driver) to take full advantage of other context features. However, logging in functions which do not have `ctx` as an argument must be done directly, thus, these messages will not have correlation information.

avitova:
Rebased and addressed the comments.

internal/db/db_blueprints.go
```diff
@@ -24,6 +24,18 @@ const (
...
+		SELECT 
+			$1,
+			$2,
+			MAX(version) + 1, 
+			$3
+		FROM 
+			blueprint_versions
+		WHERE 
+			blueprint_id = $2`
+
```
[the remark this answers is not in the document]
ezr-ondrej: I think the blueprint_id is not unique, but we need to generate a new UUID for the version
We're doing it in handlers in other instances and passing it to the DAO layer as parameters so I'd just go with that here as well? :)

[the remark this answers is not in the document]
avitova: Yeah, that is what I have noticed. :thinking: I have updated it to a version of how I thought we want this to look like. Feel free to comment on that.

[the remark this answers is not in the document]
lzap: Ah I forgot that this schema does not use integers but uuids for PKs, sorry. This was irrelevant.

internal/v1/api.yaml
```diff
@@ -718,6 +752,26 @@ components:
...
+        image_requests:
+          type: array
+          minItems: 1
+          items:
+            $ref: '#/components/schemas/ImageRequest'
+          uniqueItems: true
+          description: |
+            Array of image requests. Having more image requests in a single blueprint is currently not supported.
+        customizations:
+          $ref: '#/components/schemas/Customizations'
```
lzap, line 774: So you removed "name" and "description".

Isn’t simply allowing update of all the fields more straightforward for the API user? I always try to find solutions which are easier for the user, not for the developer. This would mean updating both tables in a transaction and documenting the behavior: name and description are not subject of versioning. @croissanne ?

If we do it this way, then we can actually reuse the OpenAPI component between create and update.

internal/db/db_blueprints.go
```diff
@@ -24,6 +25,18 @@ const (
...
+		INSERT INTO blueprint_versions (id, blueprint_id, version, body)
+		SELECT 
+			$1,
+			$2,
+			MAX(version) + 1, 
+			$3
+		FROM 
+			blueprint_versions
+		WHERE 
+			blueprint_id = $2`
```
lzap, line 38: I believe we should be checking for org_id her so that when an attacker crafts a valid UUID without being signed into the correct organization, the insert would do nothing:

```
test=> create table test (id serial primary key, text text);
CREATE TABLE
test=> insert into test (text) values ('test');
INSERT 0 1
test=> insert into test (text) select text from test where id = -1;
INSERT 0 0
```

lzap (changes requested):
Nice. I have one security concern and one name/description update concern.

internal/v1/api.yaml
avitova: Yeah, updating the description and the name makes totally sense to me as well. I will incorporate the idea into the PR, and we can decide later, if we want to drop this, what do you think?

internal/db/db_blueprints.go
avitova: Yeah, I believe we should return 404 on access of the user with a different org id / account number. Thanks, I had this in mind, but forgot to implement it. :+1: Definitely legit.

avitova:
I think that the tests are now failing only because of the low coverage, which we can fix later after we create more of [these handler tests](https://github.com/osbuild/image-builder/pull/951).

cmd/image-builder-db-test/main_test.go
```diff
@@ -376,6 +376,42 @@ func testBlueprints(t *testing.T) {
...
+	err = d.UpdateBlueprint(newBlueprintId, entry.Id, ORGID2, "name", "desc", newBlueprintVersionBody)
+	require.Error(t, err)
+	entryAfterInvalidUpdate, err := d.GetBlueprint(entry.Id, ORGID1, ANR1)
+	require.NoError(t, err)
+	bodyNotUpdated, err := v1.BlueprintFromEntry(entryAfterInvalidUpdate)
+	require.NoError(t, err)
+	require.Equal(t, updated, bodyNotUpdated)
+
+	require.NotEqual(t, b1, bodyNotUpdated)
+
```
lzap, line 414: Also test `version` column (not only the version stored in the JSON).

I think this test reveals that storing blueprint version in the column AND in the body column is counter-productive. I suggest to remove it from the `BlueprintV1` type completely, we need to copy the data anyway into OpenAPI specification type so we can copy from the column rather than from the body content.

internal/db/db_blueprints.go
lzap: Can you add a WHERE clause tho? Just in case someone copy-paste this SQL code, we should definitely check orgs at the database level. Join is required but not big of a deal.

You don’t have to check for this case in handler, I don’t care what an attacker see, it can be just 500. We just care about regular users :-)

internal/v1/api.yaml
lzap: Thanks, I see you added it into the code logic, can you reuse the OpenAPI type used for creation here?

lzap (changes requested):
It is great, few more small changes and this is ready. 👍🏻

cmd/image-builder-db-test/main_test.go
avitova: Should I do it in this PR, or do we do this in a followup?

internal/db/db_blueprints.go
avitova: I think I understand why you would check twice...but I just want to check if you are aware that I run the update in the transaction, where I already check the org_id for blueprint?
And in case it does not succeed, we return error from database.
So if I understand correctly, we want to do a join in case someone use this from outside the transaction?

cmd/image-builder-db-test/main_test.go
ezr-ondrej: I'd do that in a follow-up :)

ezr-ondrej: I'd say that description and name are the same case, we do have columns for those as well :)

internal/db/db_blueprints.go
ezr-ondrej: if I understand the resoning, it's to prevent copy paste errors if someone would copy your query and used it standalone outside of the transaction.
I guess we could just add it in the variable name to make it obvious smthing like `sqlOrgUnboundUpdateBlueprintVersion` ? :shrug:

ezr-ondrej (dismissed):
I'm all for it, looks pretty sane to me :+1: 
IMO extracting the fields out of the Blueprint structure makes sense, but lets do that when we work on the schema itself.

I'd love an handler test, but I guess that project failed? :)
I'm happy to add something like: https://github.com/osbuild/image-builder/pull/951 in a follow-up

internal/db/db_blueprints.go
```diff
@@ -68,12 +86,46 @@ func (db *dB) GetBlueprint(id uuid.UUID, orgID, accountNumber string) (*Blueprin
...
+	conn, err := db.Pool.Acquire(ctx)
+	if err != nil {
+		return err
+	}
+	defer conn.Release()
+
+	err = db.withTransaction(ctx, func(tx pgx.Tx) error {
+		tag, txErr := tx.Exec(ctx, sqlUpdateBlueprint, blueprintId, orgId, name, description)
+		if tag.RowsAffected() != 1 {
+			return fmt.Errorf("%w: %w, expected 1, returned %d", BlueprintNotFoundError, AffectedRowsMismatchError, tag.RowsAffected())
```
ezr-ondrej, line 110: linter does not like this and neighter do I TBH xD

```
internal/db/db_blueprints.go:110: fmt.Errorf call has more than one error-wrapping directive %w
```

I think just not using constants here would be good enough.

```suggestion
			return fmt.Errorf("blueprint not found: %w, expected 1, returned %d", AffectedRowsMismatchError, tag.RowsAffected())
```

The only use for the constant would be to programmatically distinguish wether the issue was with the first or the second query. But I don't see us needing it ATM :)

internal/db/db_blueprints.go
```diff
@@ -68,12 +86,46 @@ func (db *dB) GetBlueprint(id uuid.UUID, orgID, accountNumber string) (*Blueprin
...
+		}
+		if txErr != nil {
+			return txErr
+		}
+
+		tag, txErr = tx.Exec(ctx, sqlUpdateBlueprintVersion, id, blueprintId, body)
+		if tag.RowsAffected() != 1 {
+			return fmt.Errorf("%w: %w, expected 1, returned %d", BlueprintVersionNotFoundError, AffectedRowsMismatchError, tag.RowsAffected())
```
ezr-ondrej, line 118: ditto

ezr-ondrej (commented):
Small nits, linter has pointed that out and I'd agree :)

cmd/image-builder-db-test/main_test.go
lzap: Good call, adding it as an item here: https://issues.redhat.com/browse/HMS-3308

internal/db/db_blueprints.go
lzap: Yeah this is only possible in Go 1.21+ but this project is still on 1.19.

lzap (approved):
Thanks this is good, rebase needed tho.

ezr-ondrej (approved):
Thanks ! :orange_heart:

internal/db/db_blueprints.go
```diff
@@ -24,6 +25,29 @@ const (
...
+		FROM 
+			blueprint_versions
+		WHERE 
+			blueprint_id = $2
+        AND EXISTS (
+            SELECT 1
+            FROM blueprints
+            WHERE id = $2
+            AND org_id = $4
+        );`
```
croissanne, line 49: nitpick, indentation  should be 2 tabs here, not spaces

internal/v1/api.yaml
```diff
@@ -414,6 +414,40 @@ paths:
...
+      requestBody:
+        required: true
+        description: details of blueprint
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/CreateBlueprintRequest"
+      responses:
+        '200':
+          description: blueprint was update
```
croissanne, line 440: nitpick `blueprint was updated`

croissanne (commented):
Amazing, thank you! 2 nitpicks, otherwise lgtm!

PR #946, opened 2023-12-14
chore: update README file

avitova:
Hello, this is only a suggestion, but I got confused by the inconsistency, and this might potentially help.

Here is the reason for the change, I got confused by the other repository expecting a different approach: https://github.com/osbuild/osbuild-getting-started/pull/17

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -39,6 +39,7 @@ can be found under schutzbot.
 
 ## Unit tests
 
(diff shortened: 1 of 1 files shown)


croissanne (approved):
Generally the topic would be like `README: update unit tests instructions` or something, rather than `chore` . But for next time.

PR #949, opened 2023-12-15
api: preload swagger spec

lzap:
Function `GetSwagger` is called in many handlers which ultimately loads OpenAPI specification over and over again for no reason in every request. Since the filename is not supposed to be changing during runtime, it can be preloaded during application startup for better performance.


lzap:
Holy crap, I edited a generated file. Let me rebase this.

lzap:
Well this code is coming from openapi generator, it is not worth messing around with it. It is what it is.

PR #950, opened 2023-12-15
[wip] OpenSCAP: add kernel and services customizations

kingsleyzissou:
Update the OpenSCAP generated files to include kernel and services customizations.

Depends on #930


lavocatt (dismissed):
Lgtm! So now that means we can also send new customization for the image creation right?
Do you think there should be a new step in the wizard for the kernel and the service? Or we just pass the related customization as is?

lucasgarfield:
> Lgtm! So now that means we can also send new customization for the image creation right? Do you think there should be a new step in the wizard for the kernel and the service? Or we just pass the related customization as is?

For now this information will be read-only and displayed on the OpenSCAP step. In the future we will allow users to set these values on their own (even when not building a security compliant image) but that's not in our current work's scope.

croissanne (approved):

croissanne (approved):

PR #940, opened 2023-12-12
test: Add DB test for inserting Blueprint compose

ezr-ondrej:
Adding test for https://github.com/osbuild/image-builder/pull/921

PR #951, opened 2023-12-15
test: Add handler test for Compose Blueprint

ezr-ondrej:
Add handler test for testing the Compose Blueprint endpoint. It tests whether the endpoint contacts the composer correct amount of time and retrives ID for each compose.

Follow-up for https://github.com/osbuild/image-builder/pull/940

### DIFF
--- a/internal/tutils/idheader_utils.go
+++ b/internal/tutils/idheader_utils.go
@@ -41,7 +41,8 @@ var completeIdHeader string = `{
 		 	"is_internal": true,
 		 	"locale": "en-US"
 		},
-		"internal": {"org_id": "%s"}
+		"internal": {"org_id": "%s"},
+		"account_number": "000000"
 	}
 }`
 
@@ -58,7 +59,8 @@ var idHeaderWithoutEntitlements = `{
 			"is_internal": true,
 			"locale": "en-US"
 		},
-		"internal": {"org_id": "%s"}
+		"internal": {"org_id": "%s"},
+		"account_number": "000000"
 	}
 }`
 
--- a/internal/v1/handler_blueprints_test.go
+++ b/internal/v1/handler_blueprints_test.go
@@ -1,0 +1,99 @@
+package v1
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/osbuild/image-builder/internal/common"
+	"github.com/osbuild/image-builder/internal/composer"
+	"github.com/stretchr/testify/require"
+
+	"github.com/osbuild/image-builder/internal/tutils"
+)
+
+func TestHandlers_ComposeBlueprint(t *testing.T) {
+	ids := []uuid.UUID{uuid.New(), uuid.New()}
+	idx := 0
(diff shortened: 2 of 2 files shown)


internal/v1/handler_blueprints_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	var message []byte
+	message, err = json.Marshal(blueprintBody)
+	err = dbase.InsertBlueprint(id, versionId, "000000", "000000", "blueprint", "desc", message)
+	require.NoError(t, err)
+
+	respStatusCode, body := tutils.PostResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/experimental/blueprint/%s/compose", id.String()), map[string]string{})
+	require.Equal(t, http.StatusCreated, respStatusCode)
+
+	var result []ComposeResponse
```
ezr-ondrej, line 92: I was bit surprised that we respond with array of objects each containing ID, feels bit odd and I missed that on review.
Do we want to think about the response format bit more @amirfefer and @lzap, or is it what we want?
Also happy for your opinion @avitova :)

Keeping in draft until we decide that, if we want to change the response format, this PR might be perfect opportunity :)

internal/tutils/idheader_utils.go
```diff
@@ -41,7 +41,8 @@ var completeIdHeader string = `{
 		 	"is_internal": true,
 		 	"locale": "en-US"
 		},
-		"internal": {"org_id": "%s"}
+		"internal": {"org_id": "%s"},
+		"account_number": "000000"
```
ezr-ondrej, line 45: We might want to just remove account_number usage completely and keep it null for all records, IB seems to have dropped the usage of it.

internal/v1/handler_blueprints_test.go
avitova: Oh yeah. You might be correct. We might want to use ComposesResponse, right?

internal/tutils/idheader_utils.go
avitova: What do you mean they dropped it? I still can see it somewhere in IB codebase. :thinking:

lzap: Looks like account is still getting stored, however, it is unused for queries. Looks like some transition phase, to be on the safe side I suggest to still use it. It can be deleted in the final removal patch, I guess that is planned at some point.

internal/v1/handler_blueprints_test.go
lzap: Well the other option what I considered was to return nothing. This is a blueprint REST endpoint after all.

Are you asking about what to return, or how we return it? Returning slice is definitely not a good practice, we should return a type with some sort of root element just in case we need pagination later. I suggest to add it into our refactoring issue perhaps: https://issues.redhat.com/browse/HMS-3308

lzap (commented):
Let's add a test and improve reply later.

internal/v1/handler_blueprints_test.go
ezr-ondrej: Yup, added it there, lets not add to the scope of the PR :+1:

lzap (approved):

croissanne (approved):

PR #952, opened 2023-12-17
api: add meta type to spec

amirfefer:
Adding blueprints increase the endpoints with pagination. This PR add `Meta` type to spec.


internal/v1/api.yaml
```diff
@@ -475,7 +475,7 @@ components:
         - readiness
       properties:
         readiness:
-          type: string
+          type: string
```
avitova, line 478: Is this meant to be there?

lzap (changes requested):
Yeah please remove the trailing whitespace and the change can go.

lzap:
I suggest to hold on this refactoring until all blueprint PRs are merged, the metadata deserves a full type for all the fields:

* count
* limit
* offset

The change will be larger then and to avoid rebasing issues I think we might actually do it after everything is done.

PR #959, opened 2023-12-19
.github/workflows: run workflows on merge_group events

croissanne:
This will enable a merge queue!


ondrejbudai (approved):
Yup!

ochosi:
@croissanne I think it'd be nice if the PR description would contain some hints as to what we're enabling here (merge queues).
Also, the PR seems to be approved and green and mergeable, but I don't really dare to push the button for the lack of description/clarity :)

PR #961, opened 2023-12-20
api: add search param to blueprints endpoint

amirfefer:
This PR extends the `/blueprints` endpoint by adding a search query for searching blueprints by name or descriptions.


lzap (changes requested):
It is okay and it will work.

But I would really like to avoid table scans for all listing operations by introducing a special function for this. That would be much on the more safe side.

amirfefer:
@lzap, thanks! I thought that Postgres has optimization on searching, but I agree, better safe than sorry, also it improves readability.

internal/db/db_blueprints.go
```diff
@@ -58,6 +58,19 @@ const (
 		ORDER BY last_modified_at DESC
 		LIMIT $3 OFFSET $4`
 
+	sqlFindBlueprints = `
+		SELECT blueprints.id, blueprints.name, blueprints.description, MAX(blueprint_versions.version) as version, MAX(blueprint_versions.created_at) as last_modified_at
+		FROM blueprints INNER JOIN blueprint_versions ON blueprint_versions.blueprint_id = blueprints.id
+		WHERE blueprints.org_id = $1 AND blueprints.account_number = $2 AND (blueprints.name ILIKE $5 OR blueprints.description ILIKE $5)
```
croissanne, line 64: I'd just filter on org_id, account_number is just there for show

internal/db/db_blueprints.go
```diff
@@ -58,6 +58,19 @@ const (
...
+		GROUP BY blueprints.id
+		ORDER BY last_modified_at DESC
+		LIMIT $3 OFFSET $4`
+
+	sqlCountFilteredBlueprints = `
+		SELECT COUNT(*)
+		FROM blueprints
+		WHERE blueprints.org_id = $1 AND blueprints.account_number = $2 AND (blueprints.name ILIKE $3 OR blueprints.description ILIKE $3)`
```
croissanne, line 72: ditto

internal/db/db_blueprints.go
```diff
@@ -53,15 +53,28 @@ const (
...
+		WHERE blueprints.org_id = $1
 		GROUP BY blueprints.id
 		ORDER BY last_modified_at DESC
-		LIMIT $3 OFFSET $4`
+		LIMIT $2 OFFSET $3`
+
+	sqlFindBlueprints = `
+		SELECT blueprints.id, blueprints.name, blueprints.description, MAX(blueprint_versions.version) as version, MAX(blueprint_versions.created_at) as last_modified_at
+		FROM blueprints INNER JOIN blueprint_versions ON blueprint_versions.blueprint_id = blueprints.id
+		WHERE blueprints.org_id = $1 AND (blueprints.name ILIKE $4 OR blueprints.description ILIKE $4)
```
avitova, line 64:
```suggestion
		WHERE blueprints.org_id = $1 AND (blueprints.name ILIKE $4 AND blueprints.description ILIKE $4)
```

Not sure but this would be more helpful, I think. If I inserted too general name, then the description would not really filter out the result.

internal/db/db_blueprints.go
```diff
@@ -53,15 +53,28 @@ const (
...
+		GROUP BY blueprints.id
+		ORDER BY last_modified_at DESC
+		LIMIT $2 OFFSET $3`
+
+	sqlCountFilteredBlueprints = `
+		SELECT COUNT(*)
+		FROM blueprints
+		WHERE blueprints.org_id = $1 AND (blueprints.name ILIKE $2 OR blueprints.description ILIKE $2)`
```
avitova, line 72: ditto

internal/db/db_blueprints.go
amirfefer: But we want to filter blueprints by name or by description. no?

avitova: Oh, I see now. Yeah, we have the same argument $4, then this makes sense. :)

croissanne (approved):
thank you! lgtm

croissanne (approved):

croissanne (approved):

internal/v1/handler_blueprints.go
```diff
@@ -161,10 +161,19 @@ func (h *Handlers) GetBlueprints(ctx echo.Context, params GetBlueprintsParams) e
 	if params.Offset != nil {
 		offset = *params.Offset
 	}
+	var blueprints []db.BlueprintWithNoBody
+	var count int
 
-	blueprints, count, err := h.server.db.GetBlueprints(idHeader.Identity.OrgID, idHeader.Identity.AccountNumber, limit, offset)
-	if err != nil {
-		return err
+	if params.Search != nil {
```
lzap, line 167: I would probably also add a case when `params.Search != ""` because in that case we do a weird query `ILIKE %%%%` and it can easily happen that our JS code will be sending an empty string instead of `nil` every time, thus every single time when the app would request all blueprints we would actually perform an ILIKE search.

internal/db/db_blueprints.go
```diff
@@ -159,15 +172,47 @@ func (db *dB) DeleteBlueprint(id uuid.UUID, orgID, accountNumber string) error {
...
 
-func (db *dB) GetBlueprints(orgID, accountNumber string, limit, offset int) ([]BlueprintWithNoBody, int, error) {
+func (db *dB) FindBlueprints(orgID, search string, limit, offset int) ([]BlueprintWithNoBody, int, error) {
+	ctx := context.Background()
+	conn, err := db.Pool.Acquire(ctx)
+	if err != nil {
+		return nil, 0, err
+	}
+	defer conn.Release()
+	searchQuery := fmt.Sprintf("%s%s%s", "%", search, "%")
```
lzap, line 182: I still think this is too messy for no reason, how about:

```suggestion
	searchQuery := "%" + search + "%"
```

lzap (approved):
I am good, few small nit-picks.

One thing to note is that we do not perform any escaping, so the moment user enters `%` character they will get a SQL parsing error. Since the LIKE syntax is very simple, we could simply escape percent chars with `%%` that that should do it.

croissanne (approved):

PR #962, opened 2023-12-20
Update blueprints followup

avitova:
Followup of the nitpics from the MR #944.

I also added some simple README update:

Is generating of the api mentioned somewhere? I could not find it at first, and struggled a bit. I think this might come in handy to be somewhere with keywords "api.go" and "api.yaml" for easier search when you need it.
This could as well go into the makefile or some other docs. This is just a suggestion where I thought this might fit in, but feel free to recommend a different place. You also have the [developer part](https://www.osbuild.org/guides/developer-guide/) of your documentation, but I did not find enough info there either.
Feel free to add your opinions.


HACKING.md
```diff
@@ -32,3 +32,7 @@ file to generate a package list.
 
 If the repository requires a client tls key/cert you can supply them with
 `--key` and `--cert`.
+
+## Generating Open API
+
+For updating the api.go content to match the current api.yaml, simply run `go generate ./...`
```
bcl, line 38: There's a script, `./tools/prepare-source.sh` which runs that among other things. But it isn't mentioned in this repo :) There are similar scripts in osbuild-composer, images, and weldr-client which is probably why nobody mentioned it here.

bcl:
I would also change the api commit message to be a bit more descriptive. Something like 'clean up blueprint update formatting  and description' with the goal being to be able to read the short commit messages and have some idea of whether or not you need to look closer at the commit while debugging something in 6 months :)

lzap (dismissed):
Please follow [the commit guide](https://www.osbuild.org/guides/developer-guide/general/workflow.html):

[Preferred commit message format](https://www.osbuild.org/guides/developer-guide/general/workflow.html#preferred-commit-message-format)
<module>: Topic of the commit

Body of the commit, describing the changes in more detail.
The <module> should point to the area of the codebase (for instance tests or tools). The topic should summarize what the commit is doing.
GitHub truncates the first line if it's longer than 65 characters, which is something to keep in mind as well.

_It is a bit weird, typically a patch touches more than one packages then I am always not sure which prefix to pick._

avitova:
@bcl Thanks for clarifying! In that case, let's just address only the nitpicks Sanne had:)

lzap (approved):

avitova:
Rebased

lzap (dismissed):

ondrejbudai (dismissed):
Thanks!

lzap:
Could you rebase once again? Thanks.

avitova:
Rebased, but the PR build is still failing.

ezr-ondrej:
> Rebased, but the PR build is still failing.

Interesting, the CI was fixed and went green on few PRs already, it just doesn't seem to like yours for some reason :/
The reason is most likely unrelated to the code tho, so it's pretty weird.

avitova:
Rebased, and the tests apparently passed. :partying_face:

ezr-ondrej (approved):
Nice! this is minor thing and approved many times, lets get that in.
Thank you @avitova :+1:

PR #973, opened 2024-01-03
api: add missing ids and tags

lzap:
I noticed that some of the endpoints are missing ids. To further improve generated output from HTML/JS generator, this patch adds tags which improves the redocly generator.

This is a portion of https://github.com/osbuild/image-builder/pull/900


bcl:
I think you need to rerun the generator. Also, #900 links to #839 which has the original problem description :) Could you make the this commit message a little more verbose?

lzap:
Thanks, improved the PR description.

lzap:
Is there any chance to get this in? I stripped this down to the barebones. All I want is the link in the README which I find quite useful and better navigation which the tags do bring. Thanks.

ezr-ondrej (approved):
This looks pretty reasonable to me and no real change, so it should  not break stuff :)

PR #977, opened 2024-01-05
image-builder: log a warning msg when glitchtip not initialized

diaasami:
To help debug lack of issues showing up on glitchtip
https://issues.redhat.com/browse/HMS-1986

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -53,6 +53,10 @@ func main() {
 		panic(err)
 	}
 
+	if conf.GlitchTipDSN == "" {
+		logrus.Warn("Sentry/Glitchtip was not initialized")
+	}
+
 	if conf.CwAccessKeyID != "" {
 		err = logger.AddCloudWatchHook(logrus.StandardLogger(), conf.CwAccessKeyID, conf.CwSecretAccessKey, conf.CwRegion, conf.LogGroup)
 		if err != nil {


cmd/image-builder/main.go
```diff
@@ -46,6 +46,9 @@ func main() {
 		if err != nil {
 			panic(err)
 		}
+
+	} else {
+		log.Warn("Sentry was not initialized")
```
ochosi, line 51: How about adding a reference to the GlitchTip name as well?

ochosi (commented):
The PR description is empty, but the change makes sense to me. Wondering if it would make sense to not only refer to Sentry but also GlitchTip, because someone less familiar with the subject may be looking for that in the logs.

croissanne (changes requested):

cmd/image-builder/main.go
```diff
@@ -46,13 +45,18 @@ func main() {
 		if err != nil {
 			panic(err)
 		}
+
```
croissanne, line 48: Empty newline

cmd/image-builder/main.go
```diff
@@ -17,7 +17,6 @@ import (
 	sentryecho "github.com/getsentry/sentry-go/echo"
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
-	"github.com/labstack/gommon/log"
```
croissanne, line 20: We still need log to set the level.

croissanne (approved):

PR #978, opened 2024-01-06
Test: Add test for non existing distribution compose

elkoniu:
When non existing distribution compose is requested image builder should intercept the query and return the error to user.

### DIFF
--- a/internal/v1/handler_post_compose_test.go
+++ b/internal/v1/handler_post_compose_test.go
@@ -573,6 +573,38 @@ func TestComposeImageErrorsWhenCannotParseResponse(t *testing.T) {
 	require.Contains(t, body, "Internal Server Error")
 }
 
+// This test case queries the image-builder for a non existing type of the os distribution
+// osbuild-composer is not being mock here as the error should be intercepted by image-builder
+func TestComposeImageErrorsWhenDistributionNotExists(t *testing.T) {
+	srv, tokenSrv := startServer(t, "", "", nil)
+	defer func() {
+		err := srv.Shutdown(context.Background())
+		require.NoError(t, err)
+	}()
+	defer tokenSrv.Close()
+
+	var uo UploadRequest_Options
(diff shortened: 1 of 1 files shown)


elkoniu:
@croissanne I think the behavior mentioned in related issue has been already implemented. I just added unit test to cover it.

internal/v1/handler_post_compose_test.go
```diff
@@ -566,10 +566,60 @@ func TestComposeImageErrorsWhenCannotParseResponse(t *testing.T) {
...
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusCreated)
+		result := composer.ComposeId{
+			Id: id,
+		}
+		err := json.NewEncoder(w).Encode(result)
+		require.NoError(t, err)
+	}))
+	defer apiSrv.Close()
+
```
croissanne, line 593: I think this can be removed, image builder doesn't contact composer in this scenario.

internal/v1/handler_post_compose_test.go
```diff
@@ -566,10 +566,60 @@ func TestComposeImageErrorsWhenCannotParseResponse(t *testing.T) {
...
+	srv, tokenSrv := startServer(t, apiSrv.URL, "")
```
croissanne, line 594:
```suggestion
	srv, tokenSrv := startServer(t, "", "")
```

croissanne (changes requested):
Thanks! It seems it's already returning 400, so somewhere along the line the issue got fixed. Test is a good idea though.

internal/v1/handler_post_compose_test.go
elkoniu: Updated

internal/v1/handler_post_compose_test.go
elkoniu: I have checked and you are right. Code updated:)

croissanne (dismissed):

croissanne (approved):

PR #981, opened 2024-01-10
db: optimize blueprint search

lzap:
This optimizes query in case no search is performed. In that case empty string is passed into the db function and the argument is `%%`. The change will help PostgreSQL optimizer to drop the rest of the where clause in order to quickly return results without any tablescan.

Refs: HMS-3308

### DIFF
--- a/cmd/image-builder-db-test/main_test.go
+++ b/cmd/image-builder-db-test/main_test.go
@@ -435,7 +435,10 @@ func testBlueprints(t *testing.T) {
 	require.Equal(t, entries[1].Version, updated.Version)
 
 	err = d.InsertBlueprint(uuid.New(), uuid.New(), ORGID1, ANR1, "unique name", "unique desc", body)
-	entries, count, err := d.FindBlueprints(ORGID1, "unique", 100, 0)
+	entries, count, err := d.FindBlueprints(ORGID1, "", 100, 0)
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
Looks good to me :+1:
Do we have this tested?

cmd/image-builder-db-test/main_test.go
```diff
@@ -435,7 +435,10 @@ func testBlueprints(t *testing.T) {
 	require.Equal(t, entries[1].Version, updated.Version)
 
 	err = d.InsertBlueprint(uuid.New(), uuid.New(), ORGID1, ANR1, "unique name", "unique desc", body)
-	entries, count, err := d.FindBlueprints(ORGID1, "unique", 100, 0)
+	entries, count, err := d.FindBlueprints(ORGID1, "", 100, 0)
+	require.NoError(t, err)
+	require.Equal(t, 1, count)
```
ezr-ondrej, line 440: We've got 2 created prior this test already, I'd just bump it to 3 here, but we might want to think about splitting this test up in some refactoring PR IMO :)

lzap:
Fixed the test.

ezr-ondrej (approved):

croissanne (approved):

PR #982, opened 2024-01-10
internal/common: use regexp to compare allowed distros & wildcard orgid

croissanne:
internal/common: add global org id allowlist

For the community service we want to allow fedora, a currently
"restricted" distro, and forbid rhel, which isn't restricted. In order
to just consider all distros as restricted a global filter is needed.


bcl (approved):
Looks good to me

PR #983, opened 2024-01-10
image-builder: improve sentry integration

diaasami:
Add a logrus hook to send Error logs to sentry as well
https://issues.redhat.com/browse/HMS-1986

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/getsentry/sentry-go"
 	sentryecho "github.com/getsentry/sentry-go/echo"
+	sentrylogrus "github.com/getsentry/sentry-go/logrus"
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
 	"github.com/labstack/gommon/log"
@@ -55,6 +56,11 @@ func main() {
 
 	if conf.GlitchTipDSN == "" {
 		logrus.Warn("Sentry/Glitchtip was not initialized")
+	} else {
+		sentryhook := sentrylogrus.NewFromClient([]logrus.Level{logrus.PanicLevel,
+			logrus.FatalLevel, logrus.ErrorLevel},
+			sentry.CurrentHub().Client())
+		logrus.AddHook(sentryhook)
 	}
 
 	if conf.CwAccessKeyID != "" {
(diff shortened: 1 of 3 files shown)


cmd/image-builder/main.go
```diff
@@ -46,6 +47,7 @@ func main() {
 		if err != nil {
 			panic(err)
 		}
+		defer sentry.Flush(5 * time.Second)
```
croissanne, line 51: I'm not sure I understand the purpose of this, if it panics will it still be executed even with the 5 seconds timer?

cmd/image-builder/main.go
```diff
@@ -57,6 +58,12 @@ func main() {
 
 	if conf.GlitchTipDSN == "" {
 		logrus.Warn("Sentry/Glitchtip was not initialized")
+	} else {
+		sentryhook := sentrylogrus.NewFromClient([]logrus.Level{logrus.PanicLevel,
+			logrus.FatalLevel, logrus.ErrorLevel},
+			sentry.CurrentHub().Client())
+		logrus.AddHook(sentryhook)
+		defer sentryhook.Flush(2 * time.Second)
```
croissanne, line 66: Why is the flush needed here?

cmd/image-builder/main.go
```diff
@@ -60,7 +60,7 @@ func main() {
 		logrus.Warn("Sentry/Glitchtip was not initialized")
 	} else {
 		sentryhook := sentrylogrus.NewFromClient([]logrus.Level{logrus.PanicLevel,
-			logrus.FatalLevel, logrus.ErrorLevel},
+			logrus.FatalLevel, logrus.ErrorLevel, logrus.InfoLevel},
```
croissanne, line 63: This seems very overkill, the point of glitchtip is to only get the few errors that are so rare they might not influence the SLOs.

cmd/image-builder/main.go
diaasami: it takes a maximum 5 seconds to flush existing channels before main ends, but might return before the 5 seconds.

cmd/image-builder/main.go
diaasami: the `InfoLevel` part will be reverted later, that's why it's a separate commit, only meant for testing to help us make sure everything is working.

cmd/image-builder/main.go
diaasami: to process logrushook channels (going into sentry's channels), I believe.

croissanne (approved):

PR #995, opened 2024-01-17
templates: Include clientid in FloorPlan (HMS-3244)

ochosi:
The ClientID is being tracked in our database (https://github.com/osbuild/image-builder/pull/897) but as it's not included in the floorplan it doesn't get exported.


ochosi:
> The code change looks good to me. Didn't test it though, not sure how 😅

You cannot really test it apart from looking at the db dumps after the fact ;)

PR #996, opened 2024-01-17
v1: add support for fedora identities

croissanne:


ezr-ondrej (approved):
Dunno how to test this, but looks sane and the fallback to RHID seems done well, so at minimum it should not break stage :)

I guess we could remove account IDs soon to simplify this :)

croissanne:
> Dunno how to test this, but looks sane and the fallback to RHID seems done well, so at minimum it should not break stage :)
> 
> I guess we could remove account IDs soon to simplify this :)

Eh i tihnk it's fine to have as many fallbacks as we need. maybe at some point we have a little library which can handle multiple identities, idk.

internal/v1/server_identity.go
```diff
@@ -0,0 +1,83 @@
...
+			// the user's org does not have an associated EBS account number, these
+			// are associated when a billing relationship exists, which is a decent
+			// proxy for RHEL entitlements
+			logrus.Error("RHEL entitlement not present in identity header")
+			return i.AccountNumber() != ""
+		}
+		return entitled.IsEntitled
+	}
+	return false
+}
```
lzap, line 83: It looks like `v1` package is doomed to be the jack of all trades 😓

PR #998, opened 2024-01-18
v1: rename to blueprints

lzap:
Although there is a precedent in using both plural and singular forms (composes/compose), I believe it is worth to be consistent.

The API is still experimental, so this patch is fine.

### DIFF
--- a/internal/v1/api.go
+++ b/internal/v1/api.go
@@ -893,6 +893,18 @@ type GetComposeClonesParams struct {
 	Offset *int `form:"offset,omitempty" json:"offset,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


ezr-ondrej (approved):
Yup, from what @croissanne said I understand this is what we want, so lets get it in sooner rather then later :)

ezr-ondrej (changes requested):
Tests needs updating (we use the path in tests)

lzap:
Damn I missed these strings somehow.

ezr-ondrej (approved):
Only renames experimental enpoints, looks safe, thanks @lzap ! :+1:

ezr-ondrej:
:green_apple: :tada:

PR #1001, opened 2024-01-18
v1/api: Add proper OpenAPI components for Meta and Links

ezr-ondrej:
Creates OpenAPI components for Meta and Links to standartize List responses.

Refs HMS-3308


lzap (approved):
Much better. 

It would be lovely to see a helper function for the ugly `Sprintf` as the URL parameters will always be same. But this is fine.

ezr-ondrej:
I agree with the helper, but I'll try that out as follow-up, my first try ended up bit ugly, so I'd love to have a separate discussion around it.

lzap:
Ok tests do fail tho.

PR #1000, opened 2024-01-18
v1/api: add DeleteBlueprint endpoint

ezr-ondrej:
Adds Delete endpoint for Blueprint.

Refs HMS-3417

PR #1002, opened 2024-01-18
api: Add Get Blueprint API endpoint

avitova:
Let's merge #998 and #1000 first!:)


internal/v1/api.yaml
```diff
@@ -606,6 +606,35 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPErrorList'
+  /experimental/blueprints/{id}/detail:
```
lzap, line 609: According to REST best practices, this can be as simple as:

```suggestion
  /experimental/blueprints/{id}:
```

If there is precedens, let’s keep that.

lzap (dismissed):
LGTM

lzap (approved):
Nice, you noticed the updated identity handling. Great work.

PR #1003, opened 2024-01-19
templates/image-builder: fix fedora auth template variable

croissanne:
Variables require brackets.


bcl (approved):
Hmm, I'm not sure about that... Seems too simple ;)

croissanne:
> Hmm, I'm not sure about that... Seems too simple ;)

it's wishful coding!

lzap:
Sorry about that, my brain simply switches to shell syntax mode when I see something like this.

PR #1012, opened 2024-01-25
v1: do not inline swagger blob

lzap:
The openapi3 generator this project use can generate inline compressed/base64 encoded blob of the original spec file, however, this makes way too many git conflicts as the compression is not git friendly. This replaces the generated code with more straightforward Go runtime file embedding. @ezr-ondrej


ezr-ondrej (approved):
Great, thanks! :+1: 
I'll hold off merging for now to not add to IB teams plate in case something goes wrong, but I love it :orange_heart:

lzap:
Tests do fail because I believe: https://github.com/osbuild/image-builder/pull/1014

ezr-ondrej:
@croissanne FYI I'm merging this, I think this this is exactly what we were targeting, but if we've missed any usecase, we can revert :)

Thanks @lzap :orange_heart:

teg:
My memory might not serve me correctly, but I believe the reason we did static embedding was to improve the latency (which was very high and frequently used for some reason). I'm not sure how this patch compares to the original implementation, but might make sense to have a look at how (if) this impacted the latency.

ezr-ondrej:
Thanks for the background info, this reads it to memory at load time, just from the initial file, so it should be fine IMO.

teg:
Ah, cool. Ignore me :)

lzap:
Yeah, actually I believe that this (new) implementation is faster as there is no base64 decoding, it is just memory blob what is being served. But feelings and performance test results are two different worlds.

PR #1025, opened 2024-02-07
v1/api: composes filtering by latest blueprint version

ezr-ondrej:
Add special value parameter -1 for fetching latest blueprint version. Adds extra query for fetching the version, to keep the SQL simple and easy to read.

Refs: HMS-3412

### DIFF
--- a/internal/db/db_blueprints.go
+++ b/internal/db/db_blueprints.go
@@ -26,6 +26,11 @@ const (
 		INSERT INTO blueprint_versions(id, blueprint_id, version, body)
 		VALUES($1, $2, $3, $4)`
 
+	sqlLatestBlueprintVersion = `
+		SELECT MAX(blueprint_versions.version)
+		FROM blueprint_versions INNER JOIN blueprints ON blueprint_versions.blueprint_id = blueprints.id
+		WHERE blueprints.org_id = $1 AND blueprints.id = $2`
+
 	sqlGetBlueprintComposes = `
 		SELECT blueprint_versions.version, composes.job_id, composes.request, composes.created_at, composes.image_name, composes.client_id
 		FROM composes INNER JOIN blueprint_versions ON composes.blueprint_version_id = blueprint_versions.id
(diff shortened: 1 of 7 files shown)


internal/v1/handler_blueprints.go
```diff
@@ -254,6 +254,15 @@ func (h *Handlers) GetBlueprintComposes(ctx echo.Context, blueprintId openapi_ty
 
 	since := time.Hour * 24 * 14
 
+	if params.BlueprintVersion != nil {
+		if *params.BlueprintVersion < 0 {
```
lzap, line 258: Can you use `and` here?

lzap (changes requested):
Small nitpick, since we are only touching blueprints experimental code, let’s merge even tho composer tests will fail.

ezr-ondrej:
:green_apple:

amirfefer (commented):
Thanks @ezr-ondrej 👍 

I think this addition is great for the API, but it's not needed for the UI,
The composes are returned sorted by version, so we can hide older versions (client side filtering). When pagination is needed we can use the latest version (first compose's version) and trigger a request with that version (instead of -1)

internal/db/db_blueprints.go
```diff
@@ -26,6 +26,11 @@ const (
 		INSERT INTO blueprint_versions(id, blueprint_id, version, body)
 		VALUES($1, $2, $3, $4)`
 
+	sqlLatestBlueprintVersion = `
+		SELECT MAX(blueprint_versions.version)
+		FROM blueprint_versions INNER JOIN blueprints ON blueprint_versions.blueprint_id = blueprints.id
+		WHERE blueprints.org_id = $1 AND blueprints.id = $2`
+
```
lzap, line 33: Have you tested with multiple versions? I believe this query will return multiple records if you have multiple versions which `QueryRow` will silently ignore, yet, all those records will be transferred from DB to the backend for no reason. I suggest to add `LIMIT 1` here.

lzap (changes requested):
One nitpick.

internal/db/db_blueprints.go
ezr-ondrej: I test it with multiple versions, but might be just accident it works in the tests.

IMO `MAX()` without grouping selects only single row tho, or am I missing something?

lzap: Yeah that is right there is no group so it will return single record. What I was thinking, okay.

lzap (approved):

PR #1040, opened 2024-02-20
v1/api: add base64-encoded file data

lzap:
This is needed to workaround Akamai content filter which goes nuts the moment it sees `/usr/bin/`. All requests containing this in body are refused with 403.

https://issues.redhat.com/browse/HMS-3581

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -1582,6 +1582,11 @@ components:
         data:
           type: string
           description: Contents of the file as plain text
+        data_encoding:
+          type: string
+          enum: ['plain', 'base64']
+          description: When data is base64-encoded to prevent Akamai content filter false positives
+          default: 'plain'
         ensure_parents:
           type: boolean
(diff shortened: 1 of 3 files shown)


internal/v1/api.yaml
```diff
@@ -1619,6 +1619,10 @@ components:
         data:
           type: string
           description: Contents of the file as plain text
+        data_base64:
```
ondrejbudai, line 1622: What about `data_encoding`? Make it an enum with `plain` and `base64`, `plain` is the default value. This allows us to add more encodings in the future if we ever have such need.

lzap: Good idea, enum it is.

lzap:
Notification service config ephemeral error, trying again.

ezr-ondrej (approved):
Thanks @lzap I'm gonna follow-up with a test :)

lzap:
Thanks, I could not find any for customization, so I skipped on it...

Issue #1042, opened 2024-02-20
Developer guide link in the docs is 404


AdamWill:
The image-builder `HACKING.md` file says "Please refer to the [developer guide](https://www.osbuild.org/guides/developer-guide/index.html) to learn about our workflow, code style and more." The link there is 404. I've no idea where to find the developer guide.

If you go to the osbuild developer guide at https://osbuild.org/docs/developer-guide/index , click on Image Builder and then Image Builder contributing guide, it actually gives you the same contents of that `HACKING.md` file with the same 404 link, so yeah, really don't know where to look.

bcl:
https://github.com/osbuild/osbuild.github.io/pull/43 will fix the guides, thanks!

schuellerf:
I think this works now, please re-open if you think otherwise

PR #1043, opened 2024-02-20
distributions: add Fedora 41

AdamWill:
Fedora 40 branched around 2024-02-14, so Rawhide now "is" (to some extent) Fedora 41.

I suspect there's a scripted way to do this, but as [the developer guide is 404](https://github.com/osbuild/image-builder/issues/1042) and it wasn't anything obvious, I couldn't find it. I based this on the [F40 commit](https://github.com/osbuild/image-builder/commit/371994adee062ea6d804eeee2fb323930358d06e) with appropriate tweaks (fortunately that swagger blob is gone).

F40 branched around 2024-02-14, and 41 has "existed" to some degree (notably, I think Rawhide osbuild images should have been using this definition) since then. Updating this needs to be on someone's SOP, I guess, if I'm right. CC @nirik


AdamWill:
Rebased to drop the modular repos. Fedora has had no modular content since F39, there is no reason to include these repos (and we're debating not having the metalinks exist).

ezr-ondrej:
@croissanne could you take a look? I do not fully understand distros yet

croissanne (approved):

croissanne (approved):
thanks!

ezr-ondrej:
Thank you for your contribution @AdamWill ! :orange_heart:
And sorry it took so long to get it in, the CI was bit moody lately :/

PR #1044, opened 2024-02-20
Drop modular repos from Fedora 39 and 40 configurations

AdamWill:
Fedora has no modular content since Fedora 39. The metalinks for F39 just point to the Everything repositories, and the metalinks for F40 just don't exist at all (so if anything expects these URLs to actually work, it'll break).


ondrejbudai (approved):
Thanks!

PR #1052, opened 2024-02-26
db: drop account number check from blueprints

ezr-ondrej:
Account number will be dropped soon, we do not want to enforce it.

This is blocking stage testing currently, as there is no account number in stage.


ezr-ondrej:
Failure on pr_check is unrelated (it can't deploy `create-contentsources-user` whatever it is), but I'd like to see it pass on ephemeral :/

lzap (approved):
Sure so it will be an empty string for these cases, this is okay.

PR #1054, opened 2024-02-27
pr_check: remove content-sources from deployment

jrusz:
We don't test content-sources integration in ephemeral.


ezr-ondrej (approved):
Thanks @jrusz and @swadeley ! :orange_heart:

PR #1056, opened 2024-02-28
v1: return commit sha via version endpoint

lzap:
I was not sure recently if my change was pushed to stage or not. Since the version endpoint returns just 1.0 that does not help at all, this patch adds sha commit and build datetime:

```
$HOME/bin/curl-local-ib GET /image-builder/v1/version
{
  "build_commit": "898e84",
  "build_time": "2024-02-28T07:25:03Z",
  "version": "1.0"
}
```

It also prints the same in logs while starting the app.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -181,7 +181,7 @@ func main() {
 		panic(err)
 	}
 
-	logrus.Infof("🚀 Starting image-builder server on %v ...\n", conf.ListenAddress)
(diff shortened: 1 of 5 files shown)


ezr-ondrej:
/retest

Timeout :face_exhaling:

lzap:
All green, thanks.

ezr-ondrej (approved):
Nice addition, thanks ! :tada:

PR #1057, opened 2024-02-28
v1: pass echo context through handlers

lzap:
This is a followup for my logging improvement patch. This ensures the echo context is passed around. Generally speaking, instead of:

```
logrus.Info(...)
```

write:

```
ctx.Logger().Info(...)
```

That is all that needs to be done. In the next patch, I will update the db functions to accept context.

### DIFF
--- a/internal/v1/handler.go
+++ b/internal/v1/handler.go
@@ -19,7 +19,6 @@ import (
 	"github.com/osbuild/image-builder/internal/provisioning"
 
 	"github.com/labstack/echo/v4"
-	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


ezr-ondrej (approved):
I do love it, tho this works just for handlers I'm guessing, for other funcs we want to pass the inner go context, right? I just want to confirm for future reference :)

lzap:
> I do love it, tho this works just for handlers I'm guessing, for other funcs we want to pass the inner go context, right? I just want to confirm for future reference :)

Yes the standard context is the de-facto standard these days for other libraries. This will have a follow-up, I want to have smaller PRs.

lzap:
Not sure what exactly is failing in this case.

PR #1058, opened 2024-02-28
Make openapi spec endpoint public (HMS-1220)

schuellerf:
At least two flaws.

The testcase fails! But that should be fixed (in the code, not the test) I guess. Result is:
```
--- Expected
+++ Actual
@@ -1095,2 +1095,3 @@
 "servers":[{"url":"/api/image-builder/v1"},
+{"url":"/api/image-builder/v1.0"},
 {"url":"/api/image-builder/v1.0"}]}
```
(EDIT: is already fixed)
and the splitting of API with tags is not really nice.

I'm not a pro on openAPI and there might be a solution by using oapi-codegen with security context?
When I tried it, the context is generated _after_ the middleware callbacks so those would need to be more flexible to also accept unauthenticated requests. Please guide me to a good solution.


schuellerf:
I think I found the problem why my testcase failed...

internal/v1/no_auth/api_no_auth.go
```diff
@@ -0,0 +1,219 @@
...
+	wrapper := ServerInterfaceWrapper{
+		Handler: si,
+	}
+
+	router.GET(baseURL+"/openapi.json", wrapper.GetOpenapiJson)
+
+}
+
+// Base64 encoded, gzipped, json marshaled Swagger object
+var swaggerSpec = []string{
```
ezr-ondrej, line 127: We've moved away from inlining the spec in https://github.com/osbuild/image-builder/pull/1012

internal/v1/server.go
```diff
@@ -109,21 +110,27 @@ func Attach(conf *ServerConfig) error {
...
 	}
+
+	middlewaresNoAuth = append(middlewaresNoAuth, prometheus.PrometheusMW)
 	middlewares = append(middlewares, s.noAssociateAccounts, s.ValidateRequest, prometheus.PrometheusMW)
 
 	RegisterHandlers(s.echo.Group(fmt.Sprintf("%s/v%s", RoutePrefix(), majorVersion), middlewares...), &h)
 	RegisterHandlers(s.echo.Group(fmt.Sprintf("%s/v%s", RoutePrefix(), spec.Info.Version), middlewares...), &h)
 
+	no_auth.RegisterHandlers(s.echo.Group(fmt.Sprintf("%s/v%s", RoutePrefix(), majorVersion), middlewaresNoAuth...), &h)
+	no_auth.RegisterHandlers(s.echo.Group(fmt.Sprintf("%s/v%s", RoutePrefix(), spec.Info.Version), middlewaresNoAuth...), &h)
```
ezr-ondrej, line 132: This looks scary, can't we use the same server, but have a new function, that will register only one handler?

ezr-ondrej (commented):
Do we need a completely new server for this?
Wouldn't new function `RegisterSpecHandler` suffice? I might be missing the reson here, but IMO it could be enough to just register second group to the same echo server.
```
RegisterSpecHandler(s.echo.Group(fmt.Sprintf(fmt.Sprintf("%s/v%s", RoutePrefix(), majorVersion), middlewaresNoAuth...), &h)
RegisterSpecHandler(s.echo.Group(fmt.Sprintf(fmt.Sprintf("%s/v%s", RoutePrefix(),  spec.Info.Version), middlewaresNoAuth...), &h)
```

internal/v1/server.go
schuellerf: I didn't find an option to tell oapi-codegen to do so 🤔
and the function RegisterHandlers is generated

internal/v1/no_auth/api_no_auth.go
schuellerf: ... sorry - copy'n'paste :-/

internal/v1/server.go
ezr-ondrej: hmm gotcha... and what about having this one endpoint added manually and only the rest would be generated?
I'm not sure if we need to even have the spec serving endpoint in the spec.

ezr-ondrej: manul = having extra file (for example `api_meta.go`) for the `RegisterSpecHandler` func

schuellerf: I was more thinking that in the end we might add more endpoints to "noAuth" e.g. "/version" ?
(so avoid doing this manually)

And also having this information visible in the openAPI.yaml would be nice. That's why I wanted to use the security contexts to have it more visible but that didn't work out with the generator.

Also manually adding the route won't work as either all functions, handlers... are not generated (and need to be written manually) or it's already included in "RegisterHandlers" from what I saw oapi-codegen isn't flexible enough or I didn't find the switch to do it

schuellerf: Does it make sense to show what I mean (changing the middleware functions) by creating a second DRAFT PR so we can choose?

schuellerf:
Does it make sense to show what I mean (changing the middleware functions) by creating a second DRAFT PR so we can choose?

internal/v1/server.go
ezr-ondrej: I think it can be very simple for this,  so we wouldn't need to rewrite from scratch IMO.
But yeah, implementing the handler that serves openapi spec manually, which should be pretty straight forward.
It feels a bit cleaner than having a separate generator just for this.

It's just a suggestion of course ;)

ezr-ondrej: Handling the security in the auth middleware itself sounds like a good option tho.

ezr-ondrej:
> Does it make sense to show what I mean (changing the middleware functions) by creating a second DRAFT PR so we can choose?

Up to you, but I'd love that :orange_heart:

PR #1062, opened 2024-03-01
Enable masked systemd services in the service (HMS-3661)

kingsleyzissou:
- update v1 & composer to accept `masked` services customization
- enable this in the OpenSCAP config generator
- regenerate manifests


kingsleyzissou:
I'm still testing this. So might have to regenerate some of these again

lavocatt (dismissed):
LGTM! Thanks :rocket:

cmd/oscap/main.go
```diff
@@ -58,6 +59,25 @@ type Blueprint struct {
 	Name           string
 }
 
+func containsHelper(pkg string, service *[]string) bool {
```
mvo5, line 62: This could be simplified by just passing the slice not a point to it. This is fine and efficient as a slice itself is already a kinda pointer type. The upside is that you don't need to worry about `if service == nil` because go will DTRT when for `nil` slices. It's also slightly more canonical (but I see we have a lot of *[] in the code so maybe worth keeping as consistency is also important, sorry, still new to this codebase)

mvo5: Maybe also worth adding a TODO like: `// once we move to go 1.21 use slices.Contains() instead`

kingsleyzissou: Oh good shout for both. The nil check is a bit ugly so I think your suggestion is good. And I'll add the todo too 🙂

kingsleyzissou: It was panicking with a nil dereference error, so I simplified it slightly :)

kingsleyzissou: Hold on, it looks like I can remove this entirely now. I updated the `openscap` content and it looks like they've fixed this on their end.

lavocatt (approved):

PR #1063, opened 2024-03-01
make openapi spec endpoint public - alternatesolution (HMS-1220)

schuellerf:
This is the alternate solution over #1058 

It is similar but easier to read/maintain I guess.

--
From what I learned today the openapi YAML should rather contain the security entries so also this behavior is documented but this would mean to change all middleware functions caring about authentication validation.

### DIFF
--- a/internal/v1/server.go
+++ b/internal/v1/server.go
@@ -75,8 +75,6 @@ func Attach(conf *ServerConfig) error {
 		return err
 	}
 
-	spec.AddServer(&openapi3.Server{URL: fmt.Sprintf("%s/v%s", RoutePrefix(), spec.Info.Version)})
-
 	router, err := legacyrouter.NewRouter(spec)
 	if err != nil {
 		return err
--- a/internal/v1/server.go
+++ b/internal/v1/server.go
@@ -1,4 +1,4 @@
-//go:generate go run -mod=mod github.com/deepmap/oapi-codegen/cmd/oapi-codegen --config server.cfg.yaml -o api.go api.yaml
+//go:generate go run -mod=mod github.com/deepmap/oapi-codegen/cmd/oapi-codegen --config server.cfg.yaml api.yaml
 package v1
 
 import (
@@ -121,6 +121,8 @@ func Attach(conf *ServerConfig) error {
 
 	RegisterHandlers(s.echo.Group(fmt.Sprintf("%s/v%s", RoutePrefix(), majorVersion), middlewares...), &h)
 	RegisterHandlers(s.echo.Group(fmt.Sprintf("%s/v%s", RoutePrefix(), spec.Info.Version), middlewares...), &h)
+
+	s.echo.GET("/openapi.json", h.GetOpenapiJson)
 
 	/* Used for the livenessProbe */
 	s.echo.GET("/status", func(c echo.Context) error {
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ dnf-json
 /gen-oscap
 local.env
 __debug*
+coverage.txt
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,10 @@ help:
 	@echo "    help:               Print this usage information."
 	@echo "    build:              Build the project from source code"
 	@echo "    run:                Run the project on localhost"
+	@echo "    unit-tests:         Run unit tests (calls dev-prerequisites)"
+	@echo "    dev-prerequisites:  Install necessary development prerequisites on your system"
+	@echo "    push-check:         Replicates the github workflow checks as close as possible"
+	@echo "                        (do this before pushing!)"
 
 .PHONY: build
 build:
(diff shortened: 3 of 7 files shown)


internal/v1/server.go
```diff
@@ -135,6 +142,58 @@ func Attach(conf *ServerConfig) error {
 	})
 
 	h.server.echo.GET("/metrics", echo.WrapHandler(promhttp.Handler()))
+
+	var nonFoundRoutes []string
+
+	// as we need to implement some path registrations manually, let's check if all are there
+	fixSpecPaths := regexp.MustCompile("{(.+?)}")
+	for _, server := range spec.Servers {
```
ezr-ondrej, line 150: this part looks quite wonky, but it just mean I do not understand it by quick skim... will try to undersnad next week

ezr-ondrej (commented):
Apart of the wonky manual regexp part (and api.go gone, which is probably a mistake?), this looks much better to me :)

schuellerf:
> I do not understand, is the goal to simply expose JSON publicly? There are already three paths just like that:
> 
> ```
> 	/* Used for the livenessProbe */
> 	s.echo.GET("/status", func(c echo.Context) error {
> 		return h.GetVersion(c)
> 	})
> 
> 	/* Used for the readinessProbe */
> 	h.server.echo.GET("/ready", func(c echo.Context) error {
> 		return h.GetReadiness(c)
> 	})
> 
> 	h.server.echo.GET("/metrics", echo.WrapHandler(promhttp.Handler()))
> ```
> 
> Why don’t you do the same thing? The JSON is available directly from the `v1` package via `GetSwagger()` function.

The idea was to keep the same URL and "just" make it public. Wenn adding there, the URL would be different and would not be documented (in several places inheriting info from OpennAPI YAML) I guess

internal/v1/server.go
schuellerf: I'll wrap into a function for readability. Just maps OpenAPI syntax to internal echo-server syntax to compare them

schuellerf: ...not sure if this should go into a test though

internal/v1/server.go
```diff
@@ -135,6 +151,49 @@ func Attach(conf *ServerConfig) error {
 	})
 
 	h.server.echo.GET("/metrics", echo.WrapHandler(promhttp.Handler()))
+
+	var nonFoundRoutes []string
+
+	// as we need to implement some path registrations manually, let's check if all are there
+	mapOpenAPISyntax := getMapOpenAPISyntax()
```
ezr-ondrej, line 158: this chuck sounds like a dev helper, but nor something we need in prod, I'm not sure if we need it and if it's still just to have the `/openapi.json` route in a spec, my vote goes for getting rid of it from there.

It doesn't add any value to have it there. You need to know how to retrieve the spec to get any info from it and once you have the spec, you do not need it to contain info about how to retrieve spec - because you already have it! :)

ezr-ondrej (commented):
I agree with Lukas that we could go with really simple path.
Lets remove the `/openapi.json` path from the spec, if it simplifies the code, I'd do it.
This meta info is not very relevant.

Just note, that it'd be good to verify what we are really after, I'm just expressing my opinions here :)

schuellerf:
Obviously this last version is the simplest but also changes the URL.
@ochosi , @croissanne what are the additional actions needed if we move the URL from
/api/image-builder/v1(.0?)/openapi.json to /openapi.json

schuellerf:
I'd also love to see "the other" (unauthenticated) endpoints documented somewhere.
Should we create (and finalize) a YAML file like this somewhere, so it can be post-processed by some docu-generation?
```
---
openapi: 3.0.1
info:
  version: "1.0"
  title: Image-builder API status functions
  description: Image-builder API status functions
  license:
    name: Apache 2.0
    url: https://www.apache.org/licenses/LICENSE-2.0.html

servers:
  - url: "/"

paths:
  /openapi.json:
  get:
    summary: get the openapi json specification
    operationId: getOpenapiJson
    tags:
      - meta
    responses:
      '200':
        description: returns this document
        content:
          application/json:
            schema:
              type: object

  /status:
    get:
      summary: Used for the livenessProbe
    …

  /ready:
    get:
      summary: Used for the readinessProbe
    …  

  /metrics:
    get:
      summary:  Return some metrics
    …
```

croissanne:
> Obviously this last version is the simplest but also changes the URL. @ochosi , @croissanne what are the additional actions needed if we move the URL from `/api/image-builder/v1(.0)?/openapi.json` to `/openapi.json`

The spec does describe v1 specifically though, and if we were to add a v2 we'd be a little stuck. If we end up having to register all routes ourselves under v1/ I think that's fine.

croissanne (changes requested):
So I don't like having to break the actual api just to get around the routing quirks of the helper functions of deepmap.

https://github.com/labstack/echo/blob/fa70db801e3df89c7de8b8da161c3f41a1fe84d7/router.go#L14 is just a map, so it might be possible to just overwrite routes? In which case I'd just overwrite the public routes we want without identity middleware. That way it's also a very explicit action we do, and the default is with auth.

Also keep in mind that the allowlist of non-authed routes isn't done at this level, it's done at the gateway/3scale level. So in terms of safety we have nothing to worry about.

schuellerf:
Fixed with #1067

PR #1069, opened 2024-03-06
distributions: enable gpg check

croissanne:


distributions/centos-9/centos-9.json
```diff
@@ -10,33 +10,45 @@
...
+      "check_gpg": true
     }, {
       "id": "google-compute-engine",
       "baseurl": "https://packages.cloud.google.com/yum/repos/google-compute-engine-el9-x86_64-stable",
       "rhsm": false,
-      "image_type_tags": ["gcp"]
+      "image_type_tags": ["gcp"],
+      "check_gpg": true
```
thozza, line 28: Is the missing GPG key intentional? It should still work during the build, since the GCP below will define it, but you still list keys for all RHEL repos, which is a similar case.

croissanne: woops, added! 1 out of (?idk how many) ain't bad :D

thozza (approved):

PR #1071, opened 2024-03-07
pr_check: use new marker

jrusz:
This marker doesn't include the clone test. Also dropping provisioning from the deployment as we don't test the integration here.


ezr-ondrej (approved):
Thank you, it's making it easier to debug and cheaper, amazing! :orange_heart:

PR #1077, opened 2024-03-12
chore: update to Go 1.20

lzap:
This bumps the Go version to 1.20 as ubi9 is already on 1.20. This is the first step in trying to fix dependantbot, I still don’t know what is wrong with it.

Oh this also adds caching of the downloaded Go version (SDK) and disables separate cache of Go and Go linter so it does not fight between each other. The newly added cache now covers everything.

Edit: I updated few dependencies which should be safe (unlikely bring any code change). More to follow in a separate PR.

Edit 2: Had to bump golangci-lint version because 1.50 has a memory leak which manifests in an ugly "The runner has received a shutdown signal." error with no explanation. OOM simply kills to process.


.github/workflows/spec-update.yml
```diff
@@ -11,10 +11,10 @@ jobs:
     name: "spec update"
     runs-on: ubuntu-latest
     steps:
-      - name: Set up Go 1.19
+      - name: Set up Go 1.20
         uses: actions/setup-go@v5.0.0
         with:
-          go-version: 1.19
+          go-version: '1.20'
```
lzap, line 17: MUST be quoted otherwise YAML breaks this into `1.2` which fails.

.github/workflows/tests.yml
```diff
@@ -15,22 +15,36 @@ jobs:
...
+          path: |
+            ~/sdk
+            ~/go/bin/go*
+            ~/go/pkg/mod
+            ~/.cache/go-build
+            ~/.cache/golangci-lint/
+          key: ${{ runner.os }}-go-sdk-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-sdk
+
```
lzap, line 40: I disabled cache for linter and go actions and this now serves as a common cache for everything including the downloaded SDK (Go installation). This will resolve all the "tar overwriting" problems I was running into during upgrade. Should be also faster.

.github/workflows/tests.yml
```diff
@@ -15,22 +15,36 @@ jobs:
...
       - name: Check that source has been prepared
         run: |
           ./tools/prepare-source.sh
           if [ -n "$(git status --porcelain)" ]; then
             echo
-            echo "Please include these changes in your branch: "
-            git status -vv
+            echo "Please include these changes in your branch (only first 5k lines): "
+            git status -vv | head -n5000
```
lzap, line 47: Log of the https://github.com/osbuild/image-builder/pull/1075 PR was over 50MB in size, the diff is not very useful as github UI will not even allow to display such a huge log.

.github/workflows/tests.yml
```diff
@@ -39,7 +53,8 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.50.1
+          version: v1.56.2
+          skip-cache: true
```
lzap, line 57: Upgrade was necessary to avoid OOM kill issue.

internal/tutils/psql.go
```diff
@@ -42,7 +42,6 @@ func NewPSQLContainer() (*PSQLContainer, error) {
 	}
 
 	name := fmt.Sprintf("image_builder_test_%d", time.Now().Unix())
-	rand.Seed(time.Now().Unix())
```
lzap, line 45: This is deprecated in 1.20 and linter was complaining, seeding is automatic in 1.20+. The container is already on ubi9 1.20 so this is safe.

ondrejbudai (approved):
Thanks!

PR #1081, opened 2024-03-14
main: add echo logger to insertCompose function

mgold1234:


lzap (approved):
Thank you great work. Tests are passing now.

PR #1084, opened 2024-03-18
cmd/image-builder: log insights id

lzap:
While each log record contains request_id already, insights platform also offers insights id which is also visible to the client. Example client response from the platform:

```
HTTP/2 500 
server: openresty
content-type: application/json; charset=UTF-8
content-length: 91
x-rh-insights-request-id: e6d2fe3569a24801a25b7b34f5015210
x-content-type-options: nosniff
date: Mon, 18 Mar 2024 17:03:08 GMT
set-cookie: e15dd7c1deaec3a6b67fca153d520b0b=df2a3c785f654c64d6535cd525957cbb; path=/; HttpOnly; Secure; SameSite=None
x-rh-edge-request-id: e97847d
x-rh-edge-reference-id: 0.64f06e68.1710781388.e97847d
x-rh-edge-cache-status: NotCacheable from child
x-frame-options: SAMEORIGIN
strict-transport-security: max-age=31536000; includeSubDomains
```

This patch adds insights request id into every log line so one can easily copy and paste the returned id from the reponse HTTP header and immediately find all log entries.


cmd/image-builder/logging.go
```diff
@@ -42,24 +44,33 @@ func (h *ctxHook) Fire(e *logrus.Entry) error {
 // Extract/generate request id and store it in the standard context
 func requestIdExtractMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
 	return func(c echo.Context) error {
-		// extract or generate request id
+		// extract or generate request and insights id
 		rid := c.Request().Header.Get("X-Rh-Edge-Request-Id")
```
ezr-ondrej, line 48: Why do we prefer Edge here, would Insights make sense here, to have just single request id?

lzap: Yeah but as I explain in the description, edge-id was recommended to me on Slack, however, the problem is it is not propagated by Edge API to the client via HTTP header, only insights-id is. So I propose to use both, so it is super easy to find records if you either have one, or the other.

ezr-ondrej: okey, sure :+1:

ezr-ondrej (approved):
Sure, makes sense :+1:

ezr-ondrej:
:shipit: It's in

PR #1088, opened 2024-03-20
add logger to all functions that doesn't contain context

mgold1234:
this commit add context to all ledt function at db.go file

### DIFF
--- a/internal/common/quota.go
+++ b/internal/common/quota.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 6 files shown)


internal/common/quota.go
```diff
@@ -76,9 +77,9 @@ func CheckQuota(orgID string, dB db.DB, quotaFile string) (bool, error) {
 			return false, fmt.Errorf("No default values in the quotas' file %s\n", quotaFile)
 		}
 	}
-
+	ctx := context.Background()
 	// read user created requests
-	count, err := dB.CountComposesSince(orgID, slidingWindow)
+	count, err := dB.CountComposesSince(ctx, orgID, slidingWindow)
 	if err != nil {
```
lzap, line 83: You need to pass real context in, this is not a test, is it?

lzap (changes requested):
Great work, one missed context that needs an explanation.

lzap (approved):
Great work.

PR #1089, opened 2024-03-20
v1/api: Validate Blueprint name (HMS-3801)

ezr-ondrej:
Validates name to regexp and adds uniqueness constraint for Blueprint.

Refs HMS-3801

### DIFF
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -6,15 +6,23 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strconv"
 	"time"
 
+	"github.com/jackc/pgerrcode"
+	"github.com/jackc/pgx/v5/pgconn"
 	openapi_types "github.com/oapi-codegen/runtime/types"
 	"github.com/osbuild/image-builder/internal/common"
 	"github.com/osbuild/image-builder/internal/db"
 
 	"github.com/google/uuid"
 	"github.com/labstack/echo/v4"
+)
+
+var (
+	blueprintNameRegex         = regexp.MustCompile(`\S+`)
+	blueprintInvalidNameDetail = "The blueprint name must contain at least two characters."
 )
 
 type BlueprintBody struct {
@@ -59,6 +67,15 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
 		return err
 	}
 
+	if !blueprintNameRegex.MatchString(blueprintRequest.Name) {
+		return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
+			Errors: []HTTPError{{
+				Title:  "Invalid blueprint name",
+				Detail: blueprintInvalidNameDetail,
+			}},
+		})
+	}
+
 	id := uuid.New()
 	versionId := uuid.New()
 	ctx.Logger().Infof("Inserting blueprint: %s (%s), for orgID: %s and account: %s", blueprintRequest.Name, id, userID.OrgID(), userID.AccountNumber())
@@ -69,6 +86,16 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
 	err = h.server.db.InsertBlueprint(ctx.Request().Context(), id, versionId, userID.OrgID(), userID.AccountNumber(), blueprintRequest.Name, desc, body)
 	if err != nil {
 		ctx.Logger().Errorf("Error inserting id into db: %s", err.Error())
+
+		var e *pgconn.PgError
+		if errors.As(err, &e) && e.Code == pgerrcode.UniqueViolation {
+			return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
+				Errors: []HTTPError{{
+					Title:  "Name not unique",
+					Detail: "A blueprint with the same name already exists.",
+				}},
+			})
+		}
 		return err
 	}
 	ctx.Logger().Infof("Inserted blueprint %s", id)
@@ -121,6 +148,15 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
 	body, err := json.Marshal(blueprint)
 	if err != nil {
 		return err
+	}
+
+	if !blueprintNameRegex.MatchString(blueprintRequest.Name) {
+		return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
+			Errors: []HTTPError{{
+				Title:  "Invalid blueprint name",
+				Detail: blueprintInvalidNameDetail,
+			}},
+		})
 	}
 
 	versionId := uuid.New()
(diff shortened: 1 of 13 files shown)


internal/v1/handler_blueprints.go
```diff
@@ -17,6 +20,10 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
+const (
+	blueprintNameRegex = `^[a-zA-Z0-9][a-zA-Z0-9-_ ]+[a-zA-Z0-9]$`
```
lzap, line 24: I suggest to always put dash as the last character because this can accidentally create range.

Is this because this is turned into compose name which has these constraints?

internal/v1/handler_blueprints.go
```diff
@@ -59,11 +66,29 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
...
 	}
 
+	if !regexp.MustCompile(blueprintNameRegex).MatchString(blueprintRequest.Name) {
+		return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
+			Errors: []HTTPError{{
+				Title:  "Invalid blueprint name",
+				Detail: "The blueprint name must start and end with an alphanumeric character and may contain alphanumeric characters, underscores, hyphens, and spaces.",
+			}},
+		})
+	}
```
lzap, line 76: You can move the `MustCompile` to the constant (or `init()`) so the regexp is compiled only once, not on every request.

internal/v1/handler_blueprints.go
```diff
@@ -59,11 +66,29 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
...
+		var e *pgconn.PgError
+		if errors.As(err, &e) && e.Code == pgerrcode.UniqueViolation {
+			return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
+				Errors: []HTTPError{{
+					Title:  "Name not unique",
+					Detail: "A blueprint with the same name already exists.",
+				}},
+			})
+		}
 		ctx.Logger().Errorf("Error inserting id into db: %s", err.Error())
```
lzap, line 92: Move the log statement to the beginning of the block so it is logged on unique violation too.

internal/v1/handler_blueprints.go
```diff
@@ -119,6 +144,15 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
...
 
+	if !regexp.MustCompile(blueprintNameRegex).MatchString(blueprintRequest.Name) {
+		return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
+			Errors: []HTTPError{{
+				Title:  "Invalid blueprint name",
+				Detail: "The blueprint name must start and end with an alphanumeric character and may contain alphanumeric characters, underscores, hyphens, and spaces.",
+			}},
+		})
+	}
+
```
lzap, line 155: You should really compile just once during startup of the application :-)

lzap (changes requested):
Few nitpicks.

internal/v1/handler_blueprints.go
ezr-ondrej: nope, just me adding the regexp manually randomly ... happy to switch around
I've excluded special characters here just for convenience, but I'm transposing to image name here: #1093 so we can just drop the special characters there and allow anything here as long as it starts with word? :shrug:

ezr-ondrej: While thinking about it, I think we can allow smiley faces in the name, might be cool and doesn't hurt anything.
Let's go with just requiring non-empty (contains word)

internal/v1/handler_blueprints.go
```diff
@@ -59,12 +66,31 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
 		return err
 	}
 
+	if !blueprintNameRegex.MatchString(blueprintRequest.Name) {
+		return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
+			Errors: []HTTPError{{
+				Title:  "Invalid blueprint name",
+				Detail: "The blueprint name must start and end with an alphanumeric character and may contain alphanumeric characters, underscores, hyphens, and spaces.",
```
avitova, line 73: This does not quite match the regex, right? It can start/end with the space, we just need some characters that are not whitespace.

internal/v1/handler_blueprints.go
```diff
@@ -119,6 +145,15 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
...
```
avitova, line 152: ditto

internal/v1/handler_blueprints.go
ezr-ondrej: ah, correct, thanks!

internal/v1/handler_blueprints.go
lzap: If this is not coming from an existing constraint, everything in 2024 is UTF8 and I would not filter out non-ASCII languages like Japanese or Chinese for no reason.

lzap:
I propose not to put any constraint on the blueprint name other than length, look it works just fine with non-alphanums:

image

internal/v1/handler_blueprints.go
ezr-ondrej: in #1093 I'm transposing it to machine friendly (asci) image name, so if there is no ascii chars, it will result in `___-____-___-____` image name :thinking: not sure if that's what we want xD
But why not :)

jrusz:
> I propose not to put any constraint on the blueprint name other than length, look it works just fine with non-alphanums:
> 
> image

What about the image names then? Are all the clouds ok with such names?

lzap:
Good point, let’s see:

image

ezr-ondrej:
> Are all the clouds ok with such names?

Well, it seems they are, because we do not send the names of images to the cloud, we send uuids instead today xD
So once we will send the image names, it might get tricky, but I suppose we can fix that on image level and do not limit blueprints for that :+1:

jrusz:
Hmm, so if it sends the UUID then it's fine to have whatever blueprint name, but I suppose customers who were used to name their image and find it under the same name in the cloud they would not appreciate this change and I suppose we want to keep the current behavior...

ezr-ondrej:
> I suppose customers who were used to name their image and find it under the same name in the cloud they would not appreciate this change and I suppose we want to keep the current behavior

It always worked this way (using UUID in the cloud). We should fix it so customers can find the image by it's expected name (and we were targeting it already), but IMO that's out of scope here :)

lzap:
I am pretty sure that AWS tags do support UTF-8 so it will work on EC2, if any other cloud does not let us send non-alphanum image names, well, we can just keep using UUID in the future and ask customers to create tickets for them.

avitova:
LGTM:)

ezr-ondrej:
Any other comments @lzap ? :)

ezr-ondrej:
This is :green_apple: any takers? :)

internal/v1/handler_blueprints.go
```diff
@@ -17,6 +20,11 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
+var (
+	blueprintNameRegex         = regexp.MustCompile(`\S+`)
+	blueprintInvalidNameDetail = "The blueprint name must contain at least two characters."
```
lzap, line 25: Hmm isn’t `\S+` one or more characters?

lzap (approved):
Thanks.

PR #1093, opened 2024-03-21
v1/api: Image name transposition from Blueprint name (HMS-3801)

ezr-ondrej:
On composing Blueprint, transpose its name to image name,
that's machine friendly, unlike Blueprint name that's human readable.

Refs HMS-3801

Follow-up for #1089

### DIFF
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/jackc/pgerrcode"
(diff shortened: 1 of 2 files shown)


lzap:
Not a fan of ignoring nations who do not use latin characters, Go UTF implementation is really good and there are simple ways now to filter out non-pritable characters while keeping the others. If alphanum is a requirement by composer, maybe we can talk about lifting this requirement. If that cannot be achieved, well, this is the way to go then.

ezr-ondrej:
> If that cannot be achieved, well, this is the way to go then.

I'm actually not sure where there is no support for UTF, I think it was an Azure image name. But I agree we might just fix it per the offender and not everywhere :thinking:

ezr-ondrej:
ok, lets abandon it for now ;)

PR #1096, opened 2024-03-22
Add Installer customizations to API

achilleas-k:
Update the specs and add the new Installer customizations to the API.

See https://github.com/osbuild/osbuild-composer/pull/4005 and further links.

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -1479,6 +1479,8 @@ components:
             even when there are one or more mountpoints.
         fips:
           $ref: '#/components/schemas/FIPS'
+        installer:
+          $ref: '#/components/schemas/Installer'
     Container:
       type: object
       required:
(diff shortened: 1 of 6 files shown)


internal/v1/api.yaml
```diff
@@ -1684,6 +1686,21 @@ components:
...
           default: false
+    Installer:
+      type: object
+      additionalProperties: false
+      description: Anaconda installer configuration
+      properties:
+        unattended:
+          type: boolean
+          description: |
+            Create a kickstart file for a fully automated installation
```
lzap, line 1697: Like on the installed system in `/root`?

lzap (approved):
LGTM

croissanne (approved):

croissanne (approved):

internal/v1/api.yaml
achilleas-k: On the ISO.  It's under the `Installer` customizations which are described as `Anaconda installer configuration` and only apply to ISO builds.

croissanne (approved):

PR #1104, opened 2024-03-27
api/v1: Blueprint description is optional

ezr-ondrej:


ezr-ondrej:
:green_apple: @lzap any comments? :)

lzap (approved):
No comments, tho I prefer using empty strings in Go in general as pointers are bug prone. This is fine, it follows good practice in OpenAPI I guess. TY

ezr-ondrej:
> OpenAPI I guess

yeah when not required, it does get generated as pointer  :/

PR #1110, opened 2024-04-02
Explicitly require `firewalld` package for OpenSCAP profiles that require it

kingsleyzissou:
The oscap blueprint generation tool assumes that the `firewalld` package will be installed on the image.
The image build fails if this package is not included, so we should explictly include the package for the profiles
that require the `firewalld` service to be enabled.

Automatic regeneration of these profiles could be done in a follow-up PR.

See: https://github.com/ComplianceAsCode/content/issues/11780


ezr-ondrej (approved):
Looks pretty sane to me :)

PR #1114, opened 2024-04-04
Build images based on snapshots made by the content sources service HMS-2950

croissanne:


internal/clients/content_sources/client.go
```diff
@@ -0,0 +1,90 @@
+//go:generate go run -mod=mod github.com/deepmap/oapi-codegen/v2/cmd/oapi-codegen --config client.cfg.yaml content-sources.v1.json
+
```
lzap, line 2: Just a note: this file has a go generate comment, but this particular file is NOT generated.

If I can suggest suggest creating an empty `package.go` file with only this line separately to prevent this confusion. I skipped reviewing this file in other review because I believed it was auto-generated... :-D

internal/v1/handler_compose_image.go
```diff
@@ -180,6 +190,90 @@ func buildRepositories(arch *distribution.Architecture, imageType ImageTypes) []
 	return repositories
 }
 
+func (h *Handlers) buildRepositorySnapshots(ctx echo.Context, arch *distribution.Architecture, imageType ImageTypes, snapshotDate *string) ([]composer.Repository, error) {
+	date, err := time.Parse(time.RFC3339, *snapshotDate)
```
lzap, line 194: Nil pointer error when snapshot is `nil`, since it is not required in OpenAPI I think it can be.

internal/v1/handler_compose_image.go
```diff
@@ -180,6 +190,90 @@ func buildRepositories(arch *distribution.Architecture, imageType ImageTypes) []
...
+	var csRepos content_sources.ApiRepositoryCollectionResponse
+	err = json.NewDecoder(resp.Body).Decode(&csRepos)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, repo := range *csRepos.Data {
+		repoUUIDs = append(repoUUIDs, *repo.Uuid)
+		repoMap[*repo.Uuid] = repo
+	}
```
lzap, line 230: Nil pointers, just saying. I wonder why OpenAPI fields are optional by default, that is tragic.

internal/v1/handler_compose_image.go
```diff
@@ -180,6 +190,90 @@ func buildRepositories(arch *distribution.Architecture, imageType ImageTypes) []
...
+
+	// A sanity check to make sure there's a snapshot for each repo
+	expected := len(buildRepositories(arch, imageType))
+	if len(repositories) != expected {
+		return repositories, fmt.Errorf("No snapshots found for all repositories (found %d, expected %d)", len(repositories), expected)
+	}
+	return repositories, nil
+}
+
```
lzap, line 276: Allright, I really hope the next commit introduces a test for this :-D

I would not mind seeing little bit of debug logging in this function.

internal/v1/handler_compose_image.go
```diff
@@ -180,6 +190,90 @@ func buildRepositories(arch *distribution.Architecture, imageType ImageTypes) []
...
+				contains = true
+				break
+			}
+		}
+		if contains {
+			repoURLs = append(repoURLs, *r.Baseurl)
+		}
+	}
+
+	repoUUIDs := []string{}
```
lzap, line 213: Maybe

```
var repoUUIDs []string{}
```

as append does initialize by default, but no biggie. Ideally I would call `make` to pre-allocate to the correct size but this is fine.

internal/clients/content_sources/client.go
```diff
@@ -0,0 +1,90 @@
...
+package content_sources
```
lzap, line 3: Terrible package name by the way, this will be painful to write or read.

internal/v1/handler_blueprints_test.go
```diff
@@ -24,7 +24,7 @@ func TestHandlers_CreateBlueprint(t *testing.T) {
 	dbase, err := dbc.NewDB()
 	require.NoError(t, err)
 
-	db_srv, tokenSrv := startServer(t, "", "", &ServerConfig{
+	db_srv, tokenSrv := startServer(t, &testServerClientsConf{}, &ServerConfig{
```
lzap, line 27: I was gonna say I don’t like `startServer` it is going off-hands with another client being added in other PR BUT you actually make it better by introducing a struct that is easier on eyes as things can be skipped. Nice. @mgold1234 you will need to rebase on top of this.

internal/v1/server_test.go
```diff
@@ -90,7 +90,12 @@ func makeUploadOptions(t *testing.T, uploadOptions interface{}) *composer.Upload
 	return &result
 }
 
-func startServer(t *testing.T, composerURL, provURL string, conf *ServerConfig) (*echo.Echo, *httptest.Server) {
+type testServerClientsConf struct {
+	ComposerURL string
+	ProvURL     string
+}
```
lzap, line 98: 🆗 👍🏻

lzap (dismissed):
Great work, couple of irrelevant nitpicks.

internal/v1/handler_compose_image.go
croissanne: The check on line 66 will be sufficient, but the signature makes it weird. I'll change this pointer into just a reference.

internal/v1/handler_compose_image.go
croissanne: So I think in the case of content sources it's fine to let it panic, as I don't expect these ever to not be set, correct @jlsherrill? We can add checks but we'd just 500 anyway I think.

internal/v1/handler_compose_image.go
```diff
@@ -173,6 +166,116 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
...
+	snapResp, err := h.server.csClient.GetSnapshotsForDate(ctx.Request().Context(), content_sources.ApiListSnapshotByDateRequest{
+		Date:            common.ToPtr(date.UTC().Format(time.DateOnly)),
+		RepositoryUuids: common.ToPtr(repoUUIDs),
+	})
+	if err != nil {
+		return nil, err
+	}
+	defer closeBody(ctx, snapResp.Body)
+
+	var csSnapshots []content_sources.ApiListSnapshotByDateResponse
```
ezr-ondrej, line 241: Ugh, it's a shame that we have root array here, that smells, should we ask Content to change this before we start depending on it?

croissanne: maybe, not sure. Could also merge and ask at the same time :innocent:

croissanne:
IQE is broken currently because of some registry issues.

internal/v1/handler_compose_image.go
jlsherrill: correct, they should always be set, if return a repo without a uuid something has gone tragically wrong.

lzap:
It was fixed 2 hours ago.

/retest

internal/v1/server_test.go
```diff
@@ -90,7 +91,13 @@ func makeUploadOptions(t *testing.T, uploadOptions interface{}) *composer.Upload
 	return &result
 }
 
-func startServer(t *testing.T, composerURL, provURL string, conf *ServerConfig) (*echo.Echo, *httptest.Server) {
+type testServerClientsConf struct {
+	ComposerURL string
+	ProvURL     string
+	CSURL       string
```
mgold1234, line 97: maybe change CSURL to ContentSourceURL? it will be more clear

lzap:
Tests are still failing, this must be green no matter what.

croissanne:
Composer stage is very overloaded atm, waiting list is around 2 hours.

ezr-ondrej:
The content wrapper is in, can you use the new API @croissanne pls :pray:

internal/v1/handler_compose_image.go
ezr-ondrej: It's updated, can we use the new API :pray: :)

lzap:
Failure once again :-(

croissanne:
Stage is just overloaded again ._.

lzap (approved):
Some notes (can do a followup myself):

* Ugly package name `content_sources`
* The `go:generate` metacomment should live in a separate empty file to avoid confusion
* Bunch of copy-pasted code in `clients.go`

But this was a lot of work and pain to get it green. Merging, good stuff!

PR #1119, opened 2024-04-05
distributions: switch latest minor over to rolling distros HMS-2950

croissanne:
distributions: switch latest minor over to rolling distros
    
 There is no difference in content, and the content sources service needs
    the rolling distro URL to find snapshots.
    
 Once rhel-9.4 and rhel-8.10 is introduced, the 8.9 and 9.3 URLs should
    switch over to their respective minor version repositories.


lzap:
Can be rebased.

lzap (approved):
Looks sane. I haven’t tested this and I am also not sure what these JSON are for.

PR #1121, opened 2024-04-08
internal/composer: reseek after lazy token refresh

croissanne:


lzap:
I do not like how this is copy-pasted around, I would suggest to create a dedicated type called LazyToken that would actually do the refreshing and it could also track expiration easily too, the information about expiration comes in the response. Then each client could create its own instance of this type to cleanly do lazy-initialized tokens when needed.

I am fine keeping status-quo tho, feel free to rebase I merged your big PR. Thanks.

lzap:
Anyway, this was a draft, feel free to rebase this is a correct fix.

PR #1123, opened 2024-04-09
config: fetch content sources url dynamically

croissanne:


lzap (approved):
Ah sure, no `clawdapp.yaml` change required, huh? Not sure.

croissanne:
> Ah sure, no `clawdapp.yaml` change required, huh? Not sure.

it's already in optional dependencies, so I think it should be fine

PR #1124, opened 2024-04-09
config: fix content-sources deployment name

croissanne:
The specific deployment that ib needs to target is called `service`.

---
see https://github.com/content-services/content-sources-backend/blob/de7cd3642ba312628bafe412857d7518c91d003e/deployments/deployment.yaml#L150


bcl (approved):
Seems obvious :)

lzap:
When its YAML, it is literally obvious.

PR #1131, opened 2024-04-11
v1/api: Allow fetching Blueprint by name (HMS-3801)

ezr-ondrej:
Add fetch by name in the collection endpoint.
It is bit weird when name needs to be unique.
We can do pattern search on the name in the future.


ezr-ondrej:
WDYT @lzap? :thinking:
I feel bit weird, but it's the best I could think of :shrug:

ezr-ondrej:
Let me draft it because it doesn't work now xD
But I hope you get the idea and can ack/nack it :)

lzap (approved):

PR #1133, opened 2024-04-14
refactor token in image builder (HMS-3921)

mgold1234:
this commit refactor token in this project:
1) Create a new file lazy_token.go
2) create NextToken(context.Context) string function that will check if token exists already and is not expiring in 1 minute or not expired. in other case it will require a new token, stores expiration time as well as the token itself and return it
4) Create a dummy implementation of Tokener that will return just a string "testtoken"
5) Create an interface named Tokener with the Token(ctx) method

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -3,11 +3,12 @@ package main
 import (
 	"fmt"
 
-	"github.com/osbuild/image-builder/internal/clients/recommendations"
+	"github.com/osbuild/image-builder/internal/oauth2"
 
 	"github.com/osbuild/image-builder/internal/clients/composer"
 	"github.com/osbuild/image-builder/internal/clients/content_sources"
 	"github.com/osbuild/image-builder/internal/clients/provisioning"
+	"github.com/osbuild/image-builder/internal/clients/recommendations"
 	"github.com/osbuild/image-builder/internal/common"
 	"github.com/osbuild/image-builder/internal/config"
 	"github.com/osbuild/image-builder/internal/db"
@@ -95,11 +96,13 @@ func main() {
 	}
 
 	recommendationConf := recommendations.RecommendationsClientConfig{
-		URL:          conf.RecommendURL,
-		CA:           conf.RecommendCA,
-		TokenURL:     conf.RecommendTokenURL,
-		ClientId:     conf.RecommendClientId,
-		ClientSecret: conf.RecommendSecret,
+		URL: conf.RecommendURL,
+		CA:  conf.RecommendCA,
+		Tokener: &oauth2.LazyToken{
+			Url:          conf.RecommendTokenURL,
+			ClientId:     conf.RecommendClientId,
+			ClientSecret: conf.RecommendSecret,
+		},
 	}
 	compClient, err := composer.NewClient(composerConf)
 	if err != nil {
--- a/internal/clients/recommendations/client.go
+++ b/internal/clients/recommendations/client.go
@@ -8,35 +8,24 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
-	"sync"
 
+	"github.com/osbuild/image-builder/internal/oauth2"
 	"github.com/sirupsen/logrus"
 )
 
 type RecommendationsClient struct {
-	URL          string
-	tokenURL     string
-	accessToken  string
-	clientId     string
-	clientSecret string
-	client       *http.Client
-	tokenMu      sync.RWMutex
+	URL     string
+	tokener oauth2.Tokener
+	client  *http.Client
 }
 
 type RecommendationsClientConfig struct {
-	URL          string
-	CA           string
-	TokenURL     string
-	ClientId     string
-	ClientSecret string
-}
-
-type tokenResponse struct {
-	AccessToken string `json:"access_token"`
+	URL     string
+	CA      string
+	Tokener oauth2.Tokener
 }
 
 var contentHeaders = map[string]string{"Content-Type": "application/json"}
@@ -51,11 +40,9 @@ func NewClient(conf RecommendationsClientConfig) (*RecommendationsClient, error)
 	}
 
 	rc := RecommendationsClient{
-		URL:          conf.URL,
-		tokenURL:     conf.TokenURL,
-		clientId:     conf.ClientId,
-		clientSecret: conf.ClientSecret,
-		client:       client,
+		URL:     conf.URL,
+		tokener: conf.Tokener,
+		client:  client,
 	}
 
 	return &rc, nil
@@ -98,24 +85,23 @@ func (rc *RecommendationsClient) request(method, url string, headers map[string]
 		req.Header.Add(k, v)
 	}
 
-	token := func() string {
-		rc.tokenMu.RLock()
-		defer rc.tokenMu.RUnlock()
-		return rc.accessToken
+	token, err := rc.tokener.Token(req.Context())
+	if err != nil {
+		return nil, err
 	}
-	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token()))
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
 
 	resp, err := rc.client.Do(req)
 	if err != nil {
 		return nil, err
 	}
 	if resp.StatusCode == http.StatusForbidden || resp.StatusCode == http.StatusUnauthorized {
-		err = rc.refreshToken()
+		token, err = rc.tokener.ForceRefresh(req.Context())
 		if err != nil {
 			return nil, err
 		}
 
-		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token()))
+		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
 		_, err = body.Seek(0, io.SeekStart)
 		if err != nil {
 			return nil, err
@@ -123,37 +109,6 @@ func (rc *RecommendationsClient) request(method, url string, headers map[string]
 		resp, err = rc.client.Do(req)
 	}
 	return resp, err
-}
-
-func (rc *RecommendationsClient) refreshToken() error {
-	rc.tokenMu.Lock()
-	defer rc.tokenMu.Unlock()
-
-	data := url.Values{}
-	if rc.clientSecret != "" {
-		data.Set("grant_type", "client_credentials")
-		data.Set("client_id", rc.clientId)
-		data.Set("client_secret", rc.clientSecret)
-	}
-
-	resp, err := http.PostForm(rc.tokenURL, data)
-	if err != nil {
-		return err
-	}
-	defer func() {
-		if err := resp.Body.Close(); err != nil {
-			logrus.Errorf("Error closing body after refreshing composer client token: %v", err)
-		}
-	}()
-
-	var tr tokenResponse
-	err = json.NewDecoder(resp.Body).Decode(&tr)
-	if err != nil {
-		return err
-	}
-
-	rc.accessToken = tr.AccessToken
-	return nil
 }
 
 func (rc *RecommendationsClient) RecommendationsPackages(recommendationsPackages RecommendPackageRequest) (*http.Response, error) {
--- a/internal/oauth2/dummy_token.go
+++ b/internal/oauth2/dummy_token.go
@@ -1,0 +1,16 @@
+package oauth2
+
+import "context"
+
+// DummyToken is a dummy implementation of the Tokener interface.
+type DummyToken struct{}
+
+// NextToken returns a static "testtoken" string.
+func (dt *DummyToken) NextToken(ctx context.Context) (string, error) {
+
+	return "testtoken", nil
+}
+
+func (dt *DummyToken) ForceRefresh(ctx context.Context) (string, error) {
+	return "", nil
+}
--- a/internal/oauth2/lazy_token.go
+++ b/internal/oauth2/lazy_token.go
@@ -1,0 +1,106 @@
+package oauth2
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"sync"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Tokener is an interface that defines the AccessToken method.
+type Tokener interface {
+	Token(context.Context) (string, error)
+	ForceRefresh(context.Context) (string, error)
+}
+
+type LazyToken struct {
+	// Url represents the URL used for acquiring the token.
+	Url string
+	// ClientId is the client ID used for authentication.
+	ClientId string
+	// ClientSecret is the client secret used for authentication.
+	ClientSecret string
+	// AccessToken string holds the currently cached token.
+	AccessToken string
+	// Expiration stores the expiration time of the current token.
+	Expiration time.Time
+	// mutex ensures safe concurrent access to the token and expiration fields.
+	mutex sync.Mutex
+}
+
+type tokenResponse struct {
+	AccessToken string `json:"access_token"`
+	ExpiresIn   int    `json:"expires_in"`
+}
+
+func (lt *LazyToken) acquireNewToken(ctx context.Context, forceRefresh bool) (string, error) {
+	lt.mutex.Lock()
+	defer lt.mutex.Unlock()
+
+	if forceRefresh || lt.AccessToken == "" || time.Now().Add(time.Minute).After(lt.Expiration) {
+		tokenRes, err := lt.requestToken()
+		if err != nil {
+			return "", err
+		}
+
+		lt.AccessToken = tokenRes.AccessToken
+		lt.Expiration = time.Now().Add(time.Duration(tokenRes.ExpiresIn) * time.Second)
+
+		logrus.WithContext(ctx).Infof("Acquired new token %s with expiration at: %s", lt.AccessToken, lt.Expiration)
+	} else {
+		logrus.WithContext(ctx).Infof("AccessToken reused: %s (expires at %s)", lt.AccessToken, lt.Expiration)
+	}
+
+	return lt.AccessToken, nil
+}
+
+func (lt *LazyToken) Token(ctx context.Context) (string, error) {
+	return lt.acquireNewToken(ctx, false)
+}
+
+// ForceRefresh is a function that responsible for fetching a new access token.
+func (lt *LazyToken) ForceRefresh(ctx context.Context) (string, error) {
+	return lt.acquireNewToken(ctx, true)
+}
+
+// ForceRefresh forces the acquisition of a new access token by clearing the current AccessToken and calling Token().
+func (lt *LazyToken) requestToken() (*tokenResponse, error) {
+	data := url.Values{}
+	data.Set("client_id", lt.ClientId)
+	if lt.ClientSecret != "" {
+		data.Set("grant_type", "client_credentials")
+		data.Set("client_secret", lt.ClientSecret)
+	} else {
+		return nil, fmt.Errorf("client Id, client Secret and token must be set")
+	}
+	resp, err := http.PostForm(lt.Url, data)
+
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			// Handle error reading response body
+			return nil, fmt.Errorf("failed to read response body: %v", err)
+		}
+		defer resp.Body.Close()
+		return nil, fmt.Errorf("request failed with status code %d: %s", resp.StatusCode, body)
+	}
+
+	var tokenResp tokenResponse
+	err = json.NewDecoder(resp.Body).Decode(&tokenResp)
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshaling token response: %v", err)
+	}
+
+	return &tokenResp, nil
+}
--- a/internal/oauth2/lazy_token_test.go
+++ b/internal/oauth2/lazy_token_test.go
@@ -1,0 +1,85 @@
+package oauth2
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestLazyToken(t *testing.T) {
+	nextTokenID := 0
+	mockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		nextTokenID++
+		w.Header().Set("Content-Type", "application/json")
+		tokenResponse := tokenResponse{
+			AccessToken: fmt.Sprintf("mock-token-%d", nextTokenID),
+			ExpiresIn:   3600, // expires in 1 hour
+		}
+		err := json.NewEncoder(w).Encode(tokenResponse)
+		require.NoError(t, err)
+		clientID := r.FormValue("client_id")
+		require.Equal(t, "test-client-id", clientID)
+
+		clientSecret := r.FormValue("client_secret")
+		require.Equal(t, "test-client-secret", clientSecret)
+
+		grantType := r.FormValue("grant_type")
+		require.Equal(t, "client_credentials", grantType)
+	}))
+	defer mockServer.Close()
+
+	clientID := "test-client-id"
+	clientSecret := "test-client-secret"
+	lazyToken := &LazyToken{
+		Url:          mockServer.URL,
+		ClientId:     clientID,
+		ClientSecret: clientSecret,
+	}
+	ctx := context.Background()
+	token, err := lazyToken.Token(ctx)
+	require.NoError(t, err)
+	require.Equal(t, "mock-token-1", token)
+
+	token, err = lazyToken.Token(ctx)
+	require.NoError(t, err)
+	require.Equal(t, "mock-token-1", token)
+
+	// generates a new token when token expired
+	lazyToken.Expiration = time.Now().Add(-time.Minute) // Expire the token
+	token, err = lazyToken.Token(ctx)
+	require.NoError(t, err)
+	require.Equal(t, "mock-token-2", token)
+	require.True(t, lazyToken.Expiration.After(time.Now()), "Expiration should be in the future")
+
+	// Calling ForceRefresh generates a new token
+	token, err = lazyToken.ForceRefresh(ctx)
+	require.NoError(t, err)
+	require.Equal(t, "mock-token-3", token)
+}
+
+func TestLazyTokenUnhappy(t *testing.T) {
+	// Set up a mock token server that always returns an HTTP error
+	mockTokenServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		http.Error(w, "Mock token server error", http.StatusInternalServerError)
+	}))
+	defer mockTokenServer.Close()
+	clientID := "test-client-id"
+	clientSecret := "test-client-secret"
+	lazyToken := &LazyToken{
+		Url:          mockTokenServer.URL,
+		ClientId:     clientID,
+		ClientSecret: clientSecret,
+	}
+
+	ctx := context.Background()
+
+	// Ensure that NextToken() returns an error when the token server responds with an error
+	_, err := lazyToken.Token(ctx)
+	require.Error(t, err)
+}
--- a/internal/v1/handler_recommend_package_test.go
+++ b/internal/v1/handler_recommend_package_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/osbuild/image-builder/internal/tutils"
 	"github.com/stretchr/testify/require"
@@ -179,6 +180,62 @@ func TestRecommendPackage_Success_with_packages(t *testing.T) {
 	}
 	respStatusCode, body := tutils.PostResponseBody(t, apiSrv.URL, payload)
 	require.Equal(t, http.StatusCreated, respStatusCode)
+	var result RecommendationsResponse
+	expectedResult := RecommendationsResponse{
+		Packages: []string{"vim", "python"},
+	}
+	err := json.Unmarshal([]byte(body), &result)
+	require.NoError(t, err)
+	require.Equal(t, expectedResult, result)
+}
+
+func TestRecommendPackage_with_authenticationServer(t *testing.T) {
+	apiSrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		result := RecommendationsResponse{
+			Packages: []string{"vim", "python"},
+		}
+		err := json.NewEncoder(w).Encode(result)
+		require.NoError(t, err)
+		require.Equal(t, "Bearer dog_token", r.Header.Get("Authorization"))
+	}))
+	defer apiSrv.Close()
+
+	oauthMock := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		require.NoError(t, r.ParseForm())
+		require.Equal(t, "client_credentials", r.PostForm.Get("grant_type"))
+		require.Equal(t, "secret", r.PostForm.Get("client_secret"))
+		require.Equal(t, "id", r.PostForm.Get("client_id"))
+
+		err := json.NewEncoder(w).Encode(struct {
+			AccessToken string    `json:"access_token"`
+			Expiration  time.Time `json:"expiration"`
+		}{
+			AccessToken: "dog_token",
+			Expiration:  time.Now().Add(time.Hour),
+		})
+
+		require.NoError(t, err)
+	}))
+	defer oauthMock.Close()
+
+	srv, tokenSrv := startServer(t, &testServerClientsConf{RecommendURL: apiSrv.URL, OAuthURL: oauthMock.URL}, &ServerConfig{})
+	defer func() {
+		err := srv.Shutdown(context.Background())
+		require.NoError(t, err)
+	}()
+	defer tokenSrv.Close()
+	payload := RecommendPackageRequest{
+		Packages: []string{
+			"some",
+			"packages",
+		},
+		RecommendedPackages: func() int32 {
+			recommendedPackages := int32(3)
+			return recommendedPackages
+		}(),
+	}
+	respStatusCode, body := tutils.PostResponseBody(t, "http://localhost:8086/api/image-builder/v1/experimental/recommendations", payload)
+	require.Equal(t, http.StatusOK, respStatusCode)
 	var result RecommendationsResponse
 	expectedResult := RecommendationsResponse{
 		Packages: []string{"vim", "python"},
--- a/internal/v1/server_test.go
+++ b/internal/v1/server_test.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/osbuild/image-builder/internal/clients/recommendations"
+	"github.com/osbuild/image-builder/internal/oauth2"
 
 	"github.com/labstack/echo/v4"
 	"github.com/sirupsen/logrus"
@@ -98,6 +99,7 @@ type testServerClientsConf struct {
 	ProvURL      string
 	CSURL        string
 	RecommendURL string
+	OAuthURL     string
 }
 
 func startServer(t *testing.T, tscc *testServerClientsConf, conf *ServerConfig) (*echo.Echo, *httptest.Server) {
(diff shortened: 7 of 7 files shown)


internal/clients/composer/client.go
```diff
@@ -124,8 +123,8 @@ func (cc *ComposerClient) request(method, url string, headers map[string]string,
 		if err != nil {
 			return nil, err
 		}
-
-		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token()))
+		token = cc.Tokener.Token(req.Context()) // Use the Token method from the Tokener
+		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
 		resp, err = cc.client.Do(req)
 	}
```
lzap, line 129: The whole purpose of this refactoring is to get rid of the double request. Remove this, a client request becomes a single request, you will need to update many tests and that is expected.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,123 @@
...
+func (lt *LazyToken) Token(ctx context.Context) string {
+	//echoCtx, ok := ctx.(echo.Context)
+	//	if !ok {
+	//	// Handle the case where the context is not an echo.Context
+	//	return ""
+	//	}
+	lt.mutex.Lock()
+	defer lt.mutex.Unlock()
+
+	if lt.token == "" || lt.expiration.Sub(time.Now()) < time.Minute {
```
lzap, line 56: The `lt.expiration` can be `nil` and it is not being checked.

Before you add `nil` check, here is a trick to consider - in the initializer function, set it to current time. Thus it will always be set to something and you can leave this code as is.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+// Token returns a valid token, acquiring a new one if necessary.
+func (lt *LazyToken) Token(ctx context.Context) string {
+	//echoCtx, ok := ctx.(echo.Context)
+	//	if !ok {
+	//	// Handle the case where the context is not an echo.Context
+	//	return ""
+	//	}
```
lzap, line 52: Do not do this please, no runtime type checks/casts.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// NewLazyToken creates a new LazyToken instance.
+func NewLazyToken(url, clientId, clientSecret string) *LazyToken {
+	return &LazyToken{
+		url:          url,
+		clientId:     clientId,
+		clientSecret: clientSecret,
+	}
+}
```
lzap, line 44: Please create another similar initializer function that allows to create new token with offline token too. Name it appropriately.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,123 @@
...
+	//echoCtx, ok := ctx.(echo.Context)
+	//	if !ok {
+	//	// Handle the case where the context is not an echo.Context
+	//	return ""
+	//	}
+	lt.mutex.Lock()
+	defer lt.mutex.Unlock()
+
+	if lt.token == "" || lt.expiration.Sub(time.Now()) < time.Minute {
+		//	echoCtx.Logger().Info("Acquiring new token")
```
lzap, line 57:
```suggestion
		logrus.WithContext(ctx).Infof("Acquired new token with expiration at: %s", expires.Format("xxx"))
```

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		// Handle error
+		return ""
+	}
+
+	return tokenResp.AccessToken
+}
+
+// tokenExpirationDuration is a placeholder for the actual token expiration duration.
+const tokenExpirationDuration = time.Hour
```
lzap, line 123: What is this for? Expiration is returned in the response, get rid of this.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,123 @@
...
+		lt.expiration = time.Now().Add(tokenExpirationDuration)
+	} else {
+		//	echoCtx.Logger().Debug("Token requested: %s (expires at %s)\n", lt.token, lt.expiration)
+	}
+
+	return lt.token
+}
+
+// AcquireToken is a placeholder function that simulates token acquisition.
+func AcquireToken(tokenURL, clientID, clientSecret, offlineToken string) string {
```
lzap, line 70: This function must be not exported, the `Tokener` interface is supposed to be the only way how to work with this package.

```suggestion
func acquireToken(tokenURL, clientID, clientSecret, offlineToken string) string {
```

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,123 @@
...
+		// Handle error response
+		fmt.Println("Error acquiring token:", string(body))
+		return ""
+	}
+
+	var tokenResp tokenResponse
+	err = json.Unmarshal(body, &tokenResp)
+	if err != nil {
+		// Handle error
+		return ""
```
lzap, line 116: Yeah, refactor this helper function to return error as well, refactor the `Token(...)` method to return error as well.

lzap (changes requested):
The initial version looks good, please continue.

internal/clients/composer/client.go
```diff
@@ -8,25 +8,22 @@ import (
...
 	tokenURL     string
 	offlineToken string
-	accessToken  string
 	clientId     string
 	clientSecret string
-	tokenMu      sync.RWMutex
+	Tokener      oauth2.Tokener
 
 	client *http.Client
 }
```
lzap, line 29: Client no longer needs anything else than:

* a URL
* tokener
* http.Client

As tokener will hold the rest (client id/secret, token URL, offline/access token).

internal/clients/composer/client.go
```diff
@@ -38,23 +35,20 @@ type ComposerClientConfig struct {
 	ClientId     string
 	OfflineToken string
 	ClientSecret string
-}
-
-type tokenResponse struct {
-	AccessToken string `json:"access_token"`
+	Tokener      oauth2.Tokener
 }
```
lzap, line 39: The same comment. You could create config without tokener, but it is useful for better testing.

internal/config/config.go
```diff
@@ -15,6 +15,7 @@ type ImageBuilderConfig struct {
 	ComposerClientId      string `env:"COMPOSER_CLIENT_ID"`
 	ComposerOfflineToken  string `env:"COMPOSER_OFFLINE_TOKEN"`
 	ComposerClientSecret  string `env:"COMPOSER_CLIENT_SECRET"`
+	ComposerTokener       string `env:"COMPOSER_TOKENER"`
```
lzap, line 18: What is this for?

All the input configuration for tokener is already present. I am not following what is the idea here.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,118 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Tokener is an interface that defines the Token method.
+type Tokener interface {
+	Token(context.Context) string
```
lzap, line 19:
```suggestion
	Token(context.Context) (string, error)
```

And handle errors correctly, return them via this method.

internal/oauth2/lazy_token.go
lzap: As I said, instead of one "contructor" functions, I suggest to create two of them:

* `NewLazyToken(url, clientId, clientSecret) *LazyToken`
* `NewOfflineLazyToken(url, clientId, offlineToken string) *LazyToken`

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,118 @@
...
+		expiration:   time.Now(),
+	}
+}
+
+// Token returns a valid token, acquiring a new one if necessary.
+func (lt *LazyToken) Token(ctx context.Context) string {
+	lt.mutex.Lock()
+	defer lt.mutex.Unlock()
+
+	if lt.token == "" || lt.expiration.Before(time.Now().Add(time.Minute)) {
```
lzap, line 56: The time condition is I believe not correct. Token should be refreshed if:

* Token hasn’t been fetched yet (is empty)
* Current time is older than expiration time of already fetched token minus one minute.

Write test for both use cases.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,118 @@
...
+
+		// Acquire a new token and set the expiration time
+		lt.token, _ = acquireToken(lt.url, lt.clientId, lt.clientSecret, lt.offlineToken)
```
lzap, line 59: Check errors and return them properly.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,118 @@
...
+		lt.expiration = time.Now().Add(time.Hour)
```
lzap, line 60: Replace the hardcoded one hour with expiration returned by the token service.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,118 @@
...
+
+	var tokenResp tokenResponse
+	err = json.Unmarshal(body, &tokenResp)
+	if err != nil {
+		fmt.Println("Error unmarshal token response:", tokenResp)
+		return "", err
+	}
+
+	return tokenResp.AccessToken, nil
+}
```
lzap, line 118: I suggest to update this function to return `tokenResponse` so you can work with the data in the caller.

lzap (changes requested):
Good progress. Lots of comments, I think it is the time to introduce tests for Tokenizer. It will reveal some of the issues with the current code.

lzap (changes requested):
This does not seem correct, the init function `NewLazyToken` seems unused. Also, you removed TokenURL and client id/secret from configuration. Please keep these settings as is, just refactor the way how clients are instantiated.

One more thing, please read about OAuth 2.0 and how token refreshing works. I think we can further improve the lazy token - we can drop `offlineToken` completely, it is just a token. Lazy tokener can be instantiated with an existing token, when it is expired (that is the default) then `refresh_token` grant is requested which will cause a token to be replaced with a new one. If token is empty (was not provided) then `client_credentials` should be requested.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,119 @@
...
+	Token(context.Context) (string, error)
+}
+
+// LazyToken is a struct that handles token acquisition and caching.
+type LazyToken struct {
+	url          string
+	clientId     string
+	clientSecret string
+	token        string
+	offlineToken string
```
lzap, line 28: Please remove this, this is essentially a `token`.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,119 @@
...
+// NewLazyToken creates a new LazyToken instance.
+func NewLazyToken(tokenURL, clientId, clientSecret string) *LazyToken {
+	return &LazyToken{
+		url:          tokenURL,
+		clientId:     clientId,
+		clientSecret: clientSecret,
+		token:        "",
+		expiration:   time.Now(),
+	}
+}
```
lzap, line 47: Might be useful to create also `NewXXX(url, id, existingToken)`

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,119 @@
...
+		logrus.WithContext(ctx).Infof("Acquired new token %s with expiration at: %s", lt.token, lt.expiration)
+	} else {
+		logrus.WithContext(ctx).Infof("Token requested: %s (expires at %s)", lt.token, lt.expiration)
+	}
+
+	return lt.token, nil
+}
+
+// AcquireToken is a placeholder function that simulates token acquisition.
+func acquireToken(tokenURL, clientID, clientSecret, offlineToken string) (*tokenResponse, error) {
```
lzap, line 72: Make this a method.

internal/oauth2/lazy_token_test.go
```diff
@@ -0,0 +1,106 @@
...
+	lazyToken.expiration = time.Now().Add(-time.Minute) // Simulate token expiration
+	token, err = lazyToken.Token(ctx)
+	if err != nil {
+		t.Errorf("Unexpected error acquiring token: %v", err)
+	}
+
+	if token != "mock-token" {
+		t.Errorf("Unexpected token, got %s, want %s", token, "mock-token")
+	}
+}
```
lzap, line 106: Add tests for edge cases - expired token specifically is important.

cmd/image-builder/main.go
```diff
@@ -88,18 +88,12 @@ func main() {
...
 	}
-
 	recommendationConf := recommendations.RecommendationsClientConfig{
-		URL:          conf.RecommendURL,
-		CA:           conf.RecommendCA,
-		TokenURL:     conf.RecommendTokenURL,
-		ClientId:     conf.RecommendClientId,
-		ClientSecret: conf.RecommendSecret,
+		URL: conf.RecommendURL,
+		CA:  conf.RecommendCA,
```
lzap, line 96: Please keep the configuration unchanged.

internal/config/config.go
```diff
@@ -1,6 +1,8 @@
 package config
 
-import "strings"
+import (
+	"strings"
+)
```
lzap, line 5: Please keep the configuration unchanged.

internal/clients/recommendations/client.go
```diff
@@ -51,11 +40,9 @@ func NewClient(conf RecommendationsClientConfig) (*RecommendationsClient, error)
 	}
```
lzap, line 41: In the `NewClient` use the existing *unchanged* configuration to call `NewLazyTokener` to create a new instance. Depending on `offlineToken` either call one or the other `New` function.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,119 @@
...
+// LazyToken is a struct that handles token acquisition and caching.
+type LazyToken struct {
+	url          string
+	clientId     string
+	clientSecret string
+	token        string
+	offlineToken string
+	expiration   time.Time
+	mutex        sync.Mutex
+}
```
lzap, line 31: Alternatively, you can export all fields, document them and remove the `NewXXX`.

lzap (changes requested):
Please do changes only for the recommendations client (drop composer for now) and once everyone is happy with the patch, let’s do all the other clients and tests as well.

internal/oauth2/lazy_token.go
mgold1234: why the time.Now() is not good enough?

internal/clients/recommendations/client.go
mgold1234: not sure I understand what youy ask here, in othyer commit you asked to remove the newLazyToken, so I cant use it here

internal/oauth2/lazy_token.go
lzap: I think we agreed on actually removing `New` method after all. Struct can be initialized normally, expiration time will default to year 0001 which is in the past (expired) so that is good.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,111 @@
...
+// LazyToken is a struct that handles token acquisition and caching.
+type LazyToken struct {
+	// Url represents the URL used for acquiring the token.
+	Url string
+	// ClientId is the client ID used for authentication.
+	ClientId string
+	// ClientSecret is the client secret used for authentication.
+	ClientSecret string
+	// token holds the currently cached token.
+	token string
```
lzap, line 31: Please export the `Token` as well, you can use it as pre-configured (offline) token if needed.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,111 @@
...
+	// Expiration stores the expiration time of the current token.
+	Expiration time.Time
+	// Mutex ensures safe concurrent access to the token and expiration fields.
+	Mutex sync.Mutex
```
lzap, line 35: This mutex must not be exported tho.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,111 @@
...
+		logrus.WithContext(ctx).Infof("Acquired new token %s with expiration at: %s", lt.token, lt.Expiration)
+	} else {
+		logrus.WithContext(ctx).Infof("Token requested: %s (expires at %s)", lt.token, lt.Expiration)
+	}
+
+	return lt.token, nil
+}
+
+// RequestToken is a placeholder function that simulates token acquisition.
+func (lt *LazyToken) RequestToken() (*tokenResponse, error) {
```
lzap, line 65: This should be unexported as well, I see no use case of calling this explicitly.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,111 @@
...
+	data := url.Values{}
+
+	if time.Now().After(lt.Expiration) {
+		data.Set("grant_type", "refresh_token")
+		data.Set("client_id", lt.ClientId)
+	} else if lt.ClientSecret != "" {
+		data.Set("grant_type", "client_credentials")
+		data.Set("client_id", lt.ClientId)
+		data.Set("client_secret", lt.ClientSecret)
+	}
```
lzap, line 80: This does not appear to be correct:

* If client id is not empty string AND (token is not empty string or token has expired) then do "refresh_token" call
* Else if client id and secret are both not empty string, do "client_credentials" calls.
* Else return error that client id, secret or token must be set.

lzap (changes requested):
Please improve the token type first, almost there.

ondrejbudai (commented):
I'm slightly hesitant about dropping the "retry-on-401" mechanism. Relying on the fact that a refresh token will not expire prematurely is in my opinion not defensive enough. It seems like a common good practice is to always expect that a refresh token might suddenly, see e.g. [Google](https://developers.google.com/identity/protocols/oauth2#expiration) and [oauth.com](https://www.oauth.com/oauth2-servers/making-authenticated-requests/refreshing-an-access-token/) (disclaimer: despite the domain name, this is not an official oauth docs).

Note that keeping "retry-on-401" here would be also nice for the sake of consistency - all our other projects (composer, koji-osbuild) also use this mechanism.

An easy way out might be to introduce a new `ForceRefresh()` method to the `Tokener` interface, that triggers a refresh regardless of expiration. Then, bring back the check for `401`, call `ForceRefresh()` if there's 401 and retry.

I was also thinking to abstract away the token logic into an [`http.RoundTripper`](https://pkg.go.dev/net/http#RoundTripper) (see e.g. [here](https://github.com/golang/oauth2/blob/84cb9f7f5c5a639955cd501bfdd54f0e63997e61/transport.go#L32)), and then use [`retryablehttp`](https://pkg.go.dev/github.com/hashicorp/go-retryablehttp) on top of this round tripper to provide the retrying mechanism. The downside is that everything gets more complex, the upside is that you get more resiliency against random networking/server issues. Note that we already use retryablehttp for some calls in the composer's worker.

internal/oauth2/duumy_token.go
```diff
@@ -0,0 +1,11 @@
+package oauth2
```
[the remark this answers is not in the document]
mgold1234: good catch

lzap:
I haven’t considered preliminary expiration or invalidation, good point. Let’s keep the 401-retry then, it still makes sense to extract the same code into a `Tokener` type. A method for explicit invalidation is a good solution, so if you like that @mgold1234 go for it.

The RoundTripper solution indeed looks like a great idea, tho, I think this might be quite a challenge. I am happy to assist if you choose to pursue that @mgold1234.

internal/oauth2/lazy_token.go
mgold1234: I initialized it at the client itslef -
rc := RecommendationsClient{
		URL: conf.URL,
		Tokener: &oauth2.LazyToken{
			Url:          "",
			ClientId:     "",
			ClientSecret: "",
			Token:        "",
			Expiration:   time.Time{},
		},
		client: client,
	}
should I do it in the lazy_token?

ondrejbudai (commented):
Getting nicer and nicer, thanks! :)

What's slightly concerning is that I'm pretty sure that the current implementation is buggy (the secrets are not passed correctly, see the inline comments), but the tests are green. I think we would really benefit from more tests, probably on the `clients/recommendation` level. Something like:
- set up quick and dirty oauth and recommendation server mocks
- create a client (`recommendations.NewClient()`)
- call `Client.RecommendationsPackages()`
- verify that the oauth mock was called, and that the access token got correctly propagated to the recommendation server mock
- we can also do more verifications there, like that the request matches the expected format for the recommendations API, but I do acknowledge that this is slightly out of scope for this PR as it focuses mainly on the oAuth flow

internal/clients/recommendations/client.go
```diff
@@ -51,11 +41,15 @@ func NewClient(conf RecommendationsClientConfig) (*RecommendationsClient, error)
...
 
 	rc := RecommendationsClient{
-		URL:          conf.URL,
-		tokenURL:     conf.TokenURL,
-		clientId:     conf.ClientId,
-		clientSecret: conf.ClientSecret,
-		client:       client,
+		URL: conf.URL,
+		Tokener: &oauth2.LazyToken{
```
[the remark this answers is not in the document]
mgold1234: amm yes, but where I can initalize the Expiration if I change it to conf.Tokener?

cmd/image-builder/main.go
```diff
@@ -95,11 +94,8 @@ func main() {
 	}
 
 	recommendationConf := recommendations.RecommendationsClientConfig{
```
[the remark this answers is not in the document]
mgold1234: great. and I will initialize the Expiration here as well

ondrejbudai (commented):
Awesome, we are getting there! :)

How do you feel about implementing a tiny smoke test for the whole recommendation client (see my previous comment)? We can take a look at that together if you want.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,105 @@
...
+	lt.AccessToken = ""
+	return lt.NextToken(ctx)
+}
+
+// requestToken is a function that responsible for fetching a new access token.
+func (lt *LazyToken) requestToken() (*tokenResponse, error) {
+
+	data := url.Values{}
+
+	if lt.AccessToken != "" {
```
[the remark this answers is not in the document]
mgold1234: not sure if I understand the change that I need to do
wdyt about this one if lt.AccessToken != "" {
		data.Set("grant_type", "refresh_token")
		data.Set("client_id", lt.ClientId)
		data.Set("refresh_token", lt.AccessToken)
	} else if lt.ClientSecret != "" {
		data.Set("grant_type", "client_credentials")
		data.Set("client_id", lt.ClientId)
		data.Set("client_secret", lt.ClientSecret)
	} else {

		return nil, fmt.Errorf("client Id, client Secret and token must be set")
	}

internal/oauth2/lazy_token_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestLazyToken(t *testing.T) {
+	nextTokenID := 0
+	mockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
croissanne, line 17: Could we check here that the request being sent is correct? That it has the correct grant_type, client_id, and client_secret set? If this is already happening somewhere else in a test that's ok too, I just think it's worth checking that the request the tokener sends is what we expect.

cmd/image-builder/main.go
```diff
@@ -95,11 +97,14 @@ func main() {
...
-		ClientId:     conf.RecommendClientId,
-		ClientSecret: conf.RecommendSecret,
+		URL: conf.RecommendURL,
+		CA:  conf.RecommendCA,
+		Tokener: &oauth2.LazyToken{
+			Url:          conf.RecommendTokenURL,
+			ClientId:     conf.RecommendClientId,
+			ClientSecret: conf.RecommendSecret,
+			Expiration:   time.Time{},
+		},
```
lzap, line 107: When you do not set Expiration then it is year 0000 time 0:00, therefore expired. No need to set it to current time.

internal/clients/recommendations/client.go
```diff
@@ -98,24 +85,23 @@ func (rc *RecommendationsClient) request(method, url string, headers map[string]
...
-	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token()))
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
 
 	resp, err := rc.client.Do(req)
 	if err != nil {
 		return nil, err
 	}
 	if resp.StatusCode == http.StatusForbidden || resp.StatusCode == http.StatusUnauthorized {
-		err = rc.refreshToken()
+		token, err = rc.tokener.ForceRefresh(req.Context())
```
lzap, line 99: I somehow do not like the pair `NextToken` and `ForceRefresh`. In the fist case, the call does not always return "next token", in fact it only does this once and then cache it. In the second case, I do not like that it does not contain `Token` in the method name.

Personal preference, but I would like something like `Token` and `RefreshToken` a bit more.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,105 @@
...
+	// Expiration stores the expiration time of the current token.
+	Expiration time.Time
+	// mutex ensures safe concurrent access to the token and expiration fields.
+	mutex sync.Mutex
+}
+
+type tokenResponse struct {
+	AccessToken string    `json:"access_token"`
+	Expiration  time.Time `json:"expiration"`
+}
```
lzap, line 42: I think you are missing refresh token from here. AFAIK access token cannot be used to refresh token.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,105 @@
...
+		tokenRes, err := lt.requestToken()
+		if err != nil {
+			return "", err
+		}
+
+		lt.AccessToken = tokenRes.AccessToken
+		lt.Expiration = tokenRes.Expiration
+		logrus.WithContext(ctx).Infof("Acquired new token %s with expiration at: %s", lt.AccessToken, lt.Expiration)
+	} else {
+		logrus.WithContext(ctx).Infof("AccessToken requested: %s (expires at %s)", lt.AccessToken, lt.Expiration)
```
lzap, line 60:
```suggestion
		logrus.WithContext(ctx).Infof("AccessToken reused: %s (expires at %s)", lt.AccessToken, lt.Expiration)
```

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	return lt.AccessToken, nil
+}
+
+func (lt *LazyToken) ForceRefresh(ctx context.Context) (string, error) {
+	lt.AccessToken = ""
```
lzap, line 67: This is a change which is not protected by the `mutex`, generally this is a bad practice but in this case if I am not mistaken it is not an issue since blank access token means it is gonna be refreshed on next update.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if lt.AccessToken == "" || time.Now().After(lt.Expiration.Add(-time.Minute)) {
+		// Acquire a new token and set the expiration time
+		tokenRes, err := lt.requestToken()
+		if err != nil {
+			return "", err
+		}
+
+		lt.AccessToken = tokenRes.AccessToken
+		lt.Expiration = tokenRes.Expiration
```
lzap, line 57: Here I would expect `RefreshToken` to be set too.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// requestToken is a function that responsible for fetching a new access token.
+func (lt *LazyToken) requestToken() (*tokenResponse, error) {
+
+	data := url.Values{}
+	data.Set("client_id", lt.ClientId)
+	if lt.RefreshToken != "" {
+		data.Set("refresh_token", lt.RefreshToken)
+		data.Set("grant_type", "refresh_token")
```
lzap, line 78: Unreachable code. This code appears to use `RefreshToken` which is never set.

internal/oauth2/lazy_token.go
mgold1234: yes, it will refresh on next update, I think it is what we want here

internal/oauth2/lazy_token.go
ondrejbudai: But that doesn't match what the current code does... The current code never updates the refresh token...

internal/oauth2/lazy_token.go
ondrejbudai: I think this is actually racy. Imagine `ForceRefresh` being called before this return statement: https://github.com/mgold1234/image-builder/blob/20d8326200dd50a030037bd660bb21c7589aa9d2/internal/oauth2/lazy_token.go#L65

So I guess we have to extract

```
if lt.AccessToken == "" || time.Now().Add(time.Minute).After(lt.Expiration) {
		// Acquire a new token and set the expiration time
		tokenRes, err := lt.requestToken()
		if err != nil {
			return "", err
		}

		lt.AccessToken = tokenRes.AccessToken
		lt.Expiration = tokenRes.Expiration
		lt.RefreshToken = tokenRes.RefreshToken
		logrus.WithContext(ctx).Infof("Acquired new token %s with expiration at: %s", lt.AccessToken, lt.Expiration)
	} else {
		logrus.WithContext(ctx).Infof("AccessToken reused: %s (expires at %s)", lt.AccessToken, lt.Expiration)
	}

	return lt.AccessToken, nil
```

to a separate private function, and protect both `RefreshToken` and `ForceRefresh` with the same mutex, and let them call the private function.

ondrejbudai (commented):
We are getting there I think, thanks! Just a handful of inline comments. Thanks for doing this.

internal/oauth2/dummy_token.go
```diff
@@ -0,0 +1,16 @@
...
+
+// NextToken returns a static "testtoken" string.
+func (dt *DummyToken) NextToken(ctx context.Context) (string, error) {
+
+	return "testtoken", nil
+}
+
+func (dt *DummyToken) ForceRefresh(ctx context.Context) (string, error) {
+	return "", nil
+}
```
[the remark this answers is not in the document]
ondrejbudai: Right, yeah, you are correct. I got lost here, because other tests in this file have very little value: they all test just the mock, they test basically no production code paths:

https://github.com/osbuild/image-builder/blob/b656bd58831787f84b0946ecd5b660d0ba93d888/internal/v1/handler_recommend_package_test.go

See `espStatusCode, body := tutils.PostResponseBody(t, apiSrv.URL, payload)`. This just test the mock defined a few lines above. If you remove `startServer` from these tests, the tests will be absolutely happy.

What shall we do about this? I suggest merging this PR "as is" (see my latest review comments), and then reverting to use `DummyToken` in tests, and also fixing the whole `handler_recommend_package_test`. (The newly added test is actually doing what's supposed to do.)

Btw, we should also add a test for the fallback behaviour to `ForceRefresh`. :)

internal/oauth2/lazy_token.go
lzap: Ah right my guts were right, I missed the window between setting the token and returning the statement. Yeah, good point, this needs to be extracted.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,107 @@
...
+	RefreshToken string
+	// Expiration stores the expiration time of the current token.
+	Expiration time.Time
+	// mutex ensures safe concurrent access to the token and expiration fields.
+	mutex sync.Mutex
+}
+
+type tokenResponse struct {
+	AccessToken  string    `json:"access_token"`
+	Expiration   time.Time `json:"expiration"`
```
lzap, line 41: I noticed one thing, this is not how the response actually look like from RH OIDC. It actually looks like this:

```
{
  "access_token": "xxxxxxxx",
  "expires_in": 900,
  "refresh_expires_in": 0,
  "token_type": "Bearer",
  "not-before-policy": 0,
  "scope": "api.console"
}
```

You need to provide a scope I think to [get refresh token](https://auth0.com/docs/secure/tokens/refresh-tokens/get-refresh-tokens), I haven’t tried this myself. Note the fields has also different names (expires_in vs expiration).

lzap: Ok so I did[ read more about oauth2 token refreshing](https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/) and it appears that refresh token is actually optional, our service does not appear to return it. In that case, we can stick with the initial goal - when access token expires, just request a new one with id/secret. That should not be any problem, this is what IB does for every single request after all.

So in this case, I believe you only need to fix the expiration. The field name is named differently (`expires_in`) and it is not time but amount of seconds from now so you need to convert this or update the code to use seconds instead of time.

internal/oauth2/lazy_token.go
lzap: Yeah I realized that Keycloak instance of RH SSO is not configured to return refresh tokens at all, so disregard this. Let’s just acquire new token when it expires.

lzap (changes requested):
The code still expects `expiration` field from the oauth2 service while the service returns `expires_at`. The data also do not match, it returns integer (seconds from now) not parseable date time.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,139 @@
...
+type LazyToken struct {
+	// Url represents the URL used for acquiring the token.
+	Url string
+	// ClientId is the client ID used for authentication.
+	ClientId string
+	// ClientSecret is the client secret used for authentication.
+	ClientSecret string
+	// 	AccessToken string holds the currently cached token.
+	AccessToken string
+	// 	RefreshToken string token after refresh.
```
lzap, line 31: Formatting.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,139 @@
...
+	// Expiration stores the expiration time of the current token.
+	Expiration time.Time
+	// RefreshExpiration stores the expiration time of the refresh token.
+	RefreshExpiration time.Time
+	// TokenType is the type of the access token (e.g., "Bearer").
+	TokenType string
+	// NotBeforePolicy is the not-before policy of the access token.
+	NotBeforePolicy int
+	// Scope is the scope granted for the access token.
+	Scope string
```
lzap, line 42: Just for the record, the three fields (token type, not before policy and scope) are not used. I am not against keeping this in memory, tho, there is no reason to keep it.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,139 @@
...
+	// mutex ensures safe concurrent access to the token and expiration fields.
+	mutex sync.Mutex
```
lzap, line 44: Well this does cover all fields actually if you want to be correct.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+type tokenResponse struct {
+	AccessToken      string `json:"access_token"`
+	ExpiresIn        int    `json:"expires_in"`
+	RefreshExpiresIn int    `json:"refresh_expires_in"`
```
lzap, line 50: Did you manage to get a refresh token out of the stage/prod services? I could not.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,139 @@
...
+	TokenType        string `json:"token_type"`
+	NotBeforePolicy  int    `json:"not-before-policy"`
+	Scope            string `json:"scope"`
+}
+
+func (tr *tokenResponse) expiration() time.Time {
+	return time.Now().Add(time.Duration(tr.ExpiresIn) * time.Second)
+}
```
lzap, line 58: This function is only used once, one-line functions which without code reuse are barely useful.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,139 @@
...
+		err := lt.acquireNewToken(ctx)
+		if err != nil {
+			return "", err
+		}
+	} else {
+		logrus.WithContext(ctx).Infof("AccessToken reused: %s (expires at %s)", lt.AccessToken, lt.Expiration)
+	}
+
+	return lt.AccessToken, nil
+}
```
lzap, line 93: As was suggested, you need to create an unexported function that does all the code logic of `Token()` and also it can perform the reset too. You can use a function argument to pass reset/refresh flag so it is done in the protected block and use this from both `Token` and `ForceRefresh` functions.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,139 @@
...
+
+// ForceRefresh is a function that responsible for fetching a new access token.
+func (lt *LazyToken) ForceRefresh(ctx context.Context) (string, error) {
+	lt.AccessToken = ""
```
lzap, line 97: This line needs to be done in the unexported function in the protected code area by the mutex `mutex`.

lzap (commented):
Almost there, my biggest concern is concurrency of the `ForceRefresh` method.

internal/oauth2/lazy_token.go
mgold1234: we can remove the RefreshExpiresIn, as you mentioned 'it appears that refresh token is actually optional, our service does not appear to return it. In that case, we can stick with the initial goal - when access token expires, just request a new one with id/secret.' 
no?

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,115 @@
...
+	return lt.AccessToken, nil
+}
+
+func (lt *LazyToken) Token(ctx context.Context) (string, error) {
+	return lt.acquireNewToken(ctx)
+}
+
+// ForceRefresh is a function that responsible for fetching a new access token.
+func (lt *LazyToken) ForceRefresh(ctx context.Context) (string, error) {
+	lt.AccessToken = ""
```
lzap, line 73: This must be protected by the mutex.

lzap (changes requested):
The code is still not thread-safe. You have to protect all access via the mutex.

Since our service does not support refresh tokens, I suggest to remove all the refresh token related code from the codebase. The less code the better, this is a security related part after all.

internal/oauth2/lazy_token.go
```diff
@@ -0,0 +1,107 @@
...
+
+// ForceRefresh is a function that responsible for fetching a new access token.
+func (lt *LazyToken) ForceRefresh(ctx context.Context) (string, error) {
+	return lt.acquireNewToken(ctx, true)
+}
+
+// ForceRefresh forces the acquisition of a new access token by clearing the current AccessToken and calling Token().
+func (lt *LazyToken) requestToken() (*tokenResponse, error) {
+
+	data := url.Values{}
```
lzap, line 75: A leftover empty line?

lzap (commented):
Looks good, I am leaving the final call on Ondra or Sanne. Thanks.

ondrejbudai (approved):
Looks good, thanks a lot for going through this! I think we just need one final rebase that removes the merge commit from the history. Then, it's good to do! :heart_eyes:

ondrejbudai (approved):

PR #1136, opened 2024-04-15
build: Do not create docker context in workdirectory

ezr-ondrej:


croissanne (approved):
Could you make the same change in composer?

ezr-ondrej:
> Could you make the same change in composer?

Does it make sense there? :thinking: I did not find the build script, I'm bit in the dark how composer is deployed.

PR #1139, opened 2024-04-16
pr_check: Deploy unleash-proxy to enable feature flags

jrusz:
This enabled features flags to be checked in ephemeral.


mgold1234 (dismissed):
/lgtm

croissanne (approved):

ezr-ondrej:
I'm not against this, but recently I attended a call where someone said the proxy should be only used by stuff outside of the cluster :)
Tho I get it's the easiest way to get to the feature flags :)

PR #1141, opened 2024-04-17
v1: explicitly add rhcd for openscap + subscription

kingsleyzissou:
If both openscap and subscription customizations are set, we should explicitly add `rhcd` to the list of enabled services. For some reason the service gets disabled during the openscap remediation and is then not running when the system has started.

### DIFF
--- a/internal/v1/handler_compose_image.go
+++ b/internal/v1/handler_compose_image.go
@@ -840,7 +840,18 @@ func buildCustomizations(cust *Customizations) (*composer.Customizations, error)
 		if res.Services.Enabled == nil {
 			res.Services.Enabled = &[]string{}
 		}
-		*res.Services.Enabled = append(*res.Services.Enabled, "rhcd")
+		// TODO: switch to `slice.Contains` when moving to go 1.21
+		containsService := func(services []string, service string) bool {
+			for _, s := range services {
+				if s == service {
+					return true
+				}
+			}
+			return false
+		}
+		if !containsService(*res.Services.Enabled, "rhcd") {
+			*res.Services.Enabled = append(*res.Services.Enabled, "rhcd")
+		}
 	}
 
 	if cust.Firewall != nil {
--- a/internal/v1/handler_compose_image.go
+++ b/internal/v1/handler_compose_image.go
@@ -830,6 +830,19 @@ func buildCustomizations(cust *Customizations) (*composer.Customizations, error)
 		}
 	}
 
+	// we need to explicitly add 'rhcd' to the enabled services
+	// if openscap and subscription customizations are set, otherwise
+	// the insights-client doesn't register properly
+	if cust.Subscription != nil && cust.Subscription.Insights && cust.Openscap != nil {
+		if res.Services == nil {
+			res.Services = &composer.Services{}
+		}
+		if res.Services.Enabled == nil {
+			res.Services.Enabled = &[]string{}
+		}
+		*res.Services.Enabled = append(*res.Services.Enabled, "rhcd")
+	}
+
 	if cust.Firewall != nil {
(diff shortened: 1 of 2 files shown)


internal/v1/handler_compose_image.go
```diff
@@ -830,6 +830,24 @@ func buildCustomizations(cust *Customizations) (*composer.Customizations, error)
 		}
 	}
 
+	// we need to explicitly add 'rhcd' to the enabled services
+	// if openscap and subscription customizations are set, otherwise
+	// the insights-client doesn't register properly
+	if cust.Subscription != nil && cust.Subscription.Insights && cust.Openscap != nil {
```
kingsleyzissou, line 836: Should I check also for `cust.Subscriptions.Rhc` too?

internal/v1/handler_compose_image.go
```diff
@@ -830,6 +830,24 @@ func buildCustomizations(cust *Customizations) (*composer.Customizations, error)
...
+		if res.Services == nil {
```
kingsleyzissou, line 837: This is a bit ugly. But it's kind of hard to check for all the cases because we're checking `OpenSCAP`, `Services` and `Subscriptions` customizations

internal/v1/handler_compose_image.go
```diff
@@ -830,6 +830,24 @@ func buildCustomizations(cust *Customizations) (*composer.Customizations, error)
...
+				Enabled: &[]string{"rhcd"},
+			}
+		} else {
+			enabledServices := []string{}
+			if res.Services.Enabled != nil {
+				enabledServices = *res.Services.Enabled
+			}
+			enabledServices = append(enabledServices, "rhcd")
+			res.Services.Enabled = &enabledServices
+		}
```
ezr-ondrej, line 848: Nitpick: I think we can simplify by just filling the blanks (not sure if the referencing works on append tho):
```suggestion
		if res.Services == nil {
			res.Services = &composer.Services{}
		} 
		if res.Services.Enabled == nil {
			res.Services.Enabled = &[]string{}
		}
		res.Services.Enabled = &append(*res.Services.Enabled, "rhcd")
```

ezr-ondrej: Tho we are not checking if it's actually not there yet, before adding. Should we? :thinking:

kingsleyzissou: I mean we could. But I am almost 99% sure that none of the profiles set it... I will double check though

kingsleyzissou: `grep -r -c "sshd" ./distributions/*/oscap`
Produces no results. So I think we're fine

internal/v1/handler_compose_image.go
```diff
@@ -830,6 +830,29 @@ func buildCustomizations(cust *Customizations) (*composer.Customizations, error)
...
+			res.Services.Enabled = &[]string{}
+		}
+		containsService := func(services []string, service string) bool {
+			for _, s := range services {
+				if s == service {
+					return true
+				}
+			}
+			return false
+		}
```
ezr-ondrej, line 850: Might make sense to add comment `TODO: use slices.Contains(services, "rhcd") when migrating to go 1.21`

ezr-ondrej (approved):
Looks good :+1:

internal/v1/handler_compose_image.go
kingsleyzissou: I half considered it. I think we should add it now. I'll quickly do that and bug you for a review

ezr-ondrej (approved):
I think it's good, I can't say if it's gonna fix the issue, but in that I trust it does.
I can't merge without the pr_check passing tho :)

PR #1143, opened 2024-04-18
v1: switch snapshot_date format to extended ISO 8601

croissanne:
---

Request from @andrewgdewar, it makes sense, the time part was ignored anyway.


Andrewgdewar (approved):
Much format! 
Very approve!
Ack!

PR #1155, opened 2024-04-29
handler_compose_image: Accept empty string in SnapshotDate

schuellerf:
In my local setup the frontend sends an empty string …instead of nil

Should we fix this here or in the frontend or did I mix something up and it's not needed at all?


croissanne (changes requested):
This is a bug in the frontend imo, sending an empty date string would not be a valid input.

schuellerf:
Thanks for the info 🤩

PR #1156, opened 2024-04-29
workflows: automate OpenSCAP config generation.

kingsleyzissou:
Add a workflow dispatcher for generating OpenSCAP configs. Also, create a schedule for the configs
to be checked once a week. If there are any changes create a pull request for this.

### DIFF
--- a/.github/workflows/openscap.yml
+++ b/.github/workflows/openscap.yml
@@ -1,0 +1,51 @@
+# This action checks OpenSCAP configs every Monday at 5:00 UTC.
+name: openscap
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 5 * * 1"
+
+jobs:
+  update-openscap:
+    name: "Update OpenSCAP configs"
+    if: github.repository == 'osbuild/image-builder'
+    runs-on: ubuntu-latest
+    container:
+      image: registry.fedoraproject.org/fedora:39
+
+    steps:
+      - name: Set up bootstrap Go 1.20
+        uses: actions/setup-go@v5.0.0
+        with:
+          go-version: '1.20'
+          cache: false
+        id: go
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


.github/workflows/openscap.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v6
+        with:
+          branch: update-openscap
+          delete-branch: true
+          title: "distributions: regenerate OpenSCAP configs"
+          commit-message: "distributions: regenerate OpenSCAP configs"
+          body: Update OpenSCAP configs
+          token: ${{ secrets.###TODO### }}
```
kingsleyzissou, line 50: Would just need help setting this up

ochosi, line 50:
```suggestion
          token: ${{ secrets.SCHUTZBOT_GITHUB_ACCESS_TOKEN }}
```

I have shared this org-secret with this repo now, so you should be able to use it here.

ochosi (commented):
Yay, thanks for looking into this! <3
Have you tested this somehow? In any case, the code does look very plausible :)
(Just to be clear, I have not.)

The secret I suggested should work for what you're trying to do, but I guess we can only find out once this is merged (unless you create the PR from a branch in the original repository and not your fork, which otherwise prevents secrets from being used etc.)

kingsleyzissou:
> Have you tested this somehow?

I've tested as much as I can locally. The checkout and go stuff doesn't work great locally inside containers.
And using `OpenSCAP` directly on ubuntu is challenging. So I might have to tweak this a bit down the line.

.github/workflows/openscap.yml
kingsleyzissou: Amazing. Thanks

kingsleyzissou:
Whoops. I changed the base branch to `oscap-automation` and was going to create a new PR. In doing that, this PR got automatically merged to the new `oscap-automation` branch. The new PR is: https://github.com/osbuild/image-builder/pull/1160

PR #1159, opened 2024-04-30
chore: do not break tests on codecov.io quota

lzap:
SSIA


ezr-ondrej (approved):
:orange_heart:

croissanne (changes requested):
This doesn't fix the issue, which seems to be https://github.com/codecov/feedback/issues/354

Also fwiw we don't use `chore` in the topic, just use `.github/workflows/tests`. See https://osbuild.org/docs/developer-guide/general/workflow#preferred-commit-message-format:
>The <module> should point to the area of the codebase (for instance tests or tools). The topic should summarize what the commit is doing.

lzap:
Yeah sorry

PR #1162, opened 2024-04-30
 v1: fix the response data from the recommendation service

mgold1234:
Unmarshal the response into an appropriate struct.

### DIFF
--- a/internal/v1/handler_recommend_package.go
+++ b/internal/v1/handler_recommend_package.go
@@ -39,16 +39,17 @@ func (h *Handlers) handleRecommendationsResponse(ctx echo.Context, uploadPackage
 
 	ctx.Logger().Debugf("Getting Response list of items %v", resp)
 	defer closeBody(ctx, resp.Body)
-	var packages []string
-	err = json.NewDecoder(resp.Body).Decode(&packages)
+
+	var responsePackages RecommendationsResponse
+	err = json.NewDecoder(resp.Body).Decode(&responsePackages)
 	if err != nil {
 		return RecommendationsResponse{}, err
 	}
 
-	if len(packages) == 0 {
+	if len(responsePackages.Packages) == 0 {
 		ctx.Logger().Errorf("User should define packages")
 		return RecommendationsResponse{}, nil
 	}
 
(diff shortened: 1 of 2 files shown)


internal/v1/handler_recommend_package.go
```diff
@@ -39,16 +39,19 @@ func (h *Handlers) handleRecommendationsResponse(ctx echo.Context, uploadPackage
 
 	ctx.Logger().Debugf("Getting Response list of items %v", resp)
 	defer closeBody(ctx, resp.Body)
-	var packages []string
-	err = json.NewDecoder(resp.Body).Decode(&packages)
+	type ResponsePackages struct {
```
croissanne, line 42: No need to define a new struct, just use ```RecommendationsResponse```

internal/v1/handler_recommend_package.go
```diff
@@ -39,16 +39,19 @@ func (h *Handlers) handleRecommendationsResponse(ctx echo.Context, uploadPackage
...
 	}
 
-	if len(packages) == 0 {
+	if len(responsePackages.Packages) == 0 {
 		ctx.Logger().Errorf("User should define packages")
 		return RecommendationsResponse{}, nil
 	}
 
-	return RecommendationsResponse{Packages: packages}, nil
+	return RecommendationsResponse{Packages: responsePackages.Packages}, nil
```
croissanne, line 56: Just return the RecommendationsResponse into which you unmarshal the response.

croissanne (changes requested):

croissanne (changes requested):
Thanks! Could you also:
- squash the commits
- fix the typo in the commit title
- max length of the commit title is 65
- set the module to something like `internal/v1`
- construct a RecommendationsResponse only once instead of twice.

croissanne (changes requested):
Can you reword the commit message?
Suggestion:
```
 v1: fix the response data from the recommendation service

Unmarshal the response into an appropriate struct.
```

croissanne (approved):

lzap (approved):
LGTM

PR #1163, opened 2024-04-30
.github/workflows/tests: remove codecov

croissanne:
The script we were using before no longer works, and the results weren't posted for a while now. We should use the v4 action if we want to keep using codecov.


mgold1234 (approved):
/lgtm

PR #1165, opened 2024-05-01
v1: Fixed recommendation response schema

mgold1234:
This commit removes the array from the recommendation response, resolving the issue of missing data displayed on the frontend.


regexowl (approved):
Looks great, thank you!

PR #1166, opened 2024-05-01
workflows: fix openscap deps

kingsleyzissou:
Fix the automation since it was failing as some dependencies weren’t installed.


mgold1234 (approved):
/lgtm

PR #1171, opened 2024-05-03
echo_logrus: implement adding fields

schuellerf:
Implements fields in addition to the context
for better errors context in tools like glitchtip

Please comment on
 * what are your thoughts on using fields or context
 * should we try to generalize errors (not including dynamic things like path, method,… as "text")

When using this I think tools like glitchtip can be more useful, although readability in the plain logs might suffer:

before
```
time="2024-05-02T15:19:39Z" level=debug msg="Started request GET /api/image-builder/v1/composes/fc4b8a3d-3ded-4a50-847d-5d1a5daa4483" func=main.requestIdExtractMiddleware.func1 file="/go/src/github.com/osbuild/image-builder/cmd/image-builder/logging.go:76" insights_id=FITVV1Rw5Mk0 request_id=qgyuWGSPHGge
```

after
```
time="2024-05-03T10:10:51Z" level=debug msg="Started request" func="github.com/osbuild/image-builder/internal/common.(*EchoLogrusLogger).Debugf" file="/go/src/github.com/osbuild/image-builder/internal/common/echo_logrus.go:91" composeId=28fcdbc8-79a6-4b3f-ab2b-7f1f273c1d4b insights_id=95ikbCtxXtrd method=GET path="/api/image-builder/v1/composes/:composeId" request_id=Wk9VXGSfkgGR
```

### DIFF
--- a/cmd/image-builder/logging.go
+++ b/cmd/image-builder/logging.go
@@ -65,15 +65,21 @@ func requestIdExtractMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
 		ctx = context.WithValue(ctx, insightsRequestIdCtx, iid)
 		c.SetRequest(c.Request().WithContext(ctx))
 
+		f := logrus.Fields{"method": c.Request().Method, "path": c.Path()}
+		for _, key := range c.ParamNames() {
+			f[key] = c.Param(key)
+		}
+
 		// and set echo logger to be context logger
-		ctxLogger := logrus.StandardLogger().WithContext(ctx).Logger
+		ctxLogger := logrus.StandardLogger()
 		c.SetLogger(&common.EchoLogrusLogger{
 			Logger: ctxLogger,
 			Ctx:    ctx,
+			Fields: f,
 		})
 
 		if !SkipPath(c.Path()) {
-			ctxLogger.WithContext(ctx).Debugf("Started request %s %s", c.Request().Method, c.Request().RequestURI)
+			c.Logger().Debugf("Started request")
 		}
 
 		return next(c)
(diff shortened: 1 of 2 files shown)


croissanne (approved):

lzap:
Well, big advantage of using context instead of direct fields is when `echo.Logger` interface is not used. We already have a hook that extracts fields in `cmd/image-builder/logging.go`:

```
func (h *ctxHook) Fire(e *logrus.Entry) error {
	if e.Context != nil {
		e.Data["request_id"] = e.Context.Value(requestIdCtx)
		e.Data["insights_id"] = e.Context.Value(insightsRequestIdCtx)
	}

	return nil
}
```

Putting the required data into context and extracting it here would cover more cases and in the end could be smaller patch.

schuellerf:
@lzap ok, thanks I'll check and revise

schuellerf:
@lzap Supporting fields might still be useful as it feels like fields are more lightweight (in terms of runtime and footprint) than a wrapped context object. Please ping me on slack if you have time to talk/videocall about the exact differences of those two as I don't have too much experience with logrus in that aspect.
And things like "HTTP method" and "openapi path" are available to all echo-server calls but not logrus in that case... 🤔

lzap:
A new value context consumes [literally three pointers](https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/context/context.go;l=728) in memory per request.

I do not insist, this is fine. My suggestion was to simply wrap those new fields into context and then retrieve them in the logrus hook similarly we do it for request id. I can make a PR to see the difference if you want.

schuellerf:
So you would not be concerned about "polluting" the context with things that are already available in the echo.Request() and similar echo-objects?
For sure I can remove those fields again and add to the context as I also want to add another PR anyway to fix/workaround the fact that "file" is wrong (always `echo_logrus.go`). Hmm those should probably be two separate PRs. 🤔

lzap:
I created a PR, see the link. Let’s continue there.

PR #1174, opened 2024-05-06
echo_logrus: use std context for fields

lzap:
This is a followup of https://github.com/osbuild/image-builder/pull/1171/files

I have couple of concerns with the recent changes.

* The new fields are only present when echo logger is used. There are many other logs created through logrus interface (e.g. all SQL queries done by the pgx driver) which will miss all those fields.
* All path params are added and can possibly override existing fields e.g. msg or level. This could have unwanted effects, this new patch protects the most important fields. (I haven’t tested this chances are logrus would not allow overriding level or msg fields at all.)
* There were concerns about memory consumption or performance - `logrus.Fields` is a map and context is a lightweight object by design. Paths typically contain one or two fields maximum, with path and method this is up to 4 map elements.
* Context will carry a copy of information, that is a valid point (see the original PR discussion), but as described above this is very small amount of information.
* I noticed context was not added in "Finished request" message, it is now added.

I have to add I do not believe it is a good thing to add path, method and all path params to every single log line, I think it would be better to only log this once at the beginning of the request and devs can use correlation ids to find the line from any logged line. This proposal adds those fields everywhere tho.

### DIFF
--- a/cmd/image-builder/logging.go
+++ b/cmd/image-builder/logging.go
@@ -15,6 +15,7 @@ type ctxKey int
 const (
 	requestIdCtx         ctxKey = iota
 	insightsRequestIdCtx ctxKey = iota
+	requestDataCtx       ctxKey = iota
 )
 
 // Use request id from the standard context and add it to the message as a field.
@@ -36,6 +37,10 @@ func (h *ctxHook) Fire(e *logrus.Entry) error {
 	if e.Context != nil {
 		e.Data["request_id"] = e.Context.Value(requestIdCtx)
 		e.Data["insights_id"] = e.Context.Value(insightsRequestIdCtx)
+		rd := e.Context.Value(requestDataCtx).(logrus.Fields)
+		for k, v := range rd {
+			e.Data[k] = v
+		}
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


cmd/image-builder/logging.go
```diff
@@ -59,27 +64,34 @@ func requestIdExtractMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
...
 		ctxLogger := logrus.StandardLogger()
 		c.SetLogger(&common.EchoLogrusLogger{
 			Logger: ctxLogger,
 			Ctx:    ctx,
-			Fields: f,
 		})
 
 		if !SkipPath(c.Path()) {
-			c.Logger().Debugf("Started request")
+			ctxLogger.WithContext(ctx).WithFields(rd).Debugf("Started request")
```
lzap, line 94: This hunk is probably unnecessary, I was experimenting around to find the best solution.

cmd/image-builder/main.go
```diff
@@ -154,7 +154,8 @@ func main() {
 			if values.Error != nil {
 				fields["error"] = values.Error
 			}
-			logrus.WithFields(fields).Infof("Processed request %s %s", values.Method, values.URI)
+			logrus.WithContext(c.Request().Context()).
+				WithFields(fields).Infof("Processed request %s %s", values.Method, values.URI)
```
lzap, line 158: Here `WithContext` was missing, somewhat unrelated to the change but I think it is worth fixing.

internal/common/echo_logrus.go
```diff
@@ -68,113 +66,113 @@ func (l *EchoLogrusLogger) SetPrefix(p string) {
...
 }
 
 func (l *EchoLogrusLogger) Panicj(j log.JSON) {
 	b, err := json.Marshal(j)
 	if err != nil {
 		panic(err)
 	}
-	l.Logger.WithContext(l.Ctx).WithFields(l.Fields).Panicln(string(b))
+	l.Logger.WithContext(l.Ctx).Panicln(string(b))
 }
```
lzap, line 178: This file is a full revert. No other changes.

cmd/image-builder/logging.go
schuellerf:
```suggestion
			ctxLogger.WithContext(ctx).Debugf("Started request")
```
my `WithFields` is also not needed here, right?

internal/common/echo_logrus.go
schuellerf: Awesome thanks, looks nicer and even covers more use cases!

schuellerf (changes requested):
Very nice, thanks

cmd/image-builder/logging.go
lzap: It is either this, or I can make use of the echo logger created above. Let me use the echo logger instead, looks better.

lzap:
Rebased.

cmd/image-builder/main.go
schuellerf: do you think also adding
`"path":   c.Path(),`
to the fields here would be beneficial as this is better for log entry searches/grouping (in contrast to `values.URI`) I guess
but no stopper for this PR I guess

schuellerf (approved):
LGTM

lzap:
The way I search paths is simply doing normal search across all fields and once I get request-id I search for that to get the full history of a request. This will work because the path is logged at least once on every request. I can add this if you want.

PR #1176, opened 2024-05-07
Fix caller in logs

schuellerf:
@lzap based on your #1174 I'd like to add the fix to show the correct caller (for the cases using `echo_logrus.go`)
I'd keep this as draft until yours is merged

Do you think this makes sense?

example log message is (just added some newlines for readabililaty)
before:
```
backend-1            | time="2024-05-07T09:08:36Z" level=error msg="RHEL entitlement not present in identity header"
func="github.com/osbuild/image-builder/internal/common.(*EchoLogrusLogger).Error"
file="/go/src/github.com/osbuild/image-builder/internal/common/echo_logrus.go:153"
distribution=centos-9 insights_id=eIBJqJJqkaz7 method=GET path="/api/image-builder/v1/architectures/:distribution" request_id=FbaXzyS1epxn
```
after:
```
backend-1            | time="2024-05-07T09:12:52Z" level=error msg="RHEL entitlement not present in identity header"
func="github.com/osbuild/image-builder/internal/v1.(*Identity).IsEntitled"
file="/go/src/github.com/osbuild/image-builder/internal/v1/server_identity.go:76"
distribution=centos-9 insights_id=XMckxfNStVNf method=GET path="/api/image-builder/v1/architectures/:distribution" request_id=yIP6lO4O6wmG

```

### DIFF
--- a/internal/common/echo_logrus.go
+++ b/internal/common/echo_logrus.go
@@ -4,9 +4,16 @@ import (
 	"context"
 	"encoding/json"
 	"io"
+	"runtime"
 
 	"github.com/labstack/gommon/log"
 	"github.com/sirupsen/logrus"
+)
+
+type ctxKey int
+
+const (
+	LoggingFrameCtx ctxKey = iota
 )
 
 // EchoLogrusLogger extend logrus.Logger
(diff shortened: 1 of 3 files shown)


internal/common/echo_logrus.go
```diff
@@ -4,11 +4,18 @@ import (
...
 
 	"github.com/labstack/gommon/log"
 	"github.com/sirupsen/logrus"
 )
 
+type ctxKey int
+
+const (
+	LoggingFrameCtx ctxKey = iota
+)
```
lzap, line 17: It is a good practice not to export context key types and values so other packages cannot modify the values. Here it is a mix - you export the value but not the type.

lzap:
Yeah I noticed this as well. Tho the implementation seems fragile, creating a context value per request is fine, but per log entry? There can be many log entries per request, this could get out of control. Cannot you modify the caller on the fly directly? Or even better - turn off automatic caller adding and implement your own hook that will consider this wrapper instead.

Whatever you do, please make a unit test for this, do some logging directly through `logrus` as well as through the echo facade to ensure consistency.

internal/common/echo_logrus.go
schuellerf: So we should change this in `cmd/image-builder/logging.go` too? (where I got the pattern from)

schuellerf:
Note: related implementation: https://github.com/osbuild/osbuild-composer/pull/4132

PR #1180, opened 2024-05-09
v1: add /experimental/blueprints/{id}/imageTypes endpoint to image builder (HMS-4066)

mgold1234:
This commit introduces a new endpoint that retrieves a list of image types for a specific blueprint image. It also adds a handler function to implement this endpoint. Additionally, unit tests are included to verify the functionality of this new endpoint.

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -591,6 +591,39 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPErrorList'
+  /experimental/blueprints/{id}/image_types:
+    parameters:
+      - in: path
+        name: id
+        schema:
+          type: string
+          format: uuid
+        example: '123e4567-e89b-12d3-a456-426655440000'
+        required: true
+        description: UUID of a blueprint
+    get:
+      summary: Get image types from a blueprint
+      description: "get image types from a blueprint"
+      operationId: getImageTypesFromBlueprint
+      parameters:
+        - name: id
+          in: path
+          required: true
+          schema:
+            type: string
+            format: uuid
+      responses:
+        '200':
+          description: Successful response
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  data:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/ImageTypes'
   /experimental/blueprints/{id}/composes:
     get:
       summary: get composes associated with a blueprint
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -132,6 +132,31 @@ func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID) error {
 	return ctx.JSON(http.StatusOK, blueprintResponse)
 }
 
+func (h *Handlers) GetImageTypesFromBlueprint(ctx echo.Context, id openapi_types.UUID) error {
+	userID, err := h.server.getIdentity(ctx)
+	if err != nil {
+		return err
+	}
+
+	ctx.Logger().Infof("Fetching blueprint %s", id)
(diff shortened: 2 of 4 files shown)


internal/v1/api.yaml
```diff
@@ -591,6 +591,36 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPErrorList'
+  /experimental/blueprints/{id}/imageTypes:
```
ezr-ondrej, line 594: I belive we should use lowercase in API `image_types`

ezr-ondrej (commented):
I thought we'll use the `/blueprints/{uuid}` endpoint and extract it on the frontend, but if we think we need this, this looks correct :)

internal/v1/handler_blueprints.go
```diff
@@ -132,6 +132,40 @@ func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID) error {
...
+		Distribution:   blueprint.Distribution,
+		Customizations: blueprint.Customizations,
+	}
+
+	var imageTypes []ImageTypes
+	for _, req := range blueprintResponse.ImageRequests {
+		imageTypes = append(imageTypes, req.ImageType)
+	}
+
+	return ctx.JSON(http.StatusOK, imageTypes)
```
lzap, line 166: Why do you create new `BlueprintResponse` only to use `blueprintResponse.ImageRequests`? This can be simplified.

internal/v1/api.yaml
```diff
@@ -591,6 +591,36 @@ paths:
...
+            format: uuid
+      responses:
+        '200':
+          description: Successful response
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/ImageTypes'
```
lzap, line 623: It is not a good practice to return anything else than a dictionary. Probably okay for this endpoint, it is experimental after all. I am good.

lzap (dismissed):
I am fine, what @ezr-ondrej said.

internal/v1/handler_blueprints.go
```diff
@@ -132,6 +132,31 @@ func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID) error {
...
+	if err != nil {
+		return err
+	}
+
+	blueprint, err := BlueprintFromEntry(blueprintEntry)
+	if err != nil {
+		return err
+	}
+
+	var imageTypes []ImageTypes
```
lzap, line 152: Just a nitpick, you do not have to rebase this here we only have 3 elements but it is good to do this if you know the length in advance:

```suggestion
	imageTypes := make([]ImageTypes, 0, len(blueprint))
```

internal/v1/handler_blueprints_test.go
```diff
@@ -346,6 +346,82 @@ func TestHandlers_GetBlueprint(t *testing.T) {
...
+	require.Equal(t, http.StatusOK, respStatusCode)
+
+	var result []ImageTypes
+	require.Equal(t, 200, respStatusCode)
+	err = json.Unmarshal([]byte(body), &result)
+	require.NoError(t, err)
+	for i, imageType := range result {
+		expectedImageType := blueprint.ImageRequests[i].ImageType
+		require.Equal(t, expectedImageType, imageType)
+	}
```
lzap, line 422: You could use [ElementsMatch](https://pkg.go.dev/github.com/stretchr/testify/require#ElementsMatch), there are many useful functions in that package. But this is okay.

lzap (approved):
Great work.

ezr-ondrej (approved):
Thanks! :+1:

lzap (changes requested):
As we just found out, the data is already exposed through `GetBlueprint` so there is no need for this. Sorry for not catching it up earlier.

PR #1186, opened 2024-05-15
v1: add targets to the composeBlueprint endPoint (HMS-4066)

mgold1234:
this commit add optional target to composeBlueprint endPoint that bring the ability
to build image with specific image types, also add support at the handler function
add unit tests that check this changes

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -572,10 +572,23 @@ paths:
           required: true
           description: UUID of a blueprint
       summary: create new compose from blueprint
-      description: "create new compose from blueprint"
+      description: "create new compose from blueprint, optionally specifying the target image types to build"
       operationId: composeBlueprint
       tags:
         - blueprint
+      requestBody:
+          required: false
+          description: "list of target image types that the user wants to build for this compose"
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  image_types:
+                    type: array
+                    items:
+                      $ref: "#/components/schemas/ImageTypes"
+                    example: ["azure", "aws"]
       responses:
         '201':
           description: compose was created
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -16,6 +16,8 @@ import (
 	"github.com/osbuild/image-builder/internal/common"
 	"github.com/osbuild/image-builder/internal/db"
 
+	"slices"
+
 	"github.com/google/uuid"
 	"github.com/labstack/echo/v4"
 )
@@ -176,6 +178,12 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
 }
 
 func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) error {
+	var requestBody ComposeBlueprintJSONBody
+	err := ctx.Bind(&requestBody)
+	if err != nil {
+		return err
+	}
+
 	userID, err := h.server.getIdentity(ctx)
 	if err != nil {
 		return err
@@ -192,6 +200,9 @@ func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) err
 	composeResponses := make([]ComposeResponse, 0, len(blueprint.ImageRequests))
 	clientId := ClientId("api")
 	for _, imageRequest := range blueprint.ImageRequests {
+		if requestBody.ImageTypes != nil && !slices.Contains(*requestBody.ImageTypes, imageRequest.ImageType) {
+			continue
+		}
 		composeRequest := ComposeRequest{
 			Customizations:   &blueprint.Customizations,
 			Distribution:     blueprint.Distribution,
--- a/internal/v1/handler_blueprints_test.go
+++ b/internal/v1/handler_blueprints_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
@@ -111,7 +112,7 @@ func TestHandlers_UpdateBlueprint(t *testing.T) {
 	require.Equal(t, "Invalid blueprint name", jsonResp.Errors[0].Title)
 }
 
-func TestHandlers_ComposeBlueprint(t *testing.T) {
+func TestHandlers_ComposeBlueprint_without_Targets(t *testing.T) {
 	ctx := context.Background()
 	ids := []uuid.UUID{uuid.New(), uuid.New()}
 	idx := 0
@@ -185,7 +186,7 @@ func TestHandlers_ComposeBlueprint(t *testing.T) {
 	err = dbase.InsertBlueprint(ctx, id, versionId, "000000", "000000", name, description, message)
 	require.NoError(t, err)
 
-	respStatusCode, body := tutils.PostResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/experimental/blueprints/%s/compose", id.String()), map[string]string{})
+	respStatusCode, body := tutils.PostResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/experimental/blueprints/%s/compose", id.String()), strings.NewReader(""))
 	require.Equal(t, http.StatusCreated, respStatusCode)
 
 	var result []ComposeResponse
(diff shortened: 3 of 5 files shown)


ezr-ondrej (commented):
I think you're on the right track, I guess next step would be to filter the composes by `skip if imageTypes.includes(<currentImageType>)` right?

internal/v1/api.yaml
```diff
@@ -1125,6 +1132,15 @@ components:
...
       default: "api"
+    Targets:
+      required:
+        - ImageTypes
+      type: object
+      properties:
+        ImageTypes:
+          type: array
+          items:
+            type: string
```
lzap, line 1143: The `ImageTypes` element appears to be required, should not be required = build all targets.

Please make use of the existing `components/schemas/ImageTypes` type which is in fact enum. Note enums are pain to work with in Go since the language does not support enums and I personally would rather avoid enums in OpenAPI, but since the project already does use them, let’s stick with it.

Please write documentation for this endpoint now, add at least two sentences explaining the behavior.

internal/v1/handler_blueprints.go
```diff
@@ -200,13 +210,13 @@ func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) err
...
+
 		composesResponse, err := h.handleCommonCompose(ctx, composeRequest, &blueprintEntry.VersionId)
 		if err != nil {
 			return err
 		}
 		composeResponses = append(composeResponses, composesResponse)
 	}
-
 	return ctx.JSON(http.StatusCreated, composeResponses)
 }
```
lzap, line 222: This does not appear to be working code yet.

lzap (commented):
Let’s start with OpenAPI and get that right first, then we can move on to implementation.

As we discussed on the phone, you cannot easily use `slices.Contains` so I suggest to search in the list of types manually. Make a helper function and write test for it please.

ezr-ondrej:
> so I suggest to search in the list of types manually. Make a helper function and write test for it please.

Extra points if we could have a general function with another that we already have for it https://github.com/osbuild/image-builder/blob/7bf2d6fbe7f9087a9dade35730ecbe36b2c8bd11/cmd/oscap/main.go#L63

mgold1234:
its very simple condition if imageType != imageRequest.ImageType , why I need to write helper function?

internal/v1/handler_blueprints.go
```diff
@@ -191,22 +197,46 @@ func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) err
...
+				return err
+			}
+			composeResponses = append(composeResponses, composesResponse)
 		}
-		composesResponse, err := h.handleCommonCompose(ctx, composeRequest, &blueprintEntry.VersionId)
-		if err != nil {
-			return err
+	} else {
+		for _, imageRequest := range blueprint.ImageRequests {
+			for _, imageType := range imageTypes {
```
ezr-ondrej, line 219: I don't think this is correct, the nested loop will not do what we want.
What happens when I do have Blueprint with AWS, Azure, Bare metal and vmware targets and I ask for AWS, Azure and Bare metal images here? :thinking:

ezr-ondrej (changes requested):

internal/v1/api.yaml
```diff
@@ -571,11 +572,21 @@ paths:
...
+      requestBody:
+          required: false
+          description: "list of target image types that the user wants to build for this compose"
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/ImageTypes"
+                  example: ["azure", "aws"]
```
lzap, line 589: Please update the schema so the body looks like this:

```json
{
 "types": [ "..." ]
}
```

Not like this:

```json
[ "..." ]
```

Make sure the following input means "build me all types". Example 1 - an empty body. Example 2 - a body that has an empty JSON:

```json
{ }
```

In contrast, the following example means "build me nothing" not "build me all types from blueprint":

```json
{
 "types": [ ]
}
```

At least this is how I see the most logical semantics for the endpoint. What others think?

internal/v1/handler_blueprints.go
lzap: Yes, this can be vastly improved.

We had an offline chat about how to approach this. Unfortunately, the [experimental package](https://pkg.go.dev/golang.org/x/exp/slices#Contains) `slices.Contains` cannot be used because [the type is generated from OpenAPI](https://github.com/osbuild/image-builder/blob/07736c30ffa4db61c5a9921e2e9584cc67d4da9d/internal/v1/api.go#L717) and it needs an annotation

To simplify the code, I suggest to create a helper function that will implement `containsImageType(it ImageType, its []ImageType)` on the particular type. Then use it in the loop, as I said above:

* when image type is `nil` this should mean "build me all types"
* when image type is empty slice this should mean "build nothing", in other words does not need special care

lzap (changes requested):
This is a little bit tricky in Go as the language does not contain a generic "Contains" for slices yet and the experimental one will not work here.

lzap:
One more thing, it looks like this project is still on 1.20, however, 1.21 was recently made available in Red Hat UBI containers so let’s upgrade, then rebase on top of that and you can use `slices` package from the standard library: 

https://github.com/osbuild/image-builder/pull/1192

lzap (changes requested):
Marked as a draft, we talked offline about this yesterday and the API needs some changes. The whole data (array) should be wrapped as dictionary for possible changes in the future without breaking the API.

Also the handling code needs some changes, we will continue working on this.

internal/v1/handler_blueprints.go
```diff
@@ -191,22 +199,44 @@ func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) err
...
+				return err
+			}
+			composeResponses = append(composeResponses, composesResponse)
 		}
-		composesResponse, err := h.handleCommonCompose(ctx, composeRequest, &blueprintEntry.VersionId)
-		if err != nil {
-			return err
+	} else {
+		for _, imageRequest := range blueprint.ImageRequests {
+			if !slices.Contains(*requestBody.ImageTypes, imageRequest.ImageType) {
```
ezr-ondrej, line 221: Can't we just adjust this condition instead of having two branches (if/else) for basically the same thing?
Seeing the same code twice repeat after few lines scares me.
```suggestion
			if requestBody.ImageTypes != nil && !slices.Contains(*requestBody.ImageTypes, imageRequest.ImageType) {
```

internal/v1/handler_blueprints_test.go
```diff
@@ -185,7 +185,94 @@ func TestHandlers_ComposeBlueprint(t *testing.T) {
...
+
+	var result []ComposeResponse
+	err = json.Unmarshal([]byte(body), &result)
+	require.NoError(t, err)
+	require.Len(t, result, 2)
+	require.Equal(t, ids[0], result[0].Id)
+	require.Equal(t, ids[1], result[1].Id)
+}
+
+func TestHandlers_ComposeBlueprint_with_Target(t *testing.T) {
```
ezr-ondrej, line 199: The test looks great, tho could we add a second call at the end of it requesting just one of the targets and verifying it actually built just one image? :)

ezr-ondrej (changes requested):
Overall looks pretty great, I think we can be bit more dry and add bit more test coverage :)
See my inline comments and let me know what you thing ;)

internal/v1/handler_blueprints.go
```diff
@@ -176,6 +178,12 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
 }
 
 func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) error {
+	var requestBody ComposeBlueprintJSONBody
+	err := ctx.Bind(&requestBody)
```
ezr-ondrej, line 182: I'm not sure if we can check for empty body, it would be nice to be able to send empty body and not fail :)

ezr-ondrej (approved):
Nice work, looks good to me, just one nit, but not really important! :)

ezr-ondrej:
@lzap can you take a look once you get to it? :)

lzap:
Sure, needs to be rebased, tests are failing for some reason.

[API doc for the latest commit](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/osbuild/image-builder/b4aa339e386bafe9ed03ff0a67225e7c3fa24c79/internal/v1/api.yaml#tag/blueprint/operation/composeBlueprint).

internal/v1/handler_blueprints.go
lzap: There must be a test which sends no body (body length is 0), otherwise we are breaking the API thus current UI.

internal/v1/handler_blueprints.go
```diff
@@ -192,6 +200,9 @@ func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) err
 	composeResponses := make([]ComposeResponse, 0, len(blueprint.ImageRequests))
 	clientId := ClientId("api")
 	for _, imageRequest := range blueprint.ImageRequests {
+		if requestBody.ImageTypes != nil && !slices.Contains(*requestBody.ImageTypes, imageRequest.ImageType) {
+			continue
+		}
```
lzap, line 205: Three lines, yay!

internal/v1/handler_blueprints.go
```diff
@@ -206,7 +217,6 @@ func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) err
 		}
 		composeResponses = append(composeResponses, composesResponse)
 	}
-
```
lzap, line 209: I'd keep the line, personal preference. No need to rebase tho.

internal/v1/handler_blueprints_test.go
```diff
@@ -185,7 +185,7 @@ func TestHandlers_ComposeBlueprint(t *testing.T) {
 	err = dbase.InsertBlueprint(ctx, id, versionId, "000000", "000000", name, description, message)
 	require.NoError(t, err)
 
-	respStatusCode, body := tutils.PostResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/experimental/blueprints/%s/compose", id.String()), map[string]string{})
+	respStatusCode, body := tutils.PostResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/experimental/blueprints/%s/compose", id.String()), ComposeBlueprintJSONBody{})
```
lzap, line 188: Please change this test case to send zero body, this here actually sends the following body:

`{}`

This is not what UI does today, I want to be sure we will not break it when we merge this.

In case you are wondering why there was a map previously, somebody was lazy and since the helper function requires `any` value (must not be `nil`), they used a map for some reason:

```
func PostResponseBody(t *testing.T, url string, compose interface{}) (int, string)
```

The proper solution is/was to modify the helper `PostResponseBody` to handle `nil` correctly: send no body.

lzap (changes requested):
Very good, I have a request for the existing test - it currently sends a body without any field in the root which is correct, however, I want to be sure there is a test for no body (zero length body). That is what UI does today.

internal/v1/handler_blueprints_test.go
mgold1234: that exactelly what I thoght, will do it, thank you!

ezr-ondrej: Yeah, these [lazy people](https://github.com/osbuild/image-builder/commit/3c932f7e4da41c7c16d8b68594cf9e15b65c941c) :roll_eyes: (to my defense I probably have copied it somewhere).
If you struggle to actually send empty request, let me know, I can help out correct my past mistakes :)

lzap: Wow I was pretty sure it was me, okay. :-D

internal/v1/handler_blueprints.go
mgold1234: added test that checjk it

lzap:
Test failures seem relevant.

internal/tutils/tutils.go
```diff
@@ -55,9 +55,14 @@ func GetResponseBody(t *testing.T, url string, auth *string) (int, string) {
...
-	require.NoError(t, err)
-
+	var buf []byte
+	var err error
+	if compose == nil {
+		buf = []byte("{}") // Send an empty JSON object if compose is nil
+	} else {
+		buf, err = json.Marshal(compose)
+		require.NoError(t, err)
+	}
```
lzap, line 65: Please do not send any body on `nil` argument. That is what I meant by zero-length body. This might be little bit tricky, let me explain. [NewRequest](https://pkg.go.dev/net/http#NewRequest) expects `io.Reader` so there are several options how you can achieve this:

* You can use an empty string: `strings.NewReader("")`
* Similarly byte slice that contains nothing: `bytes.NewReader([]byte)`
* Or perhaps `bytes.NewReader(nil)` does the same thing
* To my surprise, there is no `io.NullReader` in the standard library, there is `ioutil.Discard` do, a writer that does nothing.

lzap (changes requested):
Tests are failing and I believe it is because handling of empty POST is not yet correct.

internal/tutils/tutils.go
mgold1234: what is the different between sending strings.NewReader("") or ComposeBlueprintJSONBody{} both of them bring this result

jrusz:
Hm, the error has changed from:
```
HTTP response body: {
  "errors": [
    {
      "detail": "cannot parse request body: EOF",
      "title": "400"
    }
  ]
}
```
to:
```
12:21:57 E       HTTP response body: {
12:21:57 E         "errors": [
12:21:57 E           {
12:21:57 E             "detail": "request must be json-encoded",
12:21:57 E             "title": "415"
12:21:57 E           }
12:21:57 E         ]
12:21:57 E       }
```
Is there still something missing in the code? Regenerating the api didn't help in the plugin...

internal/tutils/tutils.go
lzap: The former sends nothing (no body). The latter sends '{}' in the body.

You can also define a helper type to do the same:

```go
type eofReader struct{}

func (eofReader) Read([]byte) (int, error) {
	return 0, io.EOF
}

var EOFReader io.Reader = eofReader{}
```

When you pass `EOFReader` it immediately returns EOF (end of file), so no body is sent. Does it make sense?

internal/v1/handler_blueprints.go
```diff
@@ -176,6 +178,12 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
 }
 
 func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) error {
+	var requestBody ComposeBlueprintJSONBody
+	err := ctx.Bind(&requestBody)
+	if err != nil {
+		return err
+	}
```
lzap, line 185: Just to be sure, these 5 lines are breaking tests. The loading (binding) should be optional. I am not sure how to do this with echo library tho, I have never done that. The docs say that binding depends on Content-Type header. Perhaps check the error when you call the endpoint without any body:

```
POST .../blueprints/UUID/compose
Content-Length: 0
```

or

```
POST .../blueprints/UUID/compose
Content-Length: 0
Content-Type: application/json
```

In both cases, I think the endpoint should work as before - compose all image types. See what UI does in Chrome inspector.

lzap: Yeah [here is the code of Bind](https://github.com/labstack/echo/blob/master/bind.go#L67).

* It returns immediately if `Content-Length: 0` this is what should be set by UI and tests at the moment.
* It tries to unmarshal JSON if and only if `Content-Length` is higher than zero and `Content-Type` is `application/json`.

lzap: @jrusz can you tell what `Content-Length` and `Content-Type` headers tests are passing at the moment?

lzap (dismissed):
The code looks good. Tests might take advantage of table-driven testing but this is not something you need to do.

Now, API tests are failing, waiting for @jrusz for comment on that. The API endpoint should now work without any body, or with a valid JSON body.

jrusz:
> @jrusz if we allow empty body then tests will fix automatically, if we enforce JSON body then the current "compose from blueprint" tests will have to send an empty JSON body `{ }` for all tests cases

Ok I see what you mean now. I've deployed this in ephemeral and tried to send the request manually and it does require at least the empty body. This means it really breaks the current behavior. Since that new `request_body` options is supposed to be optional I'd be for not breaking the current behavior.

ezr-ondrej (approved):
This is shaping up nicely :+1:

internal/v1/handler_blueprints_test.go
```diff
@@ -196,6 +197,179 @@ func TestHandlers_ComposeBlueprint(t *testing.T) {
...
+	respStatusCode, body := tutils.PostResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/experimental/blueprints/%s/compose", id.String()), payload)
+	require.Equal(t, http.StatusCreated, respStatusCode)
+
+	var result []ComposeResponse
+	err = json.Unmarshal([]byte(body), &result)
+	require.NoError(t, err)
+	require.Len(t, result, 1)
+	require.Equal(t, ids[0], result[0].Id)
+}
+
```
lzap, line 372: I do not want to block you anymore, but would you mind taking a look later at [table driven testing](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests)? I think this test (and many others) might benefit from it. This needs a research, chances are it might be not suitable but I think this is worth exploring.

internal/v1/server.go
```diff
@@ -181,6 +181,9 @@ type binder struct{}
 
 func (b binder) Bind(i interface{}, ctx echo.Context) error {
 	request := ctx.Request()
+	if request.ContentLength == 0 {
+		return nil
+	}
```
lzap, line 186: This now allows creation of endpoints without any body. Previously JSON was enforced, I believe this is okay.

lzap (approved):
Thanks, great work.

lzap:
Thanks, merged. If anyone thinks there should be JSON presence body enforcement, we can remove the three lines but IQE tests and docs will need to be changed.

internal/v1/server.go
ezr-ondrej: Yeah, I think we just didn't need it anywhere, but it was actually a "bug" IMO :)

internal/v1/handler_blueprints_test.go
ezr-ondrej: I had to give it a try: https://github.com/osbuild/image-builder/pull/1206 :)
Let me know if you'd want to take over @mgold1234 ;)

PR #1194, opened 2024-05-21
multi: use `slices.Contains`

kingsleyzissou:
Since we have upgraded to go v1.21 we can now make use of the `slices.Contains` function which checks to see if a slice contains a particular element. This commit switches all custom implementations of the `slices.Contains` function.


ezr-ondrej (approved):
Quick reaction, :orange_heart: it.
Thank you, nice cleanup :+1:

lzap:
Nice, it was a long wait :-)

PR #1202, opened 2024-05-27
floorplan: add blueprint version and id

ezr-ondrej:
Done in collaboration with @avitova and tested on few local data, so it should work :thinking:


ochosi (approved):
Looks plausible!

PR #1203, opened 2024-05-27
v1/api: move blueprints endpoints out of experimental (HMS-4200)

ezr-ondrej:
Add non-experimental version of Blueprints endpoints.

Add wrappers for experimental endpoints, so we can serve both.
Wrappers will be removed once ready in #1204

Good opportunity for an API re-review :)

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -97,6 +97,256 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPErrorList'
+  /blueprints:
+    get:
+      summary: get a collection of blueprints
+      description: "get a collection of blueprints, returns just the latest version of each blueprint"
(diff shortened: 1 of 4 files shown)


internal/v1/api.yaml
```diff
@@ -595,7 +832,7 @@ paths:
     get:
       summary: get composes associated with a blueprint
       description: "get a collection of composes associated to a blueprint, allows for filtering by version"
-      operationId: getBlueprintComposes
+      operationId: getBlueprintComposesExperimental
```
lzap, line 835: Sharing the id is not supported? That would enable you to drop the wrapper functions.

lzap (approved):

lzap (approved):
Ok let’s do this, rebasing.

ezr-ondrej:
We are :green_apple: :)

PR #1235, opened 2024-06-13
templates/iqe: add parameter to disable parallel runs

jrusz:
We don't have any parallel test cases yet.


jrusz:
I accidentally pushed added and pushed this to https://github.com/osbuild/image-builder/pull/1233 which is actually fine, so closing this.

PR #1240, opened 2024-06-19
Fix blueprints status codes and add test

avitova:
We were missing a few 404 status codes. I added a few lines returning these, and also added some tests.

### DIFF
--- a/internal/v1/handler_blueprints_test.go
+++ b/internal/v1/handler_blueprints_test.go
@@ -110,6 +110,10 @@ func TestHandlers_UpdateBlueprint(t *testing.T) {
 	err = json.Unmarshal([]byte(resp), &jsonResp)
 	require.NoError(t, err)
 	require.Equal(t, "Invalid blueprint name", jsonResp.Errors[0].Title)
+
+	body["name"] = "Changing to correct body"
+	respStatusCodeNotFound, _ := tutils.PutResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/experimental/blueprints/%s", uuid.New()), body)
+	require.Equal(t, http.StatusNotFound, respStatusCodeNotFound)
 }
 
 func TestHandlers_ComposeBlueprint(t *testing.T) {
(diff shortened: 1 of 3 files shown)


internal/v1/handler_blueprints_test.go
```diff
@@ -295,6 +299,14 @@ func TestHandlers_GetBlueprintComposes(t *testing.T) {
...
 	require.Equal(t, blueprintId, *result.Data[0].BlueprintId)
 	require.Equal(t, 2, *result.Data[0].BlueprintVersion)
+
+	//// get composes for non-existing blueprint
+	respStatusCode, body = tutils.GetResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/experimental/blueprints/%s/composes?blueprint_version=1", uuid.New().String()), &tutils.AuthString0)
+	require.Equal(t, 200, respStatusCode)
+	err = json.Unmarshal([]byte(body), &result)
+	require.NoError(t, err)
+	require.Equal(t, 0, len(result.Data))
+	require.Equal(t, 0, result.Meta.Count)
```
ezr-ondrej, line 309: I'm unsure here: should we rather return 404 as well, this feels bit weird to have empty response for non-existent blueprint

ezr-ondrej (commented):
Nice improvements! :orange_heart:
Do you think we should also return 404 on listing composes for non-existent blueprint?

internal/v1/handler_blueprints_test.go
avitova: Yes, this is interesting; we'd have to distinguish between Blueprint Not Found vs No Composes. This is the case for most of the endpoints where we use JOINS. I am happy to discuss how much we want to enhance this.

avitova:
I have no strong opinion on this; I see a difference between 404 for a non-existent blueprint and 200 for empty composes for a given blueprint. I would have a more extensive discussion on the topic outside GitHub and address this in different PR. Does that sound good to you?

ezr-ondrej:
> I would have a more extensive discussion on the topic outside GitHub and address this in different PR. Does that sound good to you?

Agreed, lets just not add tests for it here as what's tested is intentional :) and when we don't know how it should behave, lets not test it ;)

internal/v1/handler_blueprints_test.go
lzap: If a bluprint does not exist in a `/blueprints/XYZ/whatever` path to me it feels logical to see Not Found just like this was a directory on a file system.

ezr-ondrej:
Let me know once this is ready again, I still see the test there, so I'll hold off the review :)

avitova:
The mentioned test is for: "existing blueprint has no composes" case. Imo that does not have to be 404, just an empty list, and a 200 response.
There is an additional test for listing compossd of non-existing blueprint => ends up with 404.

ezr-ondrej (approved):
Thanks! :)

ezr-ondrej:
Oh wow, I was sure I've merged this, sorry! Merging.

avitova:
Thank you!:)

PR #1246, opened 2024-06-24
v1/api: Blueprints soft delete

avitova:
Soft delete should help us measure the success and collect all the data.

### DIFF
--- a/internal/db/db_blueprints.go
+++ b/internal/db/db_blueprints.go
@@ -29,40 +29,44 @@ const (
 	sqlLatestBlueprintVersion = `
 		SELECT MAX(blueprint_versions.version)
 		FROM blueprint_versions INNER JOIN blueprints ON blueprint_versions.blueprint_id = blueprints.id
-		WHERE blueprints.org_id = $1 AND blueprints.id = $2`
+		WHERE blueprints.deleted = FALSE AND blueprints.org_id = $1 AND blueprints.id = $2`
 
 	sqlGetBlueprintComposes = `
 		SELECT blueprint_versions.version, composes.job_id, composes.request, composes.created_at, composes.image_name, composes.client_id
 		FROM composes INNER JOIN blueprint_versions ON composes.blueprint_version_id = blueprint_versions.id
-		WHERE composes.org_id = $1
+		INNER JOIN blueprints ON blueprint_versions.blueprint_id = blueprints.id
+		WHERE composes.org_id = $1 AND blueprints.org_id = $1
 			AND blueprint_versions.blueprint_id = $2
 			AND ($3::integer IS NULL OR blueprint_versions.version = $3)
 			AND CURRENT_TIMESTAMP - composes.created_at <= $4
 			AND ($5::text[] is NULL OR composes.request->'image_requests'->0->>'image_type' <> ALL($5))
 			AND composes.deleted = FALSE
+		    AND blueprints.deleted = FALSE
 		ORDER BY composes.created_at DESC
 		LIMIT $6 OFFSET $7`
 
 	sqlCountActiveBlueprintComposesSince = `
 		SELECT COUNT(*)
 		FROM composes INNER JOIN blueprint_versions ON composes.blueprint_version_id = blueprint_versions.id
-		WHERE org_id=$1
+		INNER JOIN blueprints ON blueprint_versions.blueprint_id = blueprints.id
+		WHERE composes.org_id = $1 AND blueprints.org_id = $1
 			AND blueprint_versions.blueprint_id = $2
 		  	AND ($3::integer IS NULL OR blueprint_versions.version = $3)
 		  	AND CURRENT_TIMESTAMP - composes.created_at <= $4
 		  	AND composes.deleted = FALSE
+		    AND blueprints.deleted = FALSE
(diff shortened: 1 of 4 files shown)


internal/v1/handler_blueprints_test.go
```diff
@@ -494,19 +494,20 @@ func TestHandlers_DeleteBlueprint(t *testing.T) {
 	require.Len(t, result.Data, 0)
 
 	// We should not be able to update deleted blueprint
-	id5 := uuid.New()
-	err = dbase.UpdateBlueprint(ctx, id5, blueprintId, "000000", "newName", "desc2", json.RawMessage(`{"image_requests": [{"image_type": "aws"}, {"image_type": "gcp"}]}`))
+	err = dbase.UpdateBlueprint(ctx, blueprintId, blueprintId, "000000", "newName", "desc2", json.RawMessage(`{"image_requests": [{"image_type": "aws"}, {"image_type": "gcp"}]}`))
```
ezr-ondrej, line 497: Why are we changing this? I know it's useless to generate new UUID, but it might fail for other reasons when we do pass already used ID, so I'd rather keep only one possible failure, the one we test for.
Was there a reason other than safe some bytes?

templates/image-builder.yml
```diff
@@ -182,6 +182,7 @@ objects:
           req->>'snapshot_date' as snapshot_date,
           v.blueprint_id::text as blueprint_id,
           v.version as blueprint_version,
+          b.deleted_at as blueprint_deleted_at,
```
ezr-ondrej, line 185:
```suggestion
          b.deleted as blueprint_deleted,
```

internal/db/migrations-tern/012_add_blueprints_deleted_flag.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE blueprints ADD deleted boolean NOT NULL DEFAULT FALSE;
```
ezr-ondrej, line 1: Looks safe, @lzap, WDYT? :)

ezr-ondrej (commented):
Few nitpicks, floorist probably leftover from when deleted was supposed to be a timestamp? :)

internal/db/migrations-tern/012_add_blueprints_deleted_flag.sql
lzap: Yes, this is definitely correct. There is no point in creating index on a boolean column, it will almost never be used as table scan is usually faster on this type.

internal/v1/handler_blueprints_test.go
lzap: I also do not understand, if this is some irrelevant test improvement, I suggest to move it into a separate PR because this change should work without touching a single test.

lzap (commented):
What was said.

Also, I am missing a test that actually verifies that deleted records are in fact not visible. I do not think you have to test all possible SQL statements you changed tho, just pick one or few.

internal/db/db_blueprints.go
```diff
@@ -89,41 +94,41 @@ const (
 		WHERE composes.blueprint_version_id = blueprint_versions.id
 		AND composes.org_id=$1 AND blueprint_versions.blueprint_id=$2`
 
-	sqlDeleteBlueprint = `DELETE FROM blueprints WHERE id = $1 AND org_id = $2 AND account_number = $3`
+	sqlDeleteBlueprint = `UPDATE blueprints SET deleted = TRUE, name = id WHERE deleted = FALSE AND id = $1 AND org_id = $2 AND account_number = $3`
```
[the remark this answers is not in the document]
ezr-ondrej: We need to be able to create new blueprint with the same name

internal/v1/handler_blueprints_test.go
ezr-ondrej: The other change is needed, we've used the same UUID to recreate blueprint with the same name, but that won't work anymore, we need new uuids as the current one will be already be there and we'd fail unique test

internal/db/db_blueprints.go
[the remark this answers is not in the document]
avitova: Yes, after deleting this one, you should be able to create a new blueprint with the same name. And we have a constraint on the name. This is something that @croissanne suggested, and it seems reasonable. The only downside is if someone wanted to create a blueprint with the given UUID. But I do not see that like a big problem.

internal/v1/handler_blueprints_test.go
avitova: This was unintentional, and you are correct. This does not belong in here.

avitova:
Thank you for the comments. I addressed everything. The test for not being able to access deleted blueprints is [already done](https://github.com/osbuild/image-builder/blob/53e05e5d5910f3cfa5e18032167cdd56e089dac7/internal/v1/handler_blueprints_test.go#L489). I have prepared this in the [previous PR](https://github.com/osbuild/image-builder/pull/1240). Do you have more tests in mind? @lzap

ezr-ondrej (approved):
The migration is correct, the test is there, I think we are ready to go.

@croissanne would you mind checking you're ok with:
* Updating the name of deleted blueprint to UUID
* having the deleted as a boolean flag

Looks reasonable to me, but I might be missing something :)

PR #1252, opened 2024-06-26
v1: switch to RFC3339 when querying snapshots for a date (HMS-4351)

croissanne:
The content-sources api is switching to RFC3339.


xbhouse (approved):
thank you! ack!

PR #1256, opened 2024-06-30
refactor token of composer client <HMS-4369>

mgold1234:
this commit refactor token of composer client, and using the lazy token

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -87,12 +87,13 @@ func main() {
 	}
 
 	composerConf := composer.ComposerClientConfig{
-		ComposerURL:  conf.ComposerURL,
-		CA:           conf.ComposerCA,
-		TokenURL:     conf.ComposerTokenURL,
-		ClientId:     conf.ComposerClientId,
-		OfflineToken: conf.ComposerOfflineToken,
-		ClientSecret: conf.ComposerClientSecret,
+		URL: conf.ComposerURL,
+		CA:  conf.ComposerCA,
+		Tokener: &oauth2.LazyToken{
+			Url:          conf.ComposerTokenURL,
+			ClientId:     conf.ComposerClientId,
+			ClientSecret: conf.ComposerClientSecret,
+		},
 	}
 
 	recommendationConf := recommendations.RecommendationsClientConfig{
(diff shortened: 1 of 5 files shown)


internal/clients/composer/client.go
```diff
@@ -107,67 +87,33 @@ func (cc *ComposerClient) request(method, url string, headers map[string]string,
...
 
-	if resp.StatusCode == http.StatusUnauthorized {
-		err = cc.refreshToken()
+	if resp.StatusCode == http.StatusForbidden || resp.StatusCode == http.StatusUnauthorized {
+		token, err = cc.tokener.ForceRefresh(req.Context())
+		if err != nil {
+			return nil, err
+		}
+		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+		_, err = body.Seek(0, io.SeekStart)
```
croissanne, line 107: `body` could be `nil` here, so add a `if body != nil` check before seeking. If it's a get request, there's no body.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/osbuild/image-builder
 
-go 1.21
+go 1.22.1
```
croissanne, line 3: So for the go version we pick whatever is available in the registry.access.redhat.com/ubi9/go-toolset container. Which is currently `1.21.10`.

croissanne (commented):
Yay! Looks good, some small comments.

cmd/image-builder/main.go
```diff
@@ -87,12 +87,13 @@ func main() {
...
-		ComposerURL:  conf.ComposerURL,
-		CA:           conf.ComposerCA,
-		TokenURL:     conf.ComposerTokenURL,
-		ClientId:     conf.ComposerClientId,
-		OfflineToken: conf.ComposerOfflineToken,
-		ClientSecret: conf.ComposerClientSecret,
+		URL: conf.ComposerURL,
+		CA:  conf.ComposerCA,
+		Tokener: &oauth2.LazyToken{
+			Url:          conf.ComposerURL,
```
croissanne, line 93: This needs to be the `ComposerTokenURL`

croissanne (changes requested):
Thanks, small bug, otherwise ok.

croissanne (approved):
LGTM, thank you!

PR #1259, opened 2024-07-02
internal: add RECOMMENDATIONS_PROXY to the configuration

mgold1234:
this commit add RECOMMENDATIONS_PROXY to image builder configuration to support the vpn problem

### DIFF
--- a/internal/clients/recommendations/client.go
+++ b/internal/clients/recommendations/client.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -20,11 +21,13 @@ type RecommendationsClient struct {
 	URL     string
 	tokener oauth2.Tokener
 	client  *http.Client
+	Proxy   string
 }
 
 type RecommendationsClientConfig struct {
 	URL     string
 	CA      string
+	Proxy   string
 	Tokener oauth2.Tokener
 }
 
@@ -34,13 +37,14 @@ func NewClient(conf RecommendationsClientConfig) (*RecommendationsClient, error)
 	if conf.URL == "" {
 		logrus.Warn("Recommendation URL not set, client will fail")
 	}
-	client, err := createClient(conf.URL, conf.CA)
+	client, err := createClient(conf.URL, conf.CA, conf.Proxy)
 	if err != nil {
 		return nil, fmt.Errorf("error creating recommend http client")
 	}
 
 	rc := RecommendationsClient{
 		URL:     conf.URL,
+		Proxy:   conf.Proxy,
 		tokener: conf.Tokener,
 		client:  client,
 	}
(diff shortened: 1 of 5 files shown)


croissanne (dismissed):
Thanks!

internal/clients/recommendations/client.go
```diff
@@ -20,11 +20,13 @@ type RecommendationsClient struct {
 	URL     string
 	tokener oauth2.Tokener
 	client  *http.Client
+	PROXY   string
 }
 
 type RecommendationsClientConfig struct {
 	URL     string
 	CA      string
+	PROXY   string
```
croissanne, line 29: The naming convention should make it `Proxy` here.

croissanne (changes requested):

croissanne:
You'll need to add the proxy to the client:

https://pkg.go.dev/net/http#Client
https://pkg.go.dev/net/http#Transport
https://pkg.go.dev/net/http#ProxyURL

internal/clients/recommendations/client.go
```diff
@@ -34,40 +37,49 @@ func NewClient(conf RecommendationsClientConfig) (*RecommendationsClient, error)
...
+		transport = &http.Transport{
+			TLSClientConfig: tlsConfig,
+		}
+	} else {
+		transport = &http.Transport{}
 	}
 
-	transport := &http.Transport{TLSClientConfig: tlsConfig}
+	if proxyURL != "" {
+		proxyURLParsed := url.URL{Host: proxyURL, Scheme: "http"}
```
croissanne, line 80: Why not use https://pkg.go.dev/net/url#Parse?

internal/v1/server_test.go
```diff
@@ -149,6 +150,7 @@ func startServer(t *testing.T, tscc *testServerClientsConf, conf *ServerConfig)
 	}
 	recommendClient, err := recommendations.NewClient(recommendations.RecommendationsClientConfig{
 		URL:     tscc.RecommendURL,
+		PROXY:   tscc.PROXY,
```
croissanne, line 153: `PROXY` -> `Proxy`

internal/v1/server_test.go
```diff
@@ -100,6 +100,7 @@ type testServerClientsConf struct {
 	CSURL        string
 	RecommendURL string
 	OAuthURL     string
+	PROXY        string
```
croissanne, line 103: `PROXY` -> `Proxy`

croissanne (changes requested):
Thanks, some small comments/question.

internal/clients/recommendations/client.go
mgold1234: @croissanne I tried to use proxyURLParsed, err := url.Parse(proxyURL), but it is not working, 
beucase we didnt define the http:// for proxy paramemter at Vault system.

croissanne: ah well I'll update it to include the scheme then.

croissanne: updated!

internal/v1/server_test.go
croissanne: This is still valid, naming convention

croissanne (commented):
Thank you, 1 more comment

croissanne (approved):
Perfect, thank you!

PR #1264, opened 2024-07-09
v1/api: Add export endpoint for blueprints (HMS-4451)

avitova:
The idea behind this is to have an endpoint for export. We would use this in the frontend for exporting blueprints.
Using a header "Content-Type" or "Accept" with "application/json;export=true" adds the metadata for the blueprint, omits ImageRequests (targets) and omits Activation Key. Maybe it makes sense to set the Subscription to nil overall. What do you think?
Users will not be able to create a blueprint with this exported JSON, as the "ImageRequests" field is mandatory. This feels a bit wrong in the API, but this is what we wanted users to specify in the Wizard.

We can also discuss other ways to add this to the API, such as parameter/different path for export, such as GET xxx/blueprints/{id}/export. I am open to this discussion.

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -242,6 +242,35 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPErrorList'
+  /blueprints/{id}/export:
+    parameters:
+      - in: path
+        name: id
+        schema:
+          type: string
+          format: uuid
+        example: '123e4567-e89b-12d3-a456-426655440000'
+        required: true
+        description: UUID of a blueprint
+    get:
+      summary: export a blueprint
+      description: "export a blueprint"
+      operationId: exportBlueprint
+      tags:
+        - blueprint
+      responses:
+        '200':
+          description: detail of a blueprint
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BlueprintExportResponse'
+        '404':
+          description: blueprint was not found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPErrorList'
   /blueprints/{id}/compose:
     post:
       parameters:
@@ -1194,6 +1223,8 @@ components:
             Array of image requests. Having more image requests in a single blueprint is currently not supported.
         customizations:
           $ref: '#/components/schemas/Customizations'
+        metadata:
+          $ref: '#/components/schemas/BlueprintMetadata'
     CreateBlueprintResponse:
       required:
         - id
@@ -1262,6 +1293,35 @@ components:
             Array of image requests. Having more image requests in a single blueprint is currently not supported.
         customizations:
           $ref: '#/components/schemas/Customizations'
+    BlueprintExportResponse:
+      required:
+        - name
+        - description
+        - distribution
+        - customizations
+        - metadata
+      properties:
+        name:
+          type: string
+        description:
+          type: string
+        distribution:
+          $ref: '#/components/schemas/Distributions'
+        customizations:
+          $ref: '#/components/schemas/Customizations'
+        metadata:
+          $ref: '#/components/schemas/BlueprintMetadata'
+    BlueprintMetadata:
+      required:
+        - parent_id
+        - exported_at
+      properties:
+        parent_id:
+          type: string
+          format: uuid
+          nullable: true
+        exported_at:
+          type: string
     Distributions:
       type: string
       description: |
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -69,6 +69,11 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
 		return err
 	}
 
+	metadata, err := json.Marshal(blueprintRequest.Metadata)
+	if err != nil {
+		return err
+	}
+
 	if !blueprintNameRegex.MatchString(blueprintRequest.Name) {
 		return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
 			Errors: []HTTPError{{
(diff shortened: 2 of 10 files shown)


internal/v1/api.yaml
```diff
@@ -220,6 +220,9 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/BlueprintResponse'
+            application/json;export=true:
```
ezr-ondrej, line 223: I think that generaly changing response body significantly on change of a header is bit of an anti-pattern, I'm not completely against it, but it smells a bit :)
I'd rather do an edpoint personally.

lzap: I agree with an explicit endpoint, semantically import/export will be quite different from create/show operations.

internal/v1/api.yaml
```diff
@@ -1262,6 +1267,39 @@ components:
...
+      required:
+        - parent_id
+        - exported_at
+      properties:
+        parent_id:
+          type: string
+          format: uuid
+          nullable: true
+        exported_at:
+          type: string
```
lzap, line 1302: Consider creating two root elements: metadata + data so the root namespace is nice and clean. But this works too.

internal/v1/handler_blueprints.go
```diff
@@ -120,11 +125,31 @@ func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID) error {
...
+			blueprint.Customizations.Subscription.ActivationKey = ""
+		}
+
+		blueprintExportResponse := BlueprintExportResponse{
+			Name:           blueprintEntry.Name,
+			Description:    blueprintEntry.Description,
+			Distribution:   blueprint.Distribution,
+			Customizations: blueprint.Customizations,
+			Metadata: BlueprintMetadata{
+				ExportedAt: time.Now().String(),
```
lzap, line 146: This will work, I would probably suggest to return UTC. This will return a string in server timezone (but including TZ info so it is semantically correct).

internal/v1/api.yaml
avitova: Thinking about this comment. Do you like the idea of having the metadata + data only for the exported blueprint? So that we export it like that? Then there is a thought of mine that we do not even have to export the metadata at all, as it does not really tell users that much.
Or would you like to see the "data" element even in the case of the CreateBlueprintRequest, where we have the metadata as well? That would be a big change of the API.

lzap: My thinking is if we are going to create a new endpoint `xxx/export` then we can design whatever we want and I was thinking to create a "clean" root namespace by separating it into metadata part and the actual data. Now, I have weak arguments why to do it other than it creates a "clean" root namespace, it is just an idea. What you have works just fine, as long as "metadata" is not added into the blueprint itself.

If course if you want to use the Content-Type approach, then it does not make any sense and my comment is irrelevant in that case. Apologies for the confusion there.

lzap: Okay I take my comment back, I haven’t realized metadata are going to be stored alongside the data for recording purposes. This is fine, what I propose does worsen things actually.

internal/v1/api.yaml
lzap: As an alternative, consider using just URL parameter: `GET blueprint/XXX?export=true` if we document exactly what it does that is something that can be less confusing.

avitova:
Changes: Create a /blueprints/{id}/export, change timestamp to UTC.

internal/v1/handler_blueprints.go
```diff
@@ -137,6 +143,43 @@ func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID) error {
...
+		return err
+	}
+
+	blueprint, err := BlueprintFromEntry(blueprintEntry)
+	if err != nil {
+		return err
+	}
+
+	if blueprint.Customizations.Subscription != nil {
+		blueprint.Customizations.Subscription.ActivationKey = ""
```
amirfefer, line 167: No need to omit the `Subscription` object  entirely?

internal/v1/api.yaml
```diff
@@ -242,6 +242,35 @@ paths:
...
+        name: id
+        schema:
+          type: string
+          format: uuid
+        example: '123e4567-e89b-12d3-a456-426655440000'
+        required: true
+        description: UUID of a blueprint
+    get:
+      summary: get detail of a blueprint
+      description: "get a blueprint detail"
```
ezr-ondrej, line 257: This would be nice to adjust to better describe purpose of this endpoint ;)

avitova:
Addressed the comments at [1a0d19f](https://github.com/osbuild/image-builder/commit/1a0d19f3b19a3a77a4efd9c73d24410184cf6dbf) and merged main afterwards. There is also a change in image-builder.yml, we were missing plural.

lzap (approved):
I know we suggested a separate endpoint for export, but remind me what was actually wrong with returning metadata in normal get blueprint endpoint? I particularly did no like a switch that was including/excluding the metadata, that one I remember but.

What if the endpoint always returned metadata? The UI can simply ignore it. Why we did not go for that simple solution?

ezr-ondrej:
> What if the endpoint always returned metadata? The UI can simply ignore it. Why we did not go for that simple solution?

I believe we want to filter out some data in the export endpoint. Metadata would be fine in both endpoints, but next step is to filter out data, that we want to keep in the get endpoint.

avitova:
Exactly as Ondrej said. Filtering out data is the main part here. We could add metadata for the GET endpoint, but I do not see a reason why you would need to see the metadata as a user.
Import and export are currently handled by the front end, but omitting data is an extra step that I think we should already do on the backend. And I believe that someone could actually use the feature using the API.

lzap:
Okay, I rebased. I am good, thanks.

lzap (dismissed):

avitova:
One small change that did not change the functionality: id in the api.yaml. We did not export the id in the handler anyways, and do not want to, so I just removed it. Thanks for raising that @ezr-ondrej.

lzap (approved):

internal/v1/handler_blueprints.go
```diff
@@ -69,6 +69,11 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
 		return err
 	}
 
+	metadata, err := json.Marshal(blueprintRequest.Metadata)
```
ezr-ondrej, line 72: If there are no metadata, this results in `null` string (or rather byte array) https://go.dev/play/p/9o0KYX8FiNy
Are we ok with that, or would we rather have actual nil and NULL in the DB?

It's not a big deal really and can be handled in a follow-up.

ezr-ondrej (approved):

internal/v1/handler_blueprints.go
lzap: I slightly lean towards having `NULL` in the DB which is more efficient on storage.

PR #1266, opened 2024-07-10
 fix the 500 internal server error with retry (HMS-4427)

mgold1234:
this commit will fix the internal server error on production  when querying compose status


internal/v1/handler.go
```diff
@@ -229,27 +230,40 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
 		return err
 	}
 
-	resp, err := h.server.cClient.ComposeStatus(composeId)
+	var resp *http.Response
+	var body []byte
+
+	err = retry.Do(
```
mvo5, line 236: I'm not sure this will do the retries we need here. From the errors `"levelname":"error","message":"Internal error 500: Failed querying compose status, code=500, message=Failed querying compose status, internal=upstream connect error or disconnect/reset before headers. reset reason: connection failure"` in the log it seems that we get an `err == nil` here, the culprit is the underlying resp.StatusCode. I would start like this:

1. Include the failed status code from the underlying request, something like:
```go
diff --git a/internal/v1/handler.go b/internal/v1/handler.go
index e8d1491..3f75ca6 100644
--- a/internal/v1/handler.go
+++ b/internal/v1/handler.go
@@ -243,7 +243,7 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
                // Composes can get deleted in composer, usually when the image is expired
                return echo.NewHTTPError(http.StatusNotFound, string(body))
        } else if resp.StatusCode != http.StatusOK {
-               httpError := echo.NewHTTPError(http.StatusInternalServerError, "Failed querying compose status")
+               httpError := echo.NewHTTPError(http.StatusInternalServerError, "Failed querying compose status (got status code: %v", resp.StatusCode)
                body, err := io.ReadAll(resp.Body)
                if err != nil {
                        ctx.Logger().Errorf("Unable to parse composer's compose response: %v", err)
```
(not strictly needed but nice to know what exactly went wrong on the other side (or if we don't want to put it in the error itself at least log it)) (this could be it's own PR already)

2. We probably want a simple test that returns the same http response as the failure so that we can target our retry against this test

3. We probably want to extract a helper to make the code easier to follow, maybe something like:
```go
diff --git a/internal/v1/handler.go b/internal/v1/handler.go
index aea637d..52f3c39 100644
--- a/internal/v1/handler.go
+++ b/internal/v1/handler.go
@@ -224,16 +224,11 @@ func (h *Handlers) GetPackages(ctx echo.Context, params GetPackagesParams) error
        })
 }
 
-func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error {
-       composeEntry, err := h.getComposeByIdAndOrgId(ctx, composeId)
-       if err != nil {
-               return err
-       }
-
+func (h *Handlers) getComposeStatusBodyWithRetry(ctx echo.Context, composeId uuid.UUID) ([]byte, error) {
        var resp *http.Response
        var body []byte
 
-       err = retry.Do(
+       err := retry.Do(
                func() error {
                        var err error
                        resp, err = h.server.cClient.ComposeStatus(composeId)
@@ -254,19 +249,31 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
        )
 
        if err != nil {
-               return echo.NewHTTPError(http.StatusInternalServerError, "Failed querying compose status after multiple attempts")
+               return nil, echo.NewHTTPError(http.StatusInternalServerError, "Failed querying compose status after multiple attempts")
        }
 
        switch resp.StatusCode {
        case http.StatusNotFound:
                // Composes can get deleted in composer, usually when the image is expired
-               return echo.NewHTTPError(http.StatusNotFound, string(body))
+               return nil, echo.NewHTTPError(http.StatusNotFound, string(body))
        case http.StatusOK:
-               // Continue processing
+               return body, nil
        default:
                httpError := echo.NewHTTPError(http.StatusInternalServerError, "Failed querying compose status")
                _ = httpError.SetInternal(fmt.Errorf("%s", body))
-               return httpError
+               return nil, httpError
+       }
+}
+
+func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error {
+       composeEntry, err := h.getComposeByIdAndOrgId(ctx, composeId)
+       if err != nil {
+               return err
+       }
+
+       body, err := h.getComposeStatusBodyWithRetry(ctx, composeId)
+       if err != nil {
+               return err
        }
 
        var composeRequest ComposeRequest
```
[note that the above diff does not address that the helper must check for the 5xx error codes (ServiceUnavailable is probably what we want to retry on initially).

this helper raises the question if we should generalize this as we have e.g. a similar request/http status handling when querying the metadata for a compose. And in the bigger picture, the handlers are all dealing with the resp.statusCode, one idea might be to move the handling of the status codes into the underlying cClient.ComposeStatus() functions and return errors there already if the status code is not 2xx - but that is a much bigger change of course and not for this PR!

internal/v1/handler.go
```diff
@@ -229,27 +230,40 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
...
+			defer closeBody(ctx, resp.Body)
+
+			body, err = io.ReadAll(resp.Body)
+			return err
+		},
+		retry.Attempts(3),
+		retry.Delay(time.Second),
+		retry.DelayType(retry.BackOffDelay),
+		retry.OnRetry(func(n uint, err error) {
+			ctx.Logger().Warnf("Attempt %d failed: %v. Retrying...", n+1, err)
```
mvo5, line 252: This log probably needs some more context, maybe incude the composeId here too (or the url)

mvo5 (commented):
Thank you for working on this! I added some comments/ideas/suggestions inline, I hope it's helpful.

mgold1234:
Closing it because there is better fix on the envoy level

PR #1268, opened 2024-07-12
internal: retry 503 errors from composer (HMS-4427)

mvo5:
Based on https://github.com/osbuild/image-builder/pull/1266 (thanks @mgold1234 !)

See the individual commit messages. The tests are a bit repetitive and there is room to extract more helper here i feel but it's a start :)

The PR should be complete now but it is as well tested as I would like it to be tested, e.g. postBodyWithRetry` has no test and the `getBodyWithRetry` tests are also not ideal as they pseudo test only ComposeStatus(). OTOH if this is moved into "[image-builder/internal/clients/composer/client.go](https://github.com/osbuild/image-builder/blob/d22e4de51ffa49722179da18106ef0519f7446db/internal/clients/composer/client.go#L76)" the tests could be added while moving this (see the discussion down there).


kingsleyzissou (commented):
Thanks for this :) It looks pretty sane to me, I think there are just a couple of linting issues

schuellerf:
I'd rather really push this down lower e.g. to https://github.com/osbuild/image-builder/blob/d22e4de51ffa49722179da18106ef0519f7446db/internal/clients/composer/client.go#L76

Any concerns?
(I'll check next week if it's possible )

mvo5:
> I'd rather really push this down lower e.g. to
> 
> https://github.com/osbuild/image-builder/blob/d22e4de51ffa49722179da18106ef0519f7446db/internal/clients/composer/client.go#L76
> 
> Any concerns? (I'll check next week if it's possible )

I looked at this too and my initial thinking was that this is the right place.  But then changed my mind because and went with  my current approach mostly because changing it at this level changes the nature of the client library and the code responsibilities. Right now client.go:request() is an extremely thin layer that does (almost) no policy decisions. It would become much more opinionated. I mostly shyed away because it does not even interprets http status codes yet and leaves all that to the caller of the library. So moving the retry at this level would change the responsibilities of the exiting code quite a bit.

Having said this, it may still be a good idea to change the responsibilities because in many ways client.go is a better place to make policy decisions about error and retires. It is just a "bigger" change (codewise it is probably smaller and testing will also be easier which is a good indicator that it might be a good idea). So I'm fine with that, most of the retry logic and tests should be easily movable.

internal/v1/handler.go
```diff
@@ -223,35 +227,114 @@ func (h *Handlers) GetPackages(ctx echo.Context, params GetPackagesParams) error
...
+			resp, err := f()
+			if err != nil {
+				if shouldRetry(err) {
+					return err
+				}
+				return retry.Unrecoverable(echo.NewHTTPError(http.StatusInternalServerError, fmt.Sprintf("Failed %v: %v (attempts: %v)", info, err, nRetries)))
+			}
+			defer closeBody(ctx, resp.Body)
+
+			body, err = io.ReadAll(resp.Body)
```
mvo5, line 283: There is maybe a subtle bug here - if we do a POST but fail to read the reply from composer we will retry - is this safe, i.e. will we possibly create multiple jobs just because composer could not sent us all the data (but created the job)? Probably needs checking on the composer side as well.

mvo5:
> > I'd rather really push this down lower e.g. to
> > https://github.com/osbuild/image-builder/blob/d22e4de51ffa49722179da18106ef0519f7446db/internal/clients/composer/client.go#L76
> > 
> > Any concerns? (I'll check next week if it's possible )
> 
> I looked at this too and my initial thinking was that this is the right place. But then changed my mind because and went with my current approach mostly because changing it at this level changes the nature of the client library and the code responsibilities. Right now client.go:request() is an extremely thin layer that does (almost) no policy decisions. It would become much more opinionated. I mostly shyed away because it does not even interprets http status codes yet and leaves all that to the caller of the library. So moving the retry at this level would change the responsibilities of the exiting code quite a bit.
> 
> Having said this, it may still be a good idea to change the responsibilities because in many ways client.go is a better place to make policy decisions about error and retires. It is just a "bigger" change (codewise it is probably smaller and testing will also be easier which is a good indicator that it might be a good idea). So I'm fine with that, most of the retry logic and tests should be easily movable.

I opened https://github.com/osbuild/image-builder/pull/1269 as a quick draft to play with the idea, may things to like, I would leave it to the original authors to decide, I think it makes a lot of sense to move more policy into the libra

> > I'd rather really push this down lower e.g. to
> > https://github.com/osbuild/image-builder/blob/d22e4de51ffa49722179da18106ef0519f7446db/internal/clients/composer/client.go#L76
> > 
> > Any concerns? (I'll check next week if it's possible )
[..]
> Having said this, it may still be a good idea to change the responsibilities because in many ways client.go is a better place to make policy decisions about error and retires. It is just a "bigger" change (codewise it is probably smaller and testing will also be easier which is a good indicator that it might be a good idea). So I'm fine with that, most of the retry logic and tests should be easily movable.

I thought about it some more and I did a quick draft in https://github.com/osbuild/image-builder/pull/1269 to see what moving it to this layer looks like. Lot's to like, if we go this way we probably want to move more policy about the http status code into the client library. Maybe something for @croissanne or @ondrejbudai who worked more on this to decide which direction. I like moving more into the client but it will require some work it seems (I put some comments into the other PR with open questions). Thanks again for the idea!

mvo5:
Closing in favor of the simpler fix on the envoy level

PR #1269, opened 2024-07-12
client: implement retries on the client request level (HMS-4427)

mvo5:
This is a quick draft of the idea of Florian (thanks!) about moving
the retry handlign into the client.go:request() helper.

What is nice about it is that it's very local. It changes the
responsibilities of the library though, because before client.go
did (for the most part) not care about the http status code and
all error are handled on the upper layers. By adding retry the
library layer now needs to make policy decisions based on the
status code. If we go this way we should move more of the
error handling into the library I think.

The other open question in this commit is how to handle retries
if reading the body fails. Here we need to also think about
the semantic of POST and make sure that the other side will
not create multiple jobs if we sent the POST correctly but
did not get the full reply.

And this needs *tests* - it's just a quick draft to get an
idea of the drawbacks/advantages compared to PR #1268


internal/clients/composer/client.go
```diff
@@ -73,7 +77,76 @@ func createClient(composerURL string, ca string) (*http.Client, error) {
...
+	}
+	return false
+}
+
+var errRetryFromStatusCode = fmt.Errorf("retry based on http status code")
+
 func (cc *ComposerClient) request(method, url string, headers map[string]string, body io.ReadSeeker) (*http.Response, error) {
+	retryableHttpCodes := []int{http.StatusServiceUnavailable}
+
+	var resp *http.Response
```
mvo5, line 96: this needs tests, probably very similar to the cases in https://github.com/osbuild/image-builder/pull/1268/files#diff-6850c3f59702d7b54878bb8eac65ad4a329f1332df31329d8623521f93326f8dR253

schuellerf:
regarding the "retryableHttpCodes"
https://github.com/envoyproxy/envoy/blob/v1.24.11/source/common/router/router.cc#L1226C1-L1228C70

internal/clients/composer/client.go
```diff
@@ -73,7 +77,76 @@ func createClient(composerURL string, ca string) (*http.Client, error) {
...
+	err := retry.Do(
+		func() (err error) {
+			resp, err = cc.requestOnce(method, url, headers, body)
```
croissanne, line 99: On each retry we should reseek body to 0.

mvo5:
Closing for now as this is done on the envoy level now.

PR #1272, opened 2024-07-15
internal/v1/handler: increase verbosity on connection errors

schuellerf:
Fourth implementation of a connection retry (successor of #1266, #1268, #1269)
Just increases verbosity which error code we got back. The real retry could be implemented as envoy proxy setting.


schuellerf:
Leaving as draft until I found out why all tests fail on my end and when I have some tests for those new cases

PR #1276, opened 2024-07-18
common: pass external id into composer

lzap:
This pull request updates the code to pass the external id into the composer. It also includes changes to the common package to handle request ids and insights request ids.

Requires this feature to be present in composer:

https://github.com/osbuild/osbuild-composer/pull/4047

Will work without any problems with old composer versions.

### DIFF
--- a/cmd/image-builder/logging.go
+++ b/cmd/image-builder/logging.go
@@ -1,21 +1,12 @@
 package main
 
 import (
-	"context"
 	"strings"
 
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/gommon/random"
 	"github.com/osbuild/image-builder/internal/common"
 	"github.com/sirupsen/logrus"
-)
-
-type ctxKey int
-
-const (
-	requestIdCtx         ctxKey = iota
-	insightsRequestIdCtx ctxKey = iota
-	requestDataCtx       ctxKey = iota
 )
 
 // Use request id from the standard context and add it to the message as a field.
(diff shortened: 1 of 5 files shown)


internal/common/context.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func WithRequestData(ctx context.Context, data logrus.Fields) context.Context {
+	return context.WithValue(ctx, requestDataCtx, data)
+}
+
+func RequestData(ctx context.Context) logrus.Fields {
+	rd := ctx.Value(requestDataCtx)
+	if rd == nil {
+		return nil
```
ezr-ondrej, line 48: What about returning empty logrus.Fields here?

internal/common/context.go
```diff
@@ -0,0 +1,51 @@
...
+type ctxKey int
+
+const (
+	requestIdCtx         ctxKey = iota
+	insightsRequestIdCtx ctxKey = iota
+	requestDataCtx       ctxKey = iota
+)
+
+func WithRequestId(ctx context.Context, requestId string) context.Context {
+	return context.WithValue(ctx, requestIdCtx, requestId)
```
ezr-ondrej, line 18: Opportunity for future: https://github.com/RedHatInsights/platform-go-middlewares/blob/44143f6300d86c94a78cdfc5ad5bbaf66b60d53d/request_id/request_id.go#L69C6-L69C25 ? :)

cmd/image-builder/logging.go
```diff
@@ -38,13 +29,12 @@ func (h *ctxHook) Fire(e *logrus.Entry) error {
...
-	e.Data["insights_id"] = e.Context.Value(insightsRequestIdCtx)
+	e.Data["request_id"] = common.RequestId(e.Context)
+	e.Data["insights_id"] = common.InsightsRequestId(e.Context)
 
-	if rd, ok := e.Context.Value(requestDataCtx).(logrus.Fields); ok {
-		for k, v := range rd {
-			e.Data[k] = v
-		}
+	rd := common.RequestData(e.Context)
+	for k, v := range rd {
```
ezr-ondrej, line 36: will this work in case the context value of `RequestData` was not set and `rd == nil`?

ezr-ondrej (commented):
Few nitpicks, great addition, would be amazing to have it in! :orange_heart:

internal/common/context.go
lzap: This must come with v2 upgrade first. I can probably do it.

internal/common/context.go
lzap: Sure.

cmd/image-builder/logging.go
lzap: Good catch, it is alwas set but not returning nil solves it nicely.

lzap:
Rebased.

ezr-ondrej (approved):
Thanks @lzap, I hope I've gone through it carefully enough, lets get it in :)

PR #1281, opened 2024-07-25
v1: handle repositories without snapshotting (HMS-4481)

croissanne:
If snapshots are requested with repositories that don't have snapshotting enabled, return 400.

### DIFF
--- a/internal/v1/handler_compose_image.go
+++ b/internal/v1/handler_compose_image.go
@@ -67,7 +67,7 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
 	customizations, err = h.buildCustomizations(ctx, composeRequest.Customizations, composeRequest.ImageRequests[0].SnapshotDate)
 	if err != nil {
 		ctx.Logger().Errorf("Failed building customizations: %v", err)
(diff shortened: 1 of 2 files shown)


jrusz (commented):
I seem to be getting a different 500 error now:
```
{
  "errors": [
    {
      "detail": "Unable to build customizations",
      "title": "500"
    }
  ]
}
```

jrusz (changes requested):
Something is still wrong, when I select a repo with snapshotting enabled I still get a 500 internal server error without any additional information. 

The only difference I noticed is that on the last step in the wizard it says that some of the repositories don't have a snapshot with the requested data but in that case it should use the one that's closest, right?

croissanne:
> Something is still wrong, when I select a repo with snapshotting enabled I still get a 500 internal server error without any additional information.
> 
> The only difference I noticed is that on the last step in the wizard it says that some of the repositories don't have a snapshot with the requested data but in that case it should use the one that's closest, right?
> 
> EDIT: Tried to select Todays date which does have the snapshot, warning is gone but 500 internal server error is still present.

I added a test now, so I really think it should be 400 now unless there's another bug somewhere.

PR #1283, opened 2024-07-29
api/v1: Allow users customization with password (HMS-4454)

andywaltlova:
[HMS-4454](https://issues.redhat.com/browse/HMS-4454)

Customization Users with either password, ssh_key or both are now allowed. Password is stored hashed in DB, every subsequent operation is done with hashed password. Same hashing function as in composer (SHA-512), hashed only if it was passed as plaintext.

From issue:

- [x] Verify /compose API accepts users Customizations with password (`TestComposeCustomizations`) 
- [x] Verify Blueprint API accepts users Customizations with password (`TestHandlers_CreateBlueprint`)
- [x] Verify `/blueprint/<id>/compose` builds image with password customizations (`TestHandlers_ComposeBlueprint`)
- [x] Verify password can not be fetched through API = is filtered in all endpoints:
  - [x] `/blueprint/<id>` (`TestHandlers_GetBlueprint`)
  - [x] `/blueprint/<id>/export` (`TestHandlers_ExportBlueprint`)
  - [x] ``/composes/<id>` the other doesn't have data about users)

pt.2

- [x] Filter out the password, meaning read on api will return some constant = redacted password
  - returns `"<REDACTED>"` (can be changed)
- [x] allow for easy update (nil for no password, same constant to not change the password, new password to set new)
  - Throws unprocessable entity if user uses  `"<REDACTED>"` as a password, otherwise works as usual
- [x] tests updated


internal/v1/handler_blueprints_test.go
```diff
@@ -387,6 +407,11 @@ func TestHandlers_GetBlueprint(t *testing.T) {
 	require.Equal(t, blueprint.Distribution, result.Distribution)
 	require.Equal(t, blueprint.Customizations, result.Customizations)
 
+	// Check that the password returned is nil
+	customizationUser, err := (*result.Customizations.Users)[0].AsUserPassword()
+	require.NoError(t, err)
+	require.Equal(t, nil, customizationUser.Password)
```
andywaltlova, line 413: Failing right now (TODO)

internal/v1/handler_blueprints_test.go
```diff
@@ -475,6 +514,11 @@ func TestHandlers_ExportBlueprint(t *testing.T) {
 	require.Equal(t, &Subscription{}, result.Customizations.Subscription)
 	require.Equal(t, &id, result.Metadata.ParentId)
 	require.NotEqual(t, metadata.ExportedAt, result.Metadata.ExportedAt)
+
+	// Check that the password returned is nil
+	customizationUser, err := (*result.Customizations.Users)[0].AsUserPassword()
+	require.NoError(t, err)
+	require.Equal(t, nil, customizationUser.Password)
```
andywaltlova, line 521: Failing right now (TODO)

internal/v1/api.yaml
```diff
@@ -2121,6 +2121,23 @@ components:
           type: integer
           description: Group id of the group to create (optional)
     User:
+      oneOf:
+        - $ref: '#/components/schemas/UserPassword'
+        - $ref: '#/components/schemas/UserKey'
```
andywaltlova, line 2126: This may be done better ?  I feel like with two schemas I introduced more logic which is needed in code to work with User (always interpret is as a user with either key or password). 
Would you prefer to just have one schema for user with ssh key and password and somewhere in code check that at least one is non empty? (which is basically what the schema should validate :sweat_smile: ). Since I don't know the best practice here I am just questioning what I did.

internal/v1/handler.go
```diff
@@ -258,6 +258,21 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
 	if err != nil {
 		return err
 	}
+	// Remove password from user customization
+	if composeRequest.Customizations != nil {
+		for i, u := range *composeRequest.Customizations.Users {
+			userPassword, err := u.AsUserPassword()
+			userPassword.Password = "REDACTED"
```
andywaltlova, line 265: Issue does specify to make it not readable again, I chose to keep the schema and just return constant indicating it was redacted, but it can be changed (e.g. to null or drop the password property?) :shrug:

internal/v1/api.yaml
ezr-ondrej: also having both is valid, which you can't do through one of.

internal/v1/handler.go
```diff
@@ -258,6 +259,40 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
 	if err != nil {
 		return err
 	}
+	// Remove password from user customization
+	// NOTE: Maybe not neccessary? The password was sent hashed in the first place
```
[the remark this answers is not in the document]
ezr-ondrej: This is not a good place to do it I think :)
We should mostly do it here IMO: https://github.com/osbuild/image-builder/blob/dad73f279a803bc4ec8fb3799b6a799e5b35eaf7/internal/v1/handler_compose_image.go#L67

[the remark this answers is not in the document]
andywaltlova: yeah thats what I did in https://github.com/osbuild/image-builder/blob/488b509c83a5da85a055122bd11b65daa077c4a9/internal/v1/handler_compose_image.go#L673-L709

but I will simplify it without the `oneOf` schema

ezr-ondrej:
> * (probably `/composes/<id>`? the other doesn't have data about users)

Ah, correct, I'll update the ticket, sorry! :)

internal/v1/api.yaml
andywaltlova: Now the validation is done in `CreateBlueprint`, schema requires only name for each User, but the custom validation needs either password or ssh_key

internal/v1/handler.go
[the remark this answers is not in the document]
andywaltlova: removed, since tests are still passing, password should be stored hashed if passed as plaintext and any subsequent operation like composing an image should only wokr with already hashed password.

internal/v1/api.yaml
```diff
@@ -2121,6 +2121,23 @@ components:
...
+        - name
+        - password
+      properties:
+        name:
+          type: string
+          example: "user1"
+        password:
+          type: string
+          format: password
+          example: "my-secret-password"
```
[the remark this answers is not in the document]
andywaltlova: It does, but not on schema or object level on current openapi version, since 3.1 it is allowed on object level (now I would have to give examples array on top level like a request or response).

I could update the openapi to 3.1 but there is still open https://github.com/oapi-codegen/oapi-codegen/issues/373, not ideal since the generation doesn't support it.

So either I can:

- leave it as it is (example of plain password mentioning in description it won't be hashed again) - imho good enough for me until the openapi version allows it
- or create elaborate examples on level of every `requestBody` that delas with Customizations

internal/v1/handler_blueprints.go
```diff
@@ -41,13 +42,43 @@ func BlueprintFromAPI(cbr CreateBlueprintRequest) BlueprintBody {
...
+			err = user.FromUserPassword(userPassword)
+			if err != nil {
+				return err
+			}
+			(*bb.Customizations.Users)[i] = user
+
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
andywaltlova: Not sure if I did it according to best practice, but I did create method for User to crypt password property and then the BlueprintBody has method that does it for user customization (calling the User function), and it's also called during write and read from database and during ComposeImage (post on /compose)

ezr-ondrej:
The api was already updated by bot: #1271

internal/v1/handler_blueprints_test.go
```diff
@@ -114,6 +129,12 @@ func TestHandlers_UpdateBlueprint(t *testing.T) {
 	body["name"] = "Changing to correct body"
 	respStatusCodeNotFound, _ := tutils.PutResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/blueprints/%s", uuid.New()), body)
 	require.Equal(t, http.StatusNotFound, respStatusCodeNotFound)
+
+	// Test update customization users
+	body["customizations"] = map[string]interface{}{"users": []map[string]interface{}{{"name": "test", "password": "test"}}}
+	statusCode, resp = tutils.PutResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/blueprints/%s", result.Id), body)
+	t.Log(resp)
```
ezr-ondrej, line 136: Don't foget to remove this ;)

andywaltlova: removed the log from my code and also one that was above it

andywaltlova:
@lzap @ezr-ondrej I think it's now ready for official review, oneOf is gone, validation done in handler and if I did everything right password should be saved in DB hashed.

internal/v1/handler_blueprints.go
```diff
@@ -48,6 +82,10 @@ func BlueprintFromEntry(be *db.BlueprintEntry) (BlueprintBody, error) {
 		return BlueprintBody{}, err
 	}
 
+	err = result.CryptPasswords()
```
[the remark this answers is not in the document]
andywaltlova: yeah I have same dillemma in the getComposeStatus, I could add it there as well but "it's too late" if it's in db in plaintext

internal/v1/handler_blueprints_test.go
```diff
@@ -317,6 +346,39 @@ func TestHandlers_GetBlueprintComposes(t *testing.T) {
...
+	t.Run("plain password", func(t *testing.T) {
+		body := []byte(`{"name": "Blueprint", "description": "desc", "customizations": {"users": [{"name": "user", "password": "foo"}]}, "distribution": "centos-9"}`)
+		be := &db.BlueprintEntry{
+			Body: body,
+		}
+		result, err := BlueprintFromEntry(be)
+		require.NoError(t, err)
+
+		require.NotEqual(t, common.ToPtr("foo"), (*result.Customizations.Users)[0].Password)
+	})
```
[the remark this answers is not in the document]
andywaltlova: Yeah we shouldn't but if I remove the check then we sort of rely on the code to never use the insert to DB directly. I mean the problem is that the insert itself inserts whatever we give, but the API layer, the handling functions take care of the hashing, so only legitimate test cases are create/update/export blueprint. In other cases I would have to change how the data are inserted into db or just work with hashed passwords :sweat_smile:

internal/v1/handler_blueprints.go
```diff
@@ -94,6 +136,20 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
 		desc = *blueprintRequest.Description
 	}
 
+	if blueprintRequest.Customizations.Users != nil {
```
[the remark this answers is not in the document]
andywaltlova: :shrug:  I am little scared to include more in this PR, but I am open to do it separately :D

internal/v1/handler_blueprints_test.go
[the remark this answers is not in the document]
ezr-ondrej: Gotcha, I think we can leave it in with comments, that works too :)

internal/v1/handler_get_compose_status_test.go
```diff
@@ -37,7 +38,19 @@ func TestComposeStatus(t *testing.T) {
 	require.NoError(t, err)
 	cr := ComposeRequest{
 		Distribution: "rhel-9",
-	}
+		Customizations: &Customizations{
+			// Since we are not calling handleCommonCompose but inserting directly to DB
+			// there is no point in using plaintext passwords.
+			// If there is plaintext password in DB there is problem elsewhere (eg. CreateBlueprint)
```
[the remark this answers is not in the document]
ezr-ondrej: I'd probably deal with it in a follow-up, IMO we want to filter the password out completely :thinking:

internal/v1/handler_blueprints.go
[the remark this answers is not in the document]
ezr-ondrej: Yeah I guess that make sense, tho will we be able to have a follow-up this week?
I'm bit reluctant to introduce this to prod without filtering the password out.
We could also have a comment on the password field it's experimental :shrug:

andywaltlova:
Talked with @ezr-ondrej and we decided to do it all in this PR in separate commits:

1. Allow for user with password, store it hashed (read returns also hashed password)
2. Filter out the password, meaning read on api will return some constant = redacted password, this will still allow for easy update (nil for no password, constant to not change the password, new password to set new)

second part of the PR should include updated tests (maybe new ones) to test properly the update and read on all endpoints.

@ezr-ondrej feel free to edit the description or jira ticket more to reflect the requirements

internal/v1/handler_blueprints.go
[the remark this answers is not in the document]
andywaltlova: Resolving this, we decided to do it in this PR just separate commits, the part one adding the possibility of password and hashing it is done, I will do the 'filtering out' as separate commit, we talked and it should just return some constant instead of hashed password (with a note to take care of update accordingly)

internal/clients/composer/openapi.v2.yml
```diff
@@ -1764,6 +1764,12 @@ components:
         key:
           type: string
           example: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAINrGKErMYi+MMUwuHaRAJmRLoIzRf2qD2dD5z0BTx/6x"
+        password:
+          type: string
+          description: |
+            If the password starts with $6$, $5$, or $2b$ it will be stored as
+            an encrypted password. Otherwise it will be treated as a plain text
+            password.
```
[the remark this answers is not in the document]
lzap: Composer uses Anaconda to do the installation which supports both plain text and hashed passwords.

The code I am seeing here does perform hashing of plain text passwords. Which is I think a good idea since blueprint is stored and being dragged among several services. Hashed passwords are safe to store.

internal/v1/api.yaml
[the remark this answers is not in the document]
lzap: This is fine, it was just a brain dump. Let’s leave that idea.

internal/v1/handler_blueprints.go
```diff
@@ -33,12 +34,45 @@ type BlueprintBody struct {
...
+	if bb.Customizations.Users != nil {
+		var usersWithHashedPasswords []User
+		for _, u := range *bb.Customizations.Users {
+			err := u.CryptPassword()
+			if err != nil {
+				return err
+			}
+			usersWithHashedPasswords = append(usersWithHashedPasswords, u)
+		}
+		bb.Customizations.Users = &usersWithHashedPasswords
```
lzap, line 77: The `usersWithHashedPasswords` is I assume a workaround around the fact that you could not see original values being changed. This is a [common problem](https://www.calhoun.io/does-range-copy-the-slice-in-go/) and more convenient way to fix that is this:

```
for i := range slice {
	slice[i].CallMethod()
}
```

Or was there other erason to copy all users?

internal/v1/handler_blueprints.go
```diff
@@ -39,6 +39,12 @@ func (u *User) CryptPassword() error {
 	if u.Password == nil || (len(*u.Password) == 0 || crypt.PasswordIsCrypted(*u.Password)) {
 		return nil
 	}
+
+	if u.IsRedacted() {
+		return errors.New(
+			"password is redacted and thus can't be hashed and stored in database, use plaintext password or already hashed password",
+		)
+	}
```
lzap, line 47: Sort of fails the use case when a user would really really want `<REDACTED>` as a password. A weird one, granted. Just saying, I like what I am seeing here. Nicely done.

internal/v1/handler_blueprints.go
```diff
@@ -47,6 +53,17 @@ func (u *User) CryptPassword() error {
...
+func (u *User) RedactPassword() {
+	redactedPassword := "<REDACTED>"
+	if u.Password != nil {
+		*u.Password = redactedPassword
+	}
+}
+
+func (u *User) IsRedacted() bool {
+	return u.Password == nil || *u.Password == "<REDACTED>"
+}
```
lzap, line 65: _Little bit of copying is worth it when avoiding a new dependency_ -- Rob Pike

👍🏻

internal/v1/handler_blueprints.go
```diff
@@ -62,6 +79,18 @@ func (bb *BlueprintBody) CryptPasswords() error {
...
+func (bb *BlueprintBody) RedactPasswords() {
+	if bb.Customizations.Users != nil {
+		var redactedUsers []User
+		for _, u := range *bb.Customizations.Users {
+			u.RedactPassword()
+			redactedUsers = append(redactedUsers, u)
+		}
+		bb.Customizations.Users = &redactedUsers
+	}
+}
```
lzap, line 91: Same comment as I was making in the other commit - you can do this without copying all the values.

lzap (commented):
Apologies, had some days off, planning more. Just a minor thing on documentation. And there is a workaround around range loop that can be replaced just by accessing the original slice directly.

internal/v1/handler_blueprints.go
andywaltlova: Yeah, it was supposed to be just a workaround, interesting, thanks for the article, I had something like that there initially but then the unit tests were failing, I probably just messed it up somehow :shrug: works now with the methods I have :+1:

internal/v1/handler_blueprints.go
andywaltlova: yep, fixed

internal/v1/api.yaml
```diff
@@ -1882,14 +1883,20 @@ components:
...
           example: "user1"
         ssh_key:
           type: string
           example: "ssh-rsa AAAAB3NzaC1"
+        password:
+          type: string
+          format: password
+          example: "plaintext-password"
+          description: |
+            The password will be hashed using the SHA-512 algorithm (if it's not hashed already).
```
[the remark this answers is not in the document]
andywaltlova: Alright added an example and adjusted the description

lzap (approved):

lzap:
_Do not hardcode passwords in code. Found hardcoded saved in passwordToHash._

Just mask it I guess, there could be a way to turn this rule off but I have no idea how.

ezr-ondrej:
I've just ignored it, IMO it's ok in test, so why bother workarounding it

internal/v1/handler_blueprints_test.go
```diff
@@ -122,6 +136,16 @@ func TestHandlers_UpdateBlueprint(t *testing.T) {
 	body["name"] = "Changing to correct body"
 	respStatusCodeNotFound, _ := tutils.PutResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/blueprints/%s", uuid.New()), body)
 	require.Equal(t, http.StatusNotFound, respStatusCodeNotFound)
+
+	// Test update customization users - invalid redacted password
+	body["customizations"] = map[string]interface{}{"users": []map[string]interface{}{{"name": "test", "password": "<REDACTED>"}}}
+	statusCode, _ = tutils.PutResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/blueprints/%s", result.Id), body)
```
ezr-ondrej, line 142: I thought this is ok, but this makes sure we do not override the password, but that actually means users can not update Blueprint without re-entering the password, right?

I think it's ok to have for now as MVP, but I assume that best case is:
When I update Blueprint with <redacted> as password, we pull the password from previous version and fail only if the password was not previously set.
I know it complicates the PR further, so I'm happy to just create a ticket for it and merge as is. Current solution is not best UX, but works and it's not a security concern :)

ezr-ondrej (approved):
Nice job! :+1:
One thing I 'm still unsure how should we handle updates :thinking: More inline, tho I'm happy to wave it from this PR and just raise a ticket for it :)

internal/v1/handler_blueprints_test.go
andywaltlova: I think separate issue would be nice, you can describe the use case there. I will be happy to implement it. Tbh I just though that whatever you give the update method it will just overwrites the blueprint, e.g. you had users and in update you just delete them. What if you want to change the users from password to ssh_key, that is allowed?

ezr-ondrej:
I'm good to merge, but I'll hold of for Monday so we can discuss my concern if someone disagree with my conclusion to merge regardles and also to test it on stage immediately after merge :)

internal/v1/handler_blueprints_test.go
ezr-ondrej: I've tried my best here: https://issues.redhat.com/browse/HMS-4562

ezr-ondrej:
Thank you @andywaltlova! Amazing work :orange_heart:

ezr-ondrej:
I hear no objections, lets get this in! :shipit:

PR #1287, opened 2024-07-31
api/v1: Allow fetching specific blueprint version (HMS-4512)

ezr-ondrej:
Allow fetch specific Blueprint version, to allow versioning diff in the UI.

### DIFF
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -115,14 +115,21 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
 	})
 }
 
-func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID) error {
+func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID, params GetBlueprintParams) error {
 	userID, err := h.server.getIdentity(ctx)
 	if err != nil {
 		return err
 	}
 
 	ctx.Logger().Infof("Fetching blueprint %s", id)
-	blueprintEntry, err := h.server.db.GetBlueprint(ctx.Request().Context(), id, userID.OrgID())
+	if params.Version != nil && *params.Version <= 0 {
+		if *params.Version != -1 {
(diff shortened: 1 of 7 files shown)


internal/v1/handler_blueprints.go
```diff
@@ -115,14 +115,28 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
...
+		*version = *params.Version
+		if *version < 0 {
+			v, err := h.server.db.GetLatestBlueprintVersionNumber(ctx.Request().Context(), userID.OrgID(), id)
+			if err != nil {
+				return err
+			}
+			// Adjust the version number -2 becomes latest - 1
+			*version = v + *params.Version + 1
+		}
+	}
```
lzap, line 137: There is no need to create a copy, you can modify params variable it is stored in the caller frame and will be thrown away.

If you prefer to be on the safe side tho, I suggest to change the `GetBlueprint` method to accept an integer value not pointer to avoid ton of dereferencing here.

On top of that, I think the SQL is written in a way that if you pass `nil` it should give you the latest version anyway. So there is no need to explicitly call `GetLatestBlueprintVersionNumber` at all, just document this and write a test for it.

lzap (changes requested):
Looks good, tho I think the code could be improved further.

internal/v1/handler_blueprints.go
ezr-ondrej: The `GetLatestBlueprintVersionNumber` is enabling me to have negative values. Having `-2` work.
It felt cool, why I did that, that's actually a good question :thinking:
I'm not very sure why I felt it should be implemented, it just felt cool, let me drop it xD

ezr-ondrej: Dropped and thus simplified a ton :)

ezr-ondrej:
@lzap rebased, could you give it another look? :)

internal/v1/handler_blueprints.go
lzap: No cool things please, let’s solve the problem :-P

lzap (approved):
Much better thanks.

PR #1288, opened 2024-08-01
floorist: add separate blueprint floorist query (HMS-4496)

avitova:
In case you feel like we need to keep the blueprints in the first query, we can do so.

### DIFF
--- a/templates/image-builder.yml
+++ b/templates/image-builder.yml
@@ -199,6 +199,15 @@ objects:
           left outer join blueprint_versions v on c.blueprint_version_id = v.id
           left outer join blueprints b ON v.blueprint_id = b.id
           cross join lateral jsonb_array_elements(c.request->'image_requests') as req;
+    - prefix: ${FLOORIST_QUERY_PREFIX}/blueprints
+      query: >-
+        select
+          v.blueprint_id::text as blueprint_id, v.version, b.deleted,
+          b.metadata->>'exported_at' as exported_at,
+          b.metadata->>'parent_id' as parent_id
+        from
+          blueprint_versions v
+          left outer join blueprints b ON v.blueprint_id = b.id
 
 - apiVersion: v1
   kind: Service


templates/image-builder.yml
```diff
@@ -183,22 +183,25 @@ objects:
     - prefix: ${FLOORIST_QUERY_PREFIX}/builds
       query: >-
         select
-          c.job_id,c.created_at,c.client_id,c.org_id,c.account_number,c.email,c.request->>'distribution' as distribution,
+          job_id,created_at,client_id,org_id,account_number,email,request->>'distribution' as distribution,
           req->>'architecture' as architecture,req->>'image_type' as image_type,req->'upload_request'->>'type' as upload_request_type,req->'ostree'->>'url' as ostree_url,
           req->>'snapshot_date' as snapshot_date,
-          v.blueprint_id::text as blueprint_id,
```
ezr-ondrej, line 189: hmm :thinking:
I'm thinking this is still pretty useful info, I'd probably keep both blueprint_id and version, so you can connect the dots.
"What blueprint and version this compose was build from" :)

It feels to me that we want to have both :)

ezr-ondrej (dismissed):

avitova:
Alright! So, I kept all the info we have for builds, including the version_id, blueprint_id, and a deleted_at. Even deleted_at might help us know more if we see that the blueprint was never built and was deleted.
So this just adds a new query now, and is rebased.

templates/image-builder.yml
```diff
@@ -199,6 +199,15 @@ objects:
           left outer join blueprint_versions v on c.blueprint_version_id = v.id
           left outer join blueprints b ON v.blueprint_id = b.id
           cross join lateral jsonb_array_elements(c.request->'image_requests') as req;
+    - prefix: ${FLOORIST_QUERY_PREFIX}/blueprints
+      query: >-
+        select
+          v.blueprint_id::text as blueprint_id, v.version, b.deleted
```
lzap, line 205:
```suggestion
          v.blueprint_id::text as blueprint_id, v.version, b.deleted,
```

I haven’t tried the query but I would expect a comma here?

templates/image-builder.yml
```diff
@@ -199,6 +199,15 @@ objects:
...
+          b.metadata->>'exported_at' as exported_at,
+          b.metadata->>'parent_id' as parent_id,
```
ezr-ondrej, line 207:
```suggestion
          b.metadata->>'parent_id' as parent_id
```

ezr-ondrej (commented):
What Lukas said +1 more nit, but looks good to me apart of that :)

(Just an anecdote: blueprint_versions join blueprints breaks my brain for some reason, I'd always do it the other way around for some reason)

avitova:
Of course, you are right, guys. I tested the query now, and this now would really work.

ezr-ondrej (approved):
Nice! :+1:

lzap (approved):

lzap:
Schulz is red

avitova:
Let's try after rebase:)

avitova:
Meh, this failed again, even though I do not think the failure is related. :thinking:

ezr-ondrej:
Yeah probably unrelated, let me click the retry button and lets see :)

ezr-ondrej:
Nice! Thanks! :orange_heart:

PR #1292, opened 2024-08-12
internal/v1: Improve error handling in `GetDistributions` method

mgold1234:
this commit imporve GetDisributions method by
- Add `filterDistributions` function to handle the filtering of restricted distributions.
- Enhance error messages to provide more context for debugging.
- Update error handling in `GetDistributions` to ensure proper HTTP status code is returned.

### DIFF
--- a/internal/v1/handler.go
+++ b/internal/v1/handler.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/osbuild/image-builder/internal/distribution"
+
 	"github.com/google/uuid"
 
 	"github.com/osbuild/image-builder/internal/clients/composer"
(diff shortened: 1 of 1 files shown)


internal/v1/handler.go
```diff
@@ -86,17 +88,25 @@ func (h *Handlers) GetDistributions(ctx echo.Context) error {
...
 	}
 
+	distributions, err := h.filterDistributions(userID.OrgID(), *dr)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, "Failed to filter distributions").SetInternal(err)
+	}
+	return ctx.JSON(http.StatusOK, distributions)
+}
+
+func (h *Handlers) filterDistributions(orgID string, distroMap distribution.DistroRegistry) (DistributionsResponse, error) {
```
ezr-ondrej, line 101: Nit: I'd probably keep using a pointer, just because we've got a pointer in the caller.
But given the DistroRegistry structure is just a single pointer, it's competely ok to copy it as well, so it's just a question of taste, nothing really important to address.

internal/v1/handler.go
```diff
@@ -86,17 +88,25 @@ func (h *Handlers) GetDistributions(ctx echo.Context) error {
 	dr := h.server.distroRegistry(ctx)
 	userID, err := h.server.getIdentity(ctx)
 	if err != nil {
-		return err
+		return echo.NewHTTPError(http.StatusInternalServerError, fmt.Sprintf("Failed to get user identity: %v", err))
 	}
 
+	distributions, err := h.filterDistributions(userID.OrgID(), *dr)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, "Failed to filter distributions").SetInternal(err)
```
ezr-ondrej, line 96: nit: just to give user extra context why we were filtering the distributions :)
```suggestion
		return echo.NewHTTPError(http.StatusInternalServerError, "Failed to filter available distributions").SetInternal(err)
```

internal/v1/handler.go
ezr-ondrej, line 101: This could also be a function, not a method if we pass the allowList as a parameter.
It would be sligtly cleaner in my opinion (we would not pollute the handlers object with helper methods), but again this is just a question of taste.
```suggestion
func filterDistributions(orgID string, distroMap distribution.DistroRegistry, allowList *<TheAllowListType>) (DistributionsResponse, error) {
```

ezr-ondrej (approved):
Nice improvements ;)
I've left few inline thoughts, all of them just nits, feel free to disregard :+1:

internal/v1/handler.go
mgold1234: this one look more complicated I prefer to keep it as is

internal/v1/handler.go
```diff
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/osbuild/image-builder/internal/distribution"
```
ezr-ondrej, line 14: This import is miss-ordered, it should be in the osbuild local imorts group, but it's a detail :)

ezr-ondrej (approved):
Nice clean-up, thank you :+1:

PR #1294, opened 2024-08-12
support insights compliance policies (HMS-4564)

croissanne:
Adds support for insights compliance policies, behind a feature flag.

For now the customisations remain hardcoded per profile ID (=ref ID), as the tailoring is quite hard to translate into which customisations should be enabled/disabled per profile ID.


lzap:
I am confused, what is this (prod):

image

internal/unleash/unleash.go
```diff
@@ -0,0 +1,38 @@
...
+		return fmt.Errorf("Unleash error: %w", err)
+	}
+
+	return nil
+}
+
+func Enabled(flag FeatureFlag) bool {
+	ctx := ucontext.Context{}
+	return unleash.IsEnabled(string(flag), unleash.WithContext(ctx), unleash.WithFallback(true))
+}
```
lzap, line 38: What we find very useful is to also send org id to unleash so you can turn on features per accounts.

croissanne:
> I am confused, what is this (prod):
> image

That's support for openscap _profiles_, but not via insights compliance. With insights compliance you can create _policies_ which are based on a profile, but are tailored, so they have specific rules enabled or disabled.

internal/clients/compliance/client.go
```diff
@@ -0,0 +1,139 @@
...
+	defer tailoringFileResp.Body.Close()
+
+	var tailoringData json.RawMessage
+	// returns 204 if there's no tailoring attached to the policy
+	if tailoringFileResp.StatusCode != 204 {
+		tailoringData, err = io.ReadAll(tailoringFileResp.Body)
+		if err != nil {
+			return nil, err
+		}
+	}
```
lzap, line 131: I think it is worth at least logging if code is not 204.

internal/v1/api.yaml
```diff
@@ -1942,20 +1942,36 @@ components:
...
+            The profile reference ID, either the profile_id or policy_id must be set.
         profile_name:
           type: string
-          description: "The policy type"
+          description: "The profile type"
         profile_description:
           type: string
-          description: "The longform policy description"
+          description: "The longform profile description"
     CustomRepository:
```
lzap, line 1975: I find documentation string confusing. I think there should be an overall text describing the difference between the two schemas, and then description for each one of them of what exactly the ID is and where you get it.

internal/v1/handler_compose_image.go
```diff
@@ -654,8 +659,53 @@ func (h *Handlers) buildCustomizations(ctx echo.Context, cust *Customizations, s
...
+			return nil, err
+		}
+
+		if policy.PolicyId != uuid.Nil {
+
+			if !unleash.Enabled(unleash.CompliancePolicies) {
+				return nil, echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Feature %s not enabled", string(unleash.CompliancePolicies)))
+			}
+
+			pdata, err := h.server.complianceClient.PolicyDataForMinorVersion(ctx.Request().Context(), 8, 10, policy.PolicyId.String())
```
lzap, line 684: Harcoded version (8.10) reminder. Was this intentional?

lzap (changes requested):
Few nitpicks. Haven’t tested tho but looks great.

internal/v1/handler_compose_image.go
croissanne: Oh woops :facepalm:  no this should be parsed from the distro.

internal/unleash/unleash.go
croissanne: Hmm interesting, how? Do you add it to the context somehow?

internal/clients/compliance/client.go
croissanne: Oh woops this is wrong, this should only be executed if the code _is_ 204.

croissanne: Hmm, it not being 204 is kind of the expected path, because that means the policy has tailorings attached.

internal/unleash/unleash.go
lzap: There are several options, edge for example uses the native "OrgId" field from the [Unleash Context](https://github.com/Unleash/unleash-client-go?tab=readme-ov-file#unleash-context) to pass this information. That is a reasonable approach.

internal/v1/handler_test.go
```diff
@@ -995,7 +995,6 @@ func TestGetCustomizations(t *testing.T) {
 				if result.Openscap == nil {
 					require.Nil(t, customizations.Openscap)
 				} else {
-					customizations.Openscap.ProfileId = string(profile)
```
lzap, line 998: Intended change?

lzap (approved):
LGTM needs rebase.

kingsleyzissou (approved):

internal/v1/handler_test.go
croissanne: Yea, I'm not sure why this was done in this way. Like it was setting the result before comparing it with the expected outcome… It was just a mock issue, somewhere the profile id's of the mock didn't match

internal/unleash/unleash.go
croissanne: Neat neat, mind if we do it in a followup? Definitely interesting.

croissanne:
IQE tests are unhappy due to a know dependency breakage :cry:

kingsleyzissou (approved):
:pray:

PR #1297, opened 2024-08-13
internal/v1: Skip crypt() tests on macOS

mgold1234:
this commit add checks in several tests to skip them when running on macOS (darwin) 
using `runtime.GOOS == "darwin"` 
this prevents the tests from failing on macOS since the `crypt()` function is not supported on this platform.


mvo5 (approved):
Thank you, this looks fine to me.

achilleas-k (dismissed):

PR #1298, opened 2024-08-13
(HMS-4562) Allow Blueprints update without changing User password

andywaltlova:
HMS-4562

See the issue for detailed acceptance criteria, in short the PR:

- [x] Blueprint can be updated without specifying the password, and password is not changed.
- [x] `<REDACTED>` value is not used at all, instead users password property is omitted from response
- [x] either password or ssh_key must be set for every user (or both)
- [x] empty string can be used to remove password or ssh_key (while still having either ssh_key or password set) == user can be changed from 'password user' to 'ssh_key user' and back

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -1883,7 +1883,10 @@ components:
       type: object
       required:
         - name
-        # One of (password, ssh_key) must be set, validator takes care of it
+      description: |
+        At least one of password, ssh_key must be set, validator takes care of it.
+        On update empty string can be used to remove password or ssh_key,
+        but at least one of them still must be present.
       properties:
         name:
           type: string
@@ -1897,6 +1900,8 @@ components:
           example: "$6$G91SvTj7uVp3xhqj$zVa8nqnJTlewniDII5dmvsBJnj3kloL3CXWdPDu9.e677VoRQd5zB6GKwkDvfGLoRR7NTl5nXLnJywk6IPIvS."
           description: |
             Plaintext passwords are also supported, they will be hashed and stored using the SHA-512 algorithm.
+            The password is never returned in the response.
+            Empty string can be used to remove the password during update but only with ssh_key set.
     Filesystem:
       type: object
       required:
--- a/internal/v1/handler.go
+++ b/internal/v1/handler.go
@@ -268,8 +268,9 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
 		return err
 	}
 	if composeRequest.Customizations != nil && composeRequest.Customizations.Users != nil {
-		for _, u := range *composeRequest.Customizations.Users {
-			u.RedactPassword()
+		users := *composeRequest.Customizations.Users
+		for i := range users {
+			users[i].RedactPassword()
 		}
 	}
 
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -25,6 +25,7 @@ import (
 
 var (
 	blueprintNameRegex         = regexp.MustCompile(`\S+`)
+	customizationUserNameRegex = regexp.MustCompile(`\S+`)
 	blueprintInvalidNameDetail = "The blueprint name must contain at least two characters."
 )
 
@@ -36,15 +37,10 @@ type BlueprintBody struct {
 
 func (u *User) CryptPassword() error {
 	// Prevent empty and already hashed password  from being hashed
-	if u.Password == nil || (len(*u.Password) == 0 || crypt.PasswordIsCrypted(*u.Password)) {
+	if u.Password == nil || len(*u.Password) == 0 || crypt.PasswordIsCrypted(*u.Password) {
 		return nil
 	}
 
-	if u.IsRedacted() {
-		return errors.New(
-			"password is redacted and thus can't be hashed and stored in database, use plaintext password or already hashed password",
-		)
-	}
 	pw, err := crypt.CryptSHA512(*u.Password)
 	if err != nil {
 		return err
@@ -53,15 +49,9 @@ func (u *User) CryptPassword() error {
 	return nil
 }
 
+// Set password to nil if it is not nil
 func (u *User) RedactPassword() {
-	redactedPassword := "<REDACTED>"
-	if u.Password != nil {
-		*u.Password = redactedPassword
-	}
-}
-
-func (u *User) IsRedacted() bool {
-	return u.Password == nil || *u.Password == "<REDACTED>"
+	u.Password = nil
 }
 
 func (bb *BlueprintBody) CryptPasswords() error {
@@ -82,6 +72,58 @@ func (bb *BlueprintBody) RedactPasswords() {
 			(*bb.Customizations.Users)[i].RedactPassword()
 		}
 	}
+}
+
+// Merges Password or SshKey from other User struct to this User struct if it is not set
+func (u *User) MergeExisting(other User) {
+	if u.Password == nil {
+		u.Password = other.Password
+	}
+	if u.SshKey == nil {
+		u.SshKey = other.SshKey
+	}
+}
+
+// User must have name and non-empty password or ssh key
+func (u *User) Valid() error {
+	validName := customizationUserNameRegex.MatchString(u.Name)
+	validPassword := u.Password != nil && len(*u.Password) > 0
+	validSshKey := u.SshKey != nil && len(*u.SshKey) > 0
+	if !validName || !(validPassword || validSshKey) {
+		return fmt.Errorf("User ('%s') must have a name and either a password or an SSH key set.", u.Name)
+	}
+	return nil
+}
+
+func (u *User) MergeForUpdate(userData []User) error {
+	// If both password and ssh_key in request user we don't need to fetch user from DB
+	if !(u.Password != nil && len(*u.Password) > 0 && u.SshKey != nil && len(*u.SshKey) > 0) {
+		eui := slices.IndexFunc(userData, func(eu User) bool {
+			return eu.Name == u.Name
+		})
+
+		if eui == -1 { // User not found in DB
+			err := u.Valid()
+			if err != nil {
+				return err
+			}
+		} else {
+			u.MergeExisting(userData[eui])
+		}
+	}
+
+	// If there is empty string in password or ssh_key, it means that we should remove it (set to nil)
+	if u.Password != nil && *u.Password == "" {
+		u.Password = nil
+	}
+	if u.SshKey != nil && *u.SshKey == "" {
+		u.SshKey = nil
+	}
+
+	if err := u.Valid(); err != nil {
+		return err
+	}
+	return nil
 }
 
 // Util function used to create and update Blueprint from API request (WRITE)
@@ -105,7 +147,14 @@ func BlueprintFromEntry(be *db.BlueprintEntry) (BlueprintBody, error) {
 	if err != nil {
 		return BlueprintBody{}, err
 	}
+	return result, nil
+}
 
+func BlueprintFromEntryWithRedactedPasswords(be *db.BlueprintEntry) (BlueprintBody, error) {
+	result, err := BlueprintFromEntry(be)
+	if err != nil {
+		return BlueprintBody{}, err
+	}
 	result.RedactPasswords()
 	return result, nil
 }
@@ -118,16 +167,6 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
 
 	var blueprintRequest CreateBlueprintRequest
 	err = ctx.Bind(&blueprintRequest)
-	if err != nil {
-		return err
-	}
-
-	blueprint, err := BlueprintFromAPI(blueprintRequest)
-	if err != nil {
-		return err
-	}
-
-	body, err := json.Marshal(blueprint)
 	if err != nil {
 		return err
 	}
@@ -157,18 +196,29 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
 		desc = *blueprintRequest.Description
 	}
 
-	if blueprintRequest.Customizations.Users != nil {
-		for _, user := range *blueprintRequest.Customizations.Users {
+	users := blueprintRequest.Customizations.Users
+	if users != nil {
+		for _, user := range *users {
 			// Make sure every user has either ssh key or password set
-			if user.Password == nil && user.SshKey == nil {
+			if err := user.Valid(); err != nil {
 				return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
 					Errors: []HTTPError{{
 						Title:  "Invalid user",
-						Detail: "User must have either a password or an SSH key set.",
+						Detail: err.Error(),
 					}},
 				})
 			}
 		}
+	}
+
+	blueprint, err := BlueprintFromAPI(blueprintRequest)
+	if err != nil {
+		return err
+	}
+
+	body, err := json.Marshal(blueprint)
+	if err != nil {
+		return err
 	}
 
 	err = h.server.db.InsertBlueprint(ctx.Request().Context(), id, versionId, userID.OrgID(), userID.AccountNumber(), blueprintRequest.Name, desc, body, metadata)
@@ -214,7 +264,7 @@ func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID, params 
 		return err
 	}
 
-	blueprint, err := BlueprintFromEntry(blueprintEntry)
+	blueprint, err := BlueprintFromEntryWithRedactedPasswords(blueprintEntry)
 	if err != nil {
 		return err
 	}
@@ -246,7 +296,7 @@ func (h *Handlers) ExportBlueprint(ctx echo.Context, id openapi_types.UUID) erro
 		return err
 	}
 
-	blueprint, err := BlueprintFromEntry(blueprintEntry)
+	blueprint, err := BlueprintFromEntryWithRedactedPasswords(blueprintEntry)
 	if err != nil {
 		return err
 	}
@@ -277,6 +327,37 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
 		return err
 	}
 
+	if !blueprintNameRegex.MatchString(blueprintRequest.Name) {
+		return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
+			Errors: []HTTPError{{
+				Title:  "Invalid blueprint name",
+				Detail: blueprintInvalidNameDetail,
+			}},
+		})
+	}
+
+	if blueprintRequest.Customizations.Users != nil {
+		be, err := h.server.db.GetBlueprint(ctx.Request().Context(), blueprintId, userID.OrgID(), nil)
+		if err != nil {
+			return err
+		}
+		eb, err := BlueprintFromEntry(be)
+		if err != nil {
+			return err
+		}
+		for i := range *blueprintRequest.Customizations.Users {
+			err := (*blueprintRequest.Customizations.Users)[i].MergeForUpdate(*eb.Customizations.Users)
+			if err != nil {
+				return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
+					Errors: []HTTPError{{
+						Title:  "Invalid user",
+						Detail: err.Error(),
+					}},
+				})
+			}
+		}
+	}
+
 	blueprint, err := BlueprintFromAPI(blueprintRequest)
 	if err != nil {
 		return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
@@ -286,18 +367,10 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
 			}},
 		})
 	}
+
 	body, err := json.Marshal(blueprint)
 	if err != nil {
 		return err
-	}
-
-	if !blueprintNameRegex.MatchString(blueprintRequest.Name) {
-		return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
-			Errors: []HTTPError{{
-				Title:  "Invalid blueprint name",
-				Detail: blueprintInvalidNameDetail,
-			}},
-		})
 	}
 
 	versionId := uuid.New()
@@ -335,7 +408,7 @@ func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) err
 	if err != nil {
 		return err
 	}
-	blueprint, err := BlueprintFromEntry(blueprintEntry)
+	blueprint, err := BlueprintFromEntryWithRedactedPasswords(blueprintEntry)
 	if err != nil {
 		return err
 	}
--- a/internal/v1/handler_blueprints_test.go
+++ b/internal/v1/handler_blueprints_test.go
@@ -95,6 +95,363 @@ func TestHandlers_CreateBlueprint(t *testing.T) {
 	require.Equal(t, "Invalid user", jsonResp.Errors[0].Title)
 }
 
+func TestUser_MergeForUpdate(t *testing.T) {
+	tests := []struct {
+		name          string
+		newUser       User
+		existingUsers []User
+		wantPass      *string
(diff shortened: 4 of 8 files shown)


internal/v1/handler_blueprints.go
```diff
@@ -300,6 +300,23 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
...
+	if blueprintRequest.Customizations.Users != nil {
+		for _, user := range *blueprintRequest.Customizations.Users {
+			// Make sure every user has either ssh key or password set
+			if user.Password == nil && user.SshKey == nil {
+				return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
+					Errors: []HTTPError{{
+						Title:  "Invalid user",
+						Detail: "User must have either a password or an SSH key set.",
+					}},
+				})
```
ezr-ondrej, line 312: I definitely don't insist, but do you think it's worth starting a `validateBlueprint` function?
IMO it then could be shared for Create/Update.

internal/v1/handler_blueprints.go
```diff
@@ -300,6 +300,23 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
...
+			}
+			// TODO: User password "<REDACTED>"
+			// 1. Previous password exists = set old password
```
ezr-ondrej, line 315: I guess we'll need to do this before marshaling the body.
Fetching the "old" Blueprint from DB should [be easy by db.GetBlueprint](https://github.com/osbuild/image-builder/blob/e8780a2796b470aa5c5f9e8b44c2c68af60dd52b/internal/db/db.go#L79).

Then just a bit tedious for going through both the structures and comparing user by user, but I hope it shouldn't be that bad :thinking:

ezr-ondrej (commented):
Proper **TDD** (tests already done before code) :orange_heart:  it!

internal/v1/handler_blueprints.go
andywaltlova: Yep I want to, since it wasn't trivial change my logic was to

1. draft tests first
2. make functional version
3. make it pretty (the validation function sounds great, maybe even separate one for customization validation)

and now I just pushed second step :)

andywaltlova:
Current logic:

- During update user can
  - use `"<REDACTED>" as a value - this will try to use latest password or throws an error
  - omit password in request (=nil)
    - this keeps password as it is in database
    - if ssh_key passed it will just add ssh_key as well
  - use ""
    - this will remove non-empty password in database (can't use nil because that one just keeps old password)
    - is valid only if ssh_key is non-empty (in request or already in database)

internal/v1/handler_blueprints.go
```diff
@@ -160,11 +167,20 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
...
 				return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
 					Errors: []HTTPError{{
 						Title:  "Invalid user",
-						Detail: "User must have either a password or an SSH key set.",
+						Detail: fmt.Sprintf("User ('%s') must have either a password or an SSH key set.", user.Name),
+					}},
+				})
+			}
+			if user.IsRedacted() {
+				// NOTE: we could silenty set password to empty if ssh key is available
```
andywaltlova, line 179: Not sure about this, I would prefer the error to not asume something, frontend can correctly validate this edge case (enforce non-empty ssh_key or password)

internal/v1/handler_blueprints.go
```diff
@@ -300,6 +316,64 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
...
+			return err
+		}
+		latestBlueprint, err := BlueprintFromEntry(blueprintEntry)
+		if err != nil {
+			return err
+		}
+		existingUsersMap := make(map[string]User)
+		for _, user := range *latestBlueprint.Customizations.Users {
+			existingUsersMap[user.Name] = user
+		}
```
andywaltlova, line 332: extract this to function and call it only if needed

internal/v1/handler_blueprints_test.go
```diff
@@ -95,6 +95,149 @@ func TestHandlers_CreateBlueprint(t *testing.T) {
...
+	updatedBlueprint, err = BlueprintFromEntryWithRedactedPasswords(blueprintEntry)
+	require.NoError(t, err)
+	require.NotNil(t, (*updatedBlueprint.Customizations.Users)[0].Password)
+	require.True(t, (*updatedBlueprint.Customizations.Users)[0].IsRedacted())
+
+	updatedBlueprint, err = BlueprintFromEntry(blueprintEntry)
+	require.NoError(t, err)
+	require.NotNil(t, (*updatedBlueprint.Customizations.Users)[0].Password)
+	require.NotEmpty(t, *(*updatedBlueprint.Customizations.Users)[0].Password)
+	require.NotEqual(t, "test", *(*updatedBlueprint.Customizations.Users)[0].Password) // still hashed
```
andywaltlova, line 224: improve this check, have to assert the previous password, imho notEqual is not enough

internal/v1/handler_blueprints_test.go
```diff
@@ -95,6 +95,149 @@ func TestHandlers_CreateBlueprint(t *testing.T) {
...
+	require.Equal(t, "$6$foo", *(*updatedBlueprint.Customizations.Users)[0].Password)
+
+	// password == nil tries to remove password  == FAIL because of required fields
+	body["customizations"] = map[string]interface{}{"users": []map[string]interface{}{{"name": "test"}}}
+	statusCode, responseBody = tutils.PutResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/blueprints/%s", result.Id), body)
+	require.Equal(t, http.StatusUnprocessableEntity, statusCode)
+	err = json.Unmarshal([]byte(responseBody), &jsonResp)
+	require.NoError(t, err)
+	require.Equal(t, "Invalid user", jsonResp.Errors[0].Title)
+	require.Equal(t, "User ('test') must have either a password or an SSH key set.", jsonResp.Errors[0].Detail)
```
andywaltlova, line 169: This shouldn't be fail, keep old password an ssh_key still nil

lzap:
> Current logic:

It sounds a little bit complicated and while I agree with some points, here is a suggestion for easy to understand behavior:

* Omitted field = do nothing. This is very well understood API REST behavior overall.
* Empty string = an error like 4xx Not Allowed. It is a good practice for systems to not to allow empty passwords, these are more often than not an unintentional input error.
* Anything else, including the literal `<REDACTED>`, should actually hash the string and store the hashed value. The special `<REDACTED>` string should have no special treatment.

andywaltlova:
> > Current logic:
> 
> It sounds a little bit complicated and while I agree with some points, here is a suggestion for easy to understand behavior:
> 
> * Omitted field = do nothing. This is very well understood API REST behavior overall.
> * Empty string = an error like 4xx Not Allowed. It is a good practice for systems to not to allow empty passwords, these are more often than not an unintentional input error.
> * Anything else, including the literal `<REDACTED>`, should actually hash the string and store the hashed value. The special `<REDACTED>` string should have no special treatment.

but then how will you be able to update the user in the blueprint? E.g. what if I want keep the user names and remove the passwords in favor of ssh keys? One of the obvious solutions would be to just instruct the users to remove the whole user customizations and then recreate it or to just create new blueprint, if thats fine then I think we can simplify it as you suggested :)

lzap:
Good point, then I am thinking using the empty password case as "remove password" behavior.

andywaltlova:
> Good point, then I am thinking using the empty password case as "remove password" behavior.

Yeah thats what is there right now, but since the customization is saved as an json raw message, in database that means it will equal to empty string and not NULL, essentially I don't think that's bad, but thats why the code is more complicated because I have to count with possibility that if someone removes password it won't be null but can be also empty string and then make sure every time that at least one of ssh_key or password is not empty (not null and not empty string)

Just to be clear, the suggestion you have about <REDACTED> value, that I can change to just hash it and it will be the new user password but I did not because that was explicitly one of the criteria in jira issue written by @ezr-ondrej :

> Passing <REDACTED> or another special value as user password, should ensure the current password is kept on update.

So if just not passing password counts toward this criteria (password=nil), it will just keep password as it was and I can remove the <REDACTED> to also keep the password it will just be used as a password.

lzap:
Apologies, I am back from my vacation. I have to say, I do not like the `<REDACTED>` approach and I have a different proposal. Let me paste it here so we can discuss it:

Problem definition:

A User in blueprint customization is an element with an unique name and password or ssh key (not both). This should be enforced, entries with both should return an error. The only required item in the schema is name, so the schema allows a user without both password and ssh key.

Passwords should be always protected using crypt POSIX function, unless this was already done which can be detected by the dollar prefix. This should be done automatically by the backend. SSH key is public, therefore there is no problem sharing this, however, passwords are sensitive and they should never be shared (returned by the API), even in the encrypted (hashed) form.

Solution:

The API would return users without password element (set to nil). In case of password user, the result would look like: {"name": "lzap"}(this is a departure from the initial idea the app should return <REDACTED> text).

When update UI is opened, the UI will render all users and since there will be no ssh key or password, these fields will be presented blank. On submit, blank fields will not be serialized to JSON for the update, meaning the same user JSON as above will be sent (just a username). The backend then needs to iterate over existing users (name is unique), pair all matches and load passwords from the database.

Therefore I suggest the following behavior for the password field:

 * No (null) password - keep previous password.
 * Empty password - delete password (must be combined with an ssh key or error is returned).
 * Some password - overwrite password optionally hashing it.

Examples:

 * Create user: {"name": "lzap", "password": "dog8code"} should hash the password and store the data.
 * Update user: {"name": "lzap", "password": "newpw"} should hash the password and store the data overwriting what was previously in the database.
 * Update user: {"name": "lzap"} should fetch the password from database and merge it with the new data, store it. Password is unchanged.
 * Update user: {"name": "lzap", "password": "", "ssh_key": "xxx"} should remove the password from db essentially converting password-based user to key-based user.
 * Update user: {"name": "lzap", "password": ""} should return an error if existing user is a password-based user.
 * Update user: {"name": "lzap", "password": "newpw", "ssh_key": ""} should remove the ssh key and hash new password essentially converting from key-based user to password-based user.

These examples is not a complete list of all possible options, that should be covered with unit tests. We need a test for every single possibility, this is a security-related code and it needs to be well tested.

lzap:
_Just for the record, we are elaborating on the design trying to explore alternative solutions to the problem. We will report back once we agree on something._

internal/v1/handler_blueprints.go
```diff
@@ -277,6 +320,78 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
...
+			// Make sure every user has either ssh key or password set
+			passwordAndSSHKeyEmptyStringRequest := (requestUser.Password != nil && len(*requestUser.Password) == 0) && (requestUser.SshKey != nil && len(*requestUser.SshKey) == 0)
+			if passwordAndSSHKeyEmptyStringRequest {
+				return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
+					Errors: []HTTPError{{
+						Title:  "Invalid user",
+						Detail: fmt.Sprintf("User ('%s') must have either a password or an SSH key set.", requestUser.Name),
+					}},
+				})
+			}
```
lzap, line 345: As stated above, this check should be: at least one is present.

Also, I am seeing this code for the 2nd time now, this code should be de-duplicated.

internal/v1/handler_blueprints.go
```diff
@@ -277,6 +320,78 @@ func (h *Handlers) UpdateBlueprint(ctx echo.Context, blueprintId uuid.UUID) erro
...
+
+				// If we are here, it means that user has one of password or ssh_key set to non-empty value
+				// If there is empty string in password or ssh_key, it means that we should remove it (set to nil)
+				if requestUser.Password != nil && len(*requestUser.Password) == 0 {
+					(*blueprintRequest.Customizations.Users)[i].RedactPassword()
+				}
+				if requestUser.SshKey != nil && len(*requestUser.SshKey) == 0 {
+					(*blueprintRequest.Customizations.Users)[i].RedactSshKey()
+				}
+			}
```
lzap, line 391: Okay, this is quite hard to read thanks to the wide ident. Let me first think about this part how I would write it; for each requestUser (`u`):

* Validate `u` first for invalid entries.
* Try to find existing user in the database using `slices.Contains` or `ContainsFunc`, store it as `ue`.
* If it does not exist, skip to the next user (no changes required).
* If it does exist, call the proposed `u.MergeUser(ue)` which will overwrite password/key if it is `nil`.
* Replace empty password/key (blank string) with `nil` (effectively deleting them).
* Validate `u` once again just in case the changes would create an invalid user.

Well, honestly it is tough to review your code it is quite complex. I suggest to clean it up, in general:

* Create a separate method which will accept the `User` type.
* Reuse the validation code.
* Try to explore the "merging" idea when there is no need to do prior checks, that should simplify it.
* Avoid creating hash to lookup existing records.
* And most importantly, write unit tests for the newly created function, it will be much easier to test a small method than testing the whole HTTP handler.

lzap (changes requested):
Here is my initial review, I think the most important thing is to isolate the most complex logic into a separate function that takes a pointer to `User` type and modifies it accordingly. This way, you will be able to easily write a tabular tests for all possible test cases. This will simplify the overall HTTP handler tests where you only need to cover positive test case.

internal/v1/api.yaml
```diff
@@ -1884,6 +1884,7 @@ components:
       required:
         - name
         # One of (password, ssh_key) must be set, validator takes care of it
```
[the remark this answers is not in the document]
andywaltlova: added validation method and moved the comment to description

internal/v1/handler_blueprints.go
```diff
@@ -53,15 +48,17 @@ func (u *User) CryptPassword() error {
...
 	}
 }
 
-func (u *User) IsRedacted() bool {
-	return u.Password == nil || *u.Password == "<REDACTED>"
+func (u *User) RedactSshKey() {
+	if u.SshKey != nil {
+		u.SshKey = nil
+	}
 }
```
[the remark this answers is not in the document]
andywaltlova: it was used for removing it in case of empty string, bad name and it does't have to be method, it's gone now

internal/v1/handler_blueprints.go
```diff
@@ -84,6 +81,32 @@ func (bb *BlueprintBody) RedactPasswords() {
 	}
 }
 
+func (u *User) setPassword(password *string) {
```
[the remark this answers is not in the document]
andywaltlova: Decided to go with `MergeExisting`, it works as described in below comment

internal/v1/handler_blueprints.go
```diff
@@ -84,6 +81,32 @@ func (bb *BlueprintBody) RedactPasswords() {
...
+	if password != nil {
+		if u.Password == nil {
+			u.Password = new(string)
+		}
+		*u.Password = *password
+	}
+}
```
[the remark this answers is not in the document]
andywaltlova: nice improvement, applied

internal/v1/handler_blueprints.go
```diff
@@ -160,17 +180,27 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
...
 		for _, user := range *blueprintRequest.Customizations.Users {
 			// Make sure every user has either ssh key or password set
-			if user.Password == nil && user.SshKey == nil {
+			if (user.Password == nil || len(*user.Password) == 0) && (user.SshKey == nil || len(*user.SshKey) == 0) {
 				return ctx.JSON(http.StatusUnprocessableEntity, HTTPErrorList{
 					Errors: []HTTPError{{
 						Title:  "Invalid user",
-						Detail: "User must have either a password or an SSH key set.",
+						Detail: fmt.Sprintf("User ('%s') must have either a password or an SSH key set.", user.Name),
 					}},
```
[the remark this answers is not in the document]
andywaltlova: de-duplicated

internal/v1/handler_blueprints.go
```diff
@@ -81,30 +75,67 @@ func (bb *BlueprintBody) RedactPasswords() {
...
 
-func (u *User) setSshKey(sshKey *string) {
-	if sshKey != nil {
-		if u.SshKey == nil {
-			u.SshKey = new(string)
-		}
-		*u.SshKey = *sshKey
+// User must have name and non-empty password or ssh key
+func (u *User) hasValidValues() error {
+	nameRe := regexp.MustCompile(`\S+`)
```
lzap, line 90: Nitpick: If you compile regexp into package-level variable, it is only compiled once during application start. Here, it is compiled every time a user is validated, which is a tad slower.

internal/v1/handler_blueprints.go
```diff
@@ -81,30 +75,67 @@ func (bb *BlueprintBody) RedactPasswords() {
...
+	validName := nameRe.MatchString(u.Name)
+	validPassword := u.Password != nil && len(*u.Password) > 0
+	validSshKey := u.SshKey != nil && len(*u.SshKey) > 0
+	isValid := validName && (validPassword || validSshKey)
+	if !isValid {
```
lzap, line 95: What’s up with this short-lived variables you use everywhere? This could have been simply:

```
if !(validName && (validPassword || validSshKey)) {
```

internal/v1/handler_blueprints.go
```diff
@@ -81,30 +75,67 @@ func (bb *BlueprintBody) RedactPasswords() {
...
-	return u.SshKey != nil && len(*u.SshKey) > 0
+		if eui == -1 { // User not found in DB
+			err := u.hasValidValues()
+			if err != nil {
+				return err
+			}
+		} else {
+			u.MergeExisting((*eb.Customizations.Users)[eui])
+		}
+	}
```
lzap, line 125: Oh right, I meant `IndexFunc` not `ContainsFunc` nice.

internal/v1/handler_blueprints.go
```diff
@@ -81,30 +75,67 @@ func (bb *BlueprintBody) RedactPasswords() {
...
+
+	// If there is empty string in password or ssh_key, it means that we should remove it (set to nil)
+	if u.Password != nil && len(*u.Password) == 0 {
```
lzap, line 128: Nitpick: this is more common in Go (the same below):

```suggestion
	if u.Password != nil && *u.Password == "" {
```

internal/v1/handler_blueprints.go
```diff
@@ -81,30 +75,67 @@ func (bb *BlueprintBody) RedactPasswords() {
...
+	if !isValid {
+		return fmt.Errorf("User ('%s') must have a name and either a password or an SSH key set.", u.Name)
 	}
+	return nil
 }
 
-func (u *User) hasPassword() bool {
-	return u.Password != nil && len(*u.Password) > 0
-}
+func (u *User) isValidForUpdate(h *Handlers, ctx echo.Context, blueprintId uuid.UUID, userID *Identity) error {
```
lzap, line 101: This method has side effect and it is very important to advertise such thing via a good method name. This name implies that it just tests for validity: `isValidForUpdate` while it actually modifies the receiver. I suggest to come up with a better name, perhaps since the main modification is called "merge" it could be also featured in the new name.

internal/v1/handler_blueprints.go
```diff
@@ -81,30 +75,67 @@ func (bb *BlueprintBody) RedactPasswords() {
...
+	// If both password and ssh_key in request user we don't need to fetch user from DB
+	if !(u.Password != nil && len(*u.Password) > 0 && u.SshKey != nil && len(*u.SshKey) > 0) {
+		be, err := h.server.db.GetBlueprint(ctx.Request().Context(), blueprintId, userID.OrgID(), nil)
+		if err != nil {
+			return err
+		}
+		eb, err := BlueprintFromEntry(be)
+		if err != nil {
+			return err
+		}
```
lzap, line 111: This database call, which is the most expensive operation for the whole request (few milliseconds on AWS EC2) is done in a loop for every single user. This is unnecessary, please call this just once outside of the loop and pass the `Users` pointer into the function.

internal/v1/handler.go
```diff
@@ -268,8 +268,8 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
 		return err
 	}
 	if composeRequest.Customizations != nil && composeRequest.Customizations.Users != nil {
-		for _, u := range *composeRequest.Customizations.Users {
-			u.RedactPassword()
+		for i := range *composeRequest.Customizations.Users {
+			(*composeRequest.Customizations.Users)[i].RedactPassword()
 		}
 	}
```
lzap, line 274: Out of curiosity - why this change? I do not have my IDE opened but `Users` is a pointer I assume. Therefore, even when `u` is a copy it should not matter. Does it work when you leave the shorter version?

But yeah, sometimes you need to do this in Go since `u` is a copy. To come up with a shorter version this is cleaner:

```
users := *composeRequest.Customizations.Users
for i := range users {
	users[i].RedactPassword()
}
```

internal/v1/handler_blueprints.go
```diff
@@ -53,15 +48,17 @@ func (u *User) CryptPassword() error {
...
 }
 
+// Set password to nil if it is not nil
 func (u *User) RedactPassword() {
-	redactedPassword := "<REDACTED>"
 	if u.Password != nil {
-		*u.Password = redactedPassword
+		u.Password = nil
 	}
 }
```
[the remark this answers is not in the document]
lzap: What I really meant was this:

```
func (u *User) RedactPassword() {
	// no need to do this: if u.Password != nil
	u.Password = nil
}
```

If password is `nil` and you set it to `nil` it is faster and two lines shorter than:

* Check if it is `nil`
* Set it to `nil`

In other words: redacting is operation that simply removes a password before we return the data to the user, no matter what the original value was.

internal/v1/handler_blueprints.go
```diff
@@ -84,6 +75,69 @@ func (bb *BlueprintBody) RedactPasswords() {
...
+	if u.Password == nil {
+		u.Password = other.Password
+	}
+	if u.SshKey == nil {
+		u.SshKey = other.SshKey
+	}
+}
+
+// User must have name and non-empty password or ssh key
+func (u *User) hasValidValues() error {
```
lzap, line 89:
```suggestion
func (u *User) isValid() error {
```

or even more Go native:

```suggestion
func (u *User) Valid() error {
```

There is no point not exporting this function I think, it is a safe one to export.

lzap (changes requested):
Ok this looks awesome, few small nitpicks and one big ask.

I would love to see a [table-driven test](https://go.dev/wiki/TableDrivenTests) for method called `isValidForUpdate` (I actually ask you to rename it). Test for various inputs we covered in the Jira epic description.

lzap:
Correction: it looks I missed the handler test, it covers all important cases. Feel free to ignore my comment about a new TDD test for the newly created function. Not needed.

internal/v1/handler.go
andywaltlova: yes, `TestComposeStatus` was failing without this change, shorter change is nice

internal/v1/handler_blueprints.go
andywaltlova: Now called once before iterating over users

internal/v1/handler_blueprints.go
andywaltlova: applied

internal/v1/handler_blueprints.go
andywaltlova: changed

internal/v1/handler_blueprints.go
andywaltlova: moved to the top of the file next oblueprintNameRegex

internal/v1/handler_blueprints.go
andywaltlova: exported and renamed

lzap (approved):
Awesome work, just add a note in the description about reading passwords it is not mentioned and it can be surprising behavior. Please squash into a single commit, I am good to go.

lzap:
An actual (relevant) test failure.

lzap (approved):

lzap:
Thanks for the help with CI! Great work @andywaltlova.

PR #1299, opened 2024-08-14
cmd/image-builder: add deployment channel as field to the logs

schuellerf:
add deployment channel as field to the logs similar to 
https://github.com/osbuild/osbuild-composer/pull/4285
and based on
https://github.com/osbuild/osbuild-composer/pull/4301

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -59,11 +59,8 @@ jobs:
           version: v1.56.2
           skip-cache: true
 
-      - name: install tern
-        run: go install github.com/jackc/tern@latest
-
       - name: Run unit tests
-        run: go test -v -race -covermode=atomic -coverprofile=coverage.txt -coverpkg=./... ./...
+        run: make unit-tests
 
       # disabled as 'pinging codecov' hangs
       # see https://github.com/codecov/feedback/issues/354
--- a/internal/v1/server_test.go
+++ b/internal/v1/server_test.go
@@ -191,7 +191,10 @@ func startServer(t *testing.T, tscc *testServerClientsConf, conf *ServerConfig) 
 	require.NoError(t, err)
 	// execute in parallel b/c .Run() will block execution
 	go func() {
-		_ = echoServer.Start("localhost:8086")
+		err = echoServer.Start("localhost:8086")
+		if err.Error() != "http: Server closed" {
+			require.NoError(t, err)
+		}
 	}()
 
 	// wait until server is ready
--- a/templates/image-builder.yml
+++ b/templates/image-builder.yml
@@ -130,6 +130,8 @@ objects:
                 key: proxy
                 name: recommendations-secrets
                 optional: true
+          - name: CHANNEL
+            value: ${CHANNEL}
         volumeMounts:
           - name: config-volume
             mountPath: /app/config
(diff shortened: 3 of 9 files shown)


schuellerf:
keeping as draft until I understand why the (unrelated) tests fail on my PC

templates/image-builder.yml
```diff
@@ -130,6 +130,8 @@ objects:
                 key: proxy
                 name: recommendations-secrets
                 optional: true
+          - name: CHANNEL
+            value: ${CHANNEL}
         volumeMounts:
           - name: config-volume
             mountPath: /app/config
```
lzap, line 137: This is missing the variable definition further down the file. But if the intention was not to set it explicitly it is fine.

lzap (dismissed):
Haven’t tested, LGTM.

templates/image-builder.yml
schuellerf: The definition should come implicitly from 
https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/products/osdv4/environments/stage.yml?ref_type=heads#L18
by stating that this template is "staging"
https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/image-builder/deploy.yml?ref_type=heads#L78
for example

thozza (approved):
LGTM

.github/workflows/tests.yml
```diff
@@ -59,11 +59,8 @@ jobs:
           version: v1.56.2
           skip-cache: true
 
-      - name: install tern
-        run: go install github.com/jackc/tern@latest
-
       - name: Run unit tests
-        run: go test -v -race -covermode=atomic -coverprofile=coverage.txt -coverpkg=./... ./...
+        run: make unit-tests
```
mvo5, line 63: The makefile has a comment `re-implementing .github/workflows/tests.yml as close as possible` which probably needs updating now :)

mvo5 (commented):
I was a bit late with my review it seems - oh well, sorry for that! Maybe my comments are still useful

internal/v1/server_test.go
```diff
@@ -191,7 +191,10 @@ func startServer(t *testing.T, tscc *testServerClientsConf, conf *ServerConfig)
 	require.NoError(t, err)
 	// execute in parallel b/c .Run() will block execution
 	go func() {
-		_ = echoServer.Start("localhost:8086")
+		err = echoServer.Start("localhost:8086")
+		if err.Error() != "http: Server closed" {
```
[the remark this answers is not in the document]
schuellerf: ServerClose always happens on regular shutdown...
I already did implement with `GetFreePort()` (locally) but stopped after changing ~1000 strings to take a dynamic host&port 😑
Will do this "major rework" at some other point in time

.github/workflows/tests.yml
schuellerf: Thanks - good catch 🙈

internal/v1/server_test.go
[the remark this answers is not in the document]
mvo5: I did a quick sed experiement and opened https://github.com/osbuild/image-builder/pull/1303 which may help, I did not really check much, it was mostly sed doing it's thing but it might be useful. No getFreePort there but that should be easy to hook up once the other pieces are there.

It also include the tiny `http.ErrServerClosed` tweak - hope this is useful.

PR #1303, opened 2024-08-16
tests: tweaks to the way routes are tested

mvo5:
tests: tweak tutil to take `addr` and `path` instead of url

This commit avoids the hardcoded URL and uses the addr that the
echo server is using. This makes it easier to run the tests on
a different port.

Mostly auto-generated via `sed` (unfortunately some manual tweaking
was needed because the coding style is not very consistent).

### DIFF
--- a/internal/v1/handler_blueprints_test.go
+++ b/internal/v1/handler_blueprints_test.go
@@ -59,7 +59,7 @@ func TestHandlers_CreateBlueprint(t *testing.T) {
 			},
 		},
 	}
-	statusCodePost, respPost := tutils.PostResponseBody(t, "http://localhost:8086/api/image-builder/v1/blueprints", body)
+	statusCodePost, respPost := tutils.PostResponseBody(t, db_srv.URL+"/api/image-builder/v1/blueprints", body)
 	require.Equal(t, http.StatusCreated, statusCodePost)
 
 	var result CreateBlueprintResponse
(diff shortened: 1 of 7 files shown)


internal/tutils/tutils.go
```diff
@@ -20,9 +21,9 @@ var AuthString1 = GetCompleteBase64Header("000001")
 
 var FedAuth = getBase64Header(fedoraHeader, "User")
 
-func GetResponseError(url string) (*http.Response, error) {
+func GetResponseError(addr, path string) (*http.Response, error) {
 	client := &http.Client{}
-	request, err := http.NewRequest("GET", url, nil)
+	request, err := http.NewRequest("GET", fmt.Sprintf("http://%s%s", addr, path), nil)
```
mvo5, line 26: fwiw, this is not super-great if we want to do https but (afaict) we don't test it right now and it seems easy enough to tweak once we do

ezr-ondrej (approved):
:orange_heart:

ezr-ondrej:
Nice improvement ;)

schuellerf (dismissed):

croissanne (dismissed):
yay

mvo5:
Meh, sorry, had to force push to fix a conflict, diff is trivial https://github.com/osbuild/image-builder/compare/c2d70b4805738be87ce106ba38cedf68b2bc3a9c..cb493e970114a44e52feb6807a3132195219cade but this will need one more approval now.

schuellerf (approved):
🥳

lzap (approved):
Seems green now, merging thanks.

PR #1304, opened 2024-08-16
server: use error type ErrServerClosed in test

mvo5:
Instead of string comparing use the http.ErrServerClosed when checking for echoServer.Start()


mgold1234 (approved):
/lgtm

PR #1305, opened 2024-08-18
test: enable codecov to be able to see the updated coverage

mgold1234:
this commit enable codecov to be able to see the updated coverage
this issue https://github.com/codecov/feedback/issues/354 was fixed

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,11 +62,11 @@ jobs:
       - name: Run unit tests
         run: make unit-tests
 
-      # disabled as 'pinging codecov' hangs
-      # see https://github.com/codecov/feedback/issues/354
-      # - name: Send coverage to codecov.io
-      #   run: bash <(curl -s https://codecov.io/bash)
-
+      - name: Upload results to Codecov
+        uses: codecov/codecov-action@v4
+        with:
+          files: ./coverage.txt
+          token: ${{ secrets.CODECOV_TOKEN }}
   speccheck:
     name: "📋 openapi spec check"
     runs-on: ubuntu-latest


ezr-ondrej:
This hangs if there is an exceed in limit of requests to GH, I think using https://github.com/codecov/codecov-action v4 solves it. Could we use that?

.github/workflows/tests.yml
```diff
@@ -65,10 +65,21 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@main
+    - uses: codecov/codecov-action@v4
+      with:
+        fail_ci_if_error: true # optional (default = false)
+        files: ./coverage1.xml,./coverage2.xml # optional
+        flags: unittests # optional
+        name: codecov-umbrella # optional
+        token: ${{ secrets.CODECOV_TOKEN }} # required
+        verbose: true # optional (default = false)
```
ezr-ondrej, line 81: I'd just take the `codecov/codecov-action@v4` step and replace the step we have now.

.github/workflows/tests.yml
```diff
@@ -65,10 +65,21 @@ jobs:
...
-      #   run: bash <(curl -s https://codecov.io/bash)
+      - name: Send coverage to codecov.io
+        run: bash <(curl -s https://codecov.io/bash)
+
+  codecov:
+    steps:
+    - uses: actions/checkout@main
+    - uses: codecov/codecov-action@v4
+      with:
+        fail_ci_if_error: true # optional (default = false)
```
ezr-ondrej, line 76: I'd keep this false.

.github/workflows/tests.yml
```diff
@@ -65,10 +65,21 @@ jobs:
...
+        files: ./coverage1.xml,./coverage2.xml # optional
```
ezr-ondrej, line 77: I'm not sure how are the files named, but there is autodiscovery, so we might want to try if the action finds the files automagically :)

.github/workflows/tests.yml
lzap: I am against all code coverage in general, in my mind this is false metric.

However, if other folks are okay I am not against having it just for informative purposes. Definitely should not be a failure when coverage decreases and this kind of stuff.

.github/workflows/tests.yml
```diff
@@ -62,10 +62,18 @@ jobs:
...
-      # see https://github.com/codecov/feedback/issues/354
-      # - name: Send coverage to codecov.io
-      #   run: bash <(curl -s https://codecov.io/bash)
+  codecov:
+    name: Run tests and collect coverage
+    runs-on: ubuntu-latest
+    steps:
+     - uses: actions/checkout@v2
+       with:
+         fetch-depth: 2
```
ezr-ondrej, line 71: I believe these all are useless, IMO we do not need separate job for codecov.
I think we can just add it to existing unit test run. Or have you had a reson to split it into a separate job?

.github/workflows/tests.yml
```diff
@@ -62,10 +62,18 @@ jobs:
...
+     - name: Upload results to Codecov
+       uses: codecov/codecov-action@v4
+       with:
+         files: ./coverage.txt
```
schuellerf, line 75: I think, now the `make unit-tests` is missing (which is generating the `coverage.txt`)
and probably best to run it only once (as @ezr-ondrej suggested) like it was before commenting it out

so just replacing the comment with
```
     - name: Upload results to Codecov
       uses: codecov/codecov-action@v4
       with:
         files: ./coverage.txt
         token: ${{ secrets.CODECOV_TOKEN }}
```
might work

schuellerf (changes requested):

.github/workflows/tests.yml
mgold1234: looks like it is working 
```info - 2024-08-19 19:28:51,723 -- Found 1 coverage files to report
info - 2024-08-19 19:28:51,723 -- > /home/runner/work/image-builder/image-builder/coverage.txt
info - 2024-08-19 19:28:53,258 -- Your upload is now processing. When finished, results will be available at: https://app.codecov.io/github/osbuild/image-builder/commit/4245f229d5c9be4e41c5770e2d10bb4059d7ea7a
info - 2024-08-19 19:28:53,466 -- Process Upload complete```

File: .github/workflows/tests.yml [hunk 6]
```diff
@@ -62,11 +62,11 @@ jobs:
...
 
-      # disabled as 'pinging codecov' hangs
-      # see https://github.com/codecov/feedback/issues/354
-      # - name: Send coverage to codecov.io
-      #   run: bash <(curl -s https://codecov.io/bash)
-
+      - name: Upload results to Codecov
+        uses: codecov/codecov-action@v4
+        with:
+          files: ./coverage.txt
```
ezr-ondrej, line 68: I'm still unsure that this is needed, but ok :shrug:

ezr-ondrej (approved):
Lets see what it does :+1:

ezr-ondrej:
@schuellerf are you good here? :) I don't want to dissmiss your change request, but looks resolved :)

.github/workflows/tests.yml
schuellerf: I think it's better to be explicit and also, as far as I saw it tries `.xml` and `.cov` and others. Not sure about `.txt` (which it was before in our project - no idea why though)

schuellerf (approved):
/LGTM
perfect, thanks

schuellerf:
Sidenote: It was pretty hard to find (as it's not documented) and I hope that's the correct implementation used, but considering the wide pattern it's really ok to explicitly specify the file to upload
https://github.com/codecov/uploader/blob/main/src/helpers/files.ts#L161

lzap (approved):

PR #1306, opened 2024-08-18
tests: add more coverage to composeStatus when response is InternalServerError

mgold1234:
This commit add unit test for `getComposeStatus` handling the StatusInternalServerError response after composeStatus found.
Increase the coverage of GetComposeStatus from *70.3%* to *83.8%*

### DIFF
--- a/internal/v1/handler_test.go
+++ b/internal/v1/handler_test.go
@@ -132,74 +132,96 @@ func TestGetComposeEntryNotFoundResponse(t *testing.T) {
 	require.Contains(t, body, "Compose entry not found")
 }
 
-func TestGetComposeStatusNotFoundResponse(t *testing.T) {
-	ctx := context.Background()
-	composeId := uuid.New()
-	var composerStatus composer.ComposeStatus
-	apiSrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if "Bearer" == r.Header.Get("Authorization") {
-			w.WriteHeader(http.StatusUnauthorized)
-			return
-		}
-		require.Equal(t, "Bearer accesstoken", r.Header.Get("Authorization"))
-		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(http.StatusNotFound)
-		err := json.NewEncoder(w).Encode(composerStatus)
-		require.NoError(t, err)
-	}))
-	defer apiSrv.Close()
-
-	dbase, err := dbc.NewDB()
-	require.NoError(t, err)
-	cr := ComposeRequest{
-		Distribution: "rhel-9",
-		Customizations: &Customizations{
-			// Since we are not calling handleCommonCompose but inserting directly to DB
-			// there is no point in using plaintext passwords
-			// If there is plaintext password in DB there is problem elsewhere (eg. CreateBlueprint)
-			Users: &[]User{
-				{
-					Name:     "user",
-					SshKey:   common.ToPtr("ssh-rsa AAAAB3NzaC2"),
-					Password: common.ToPtr("$6$password123"),
-				},
-			},
-		}}
-
-	crRaw, err := json.Marshal(cr)
-	require.NoError(t, err)
-	err = dbase.InsertCompose(ctx, composeId, "000000", "user000000@test.test", "000000", cr.ImageName, crRaw, (*string)(cr.ClientId), nil)
-	require.NoError(t, err)
-	srv, tokenSrv := startServer(t, &testServerClientsConf{ComposerURL: apiSrv.URL}, &ServerConfig{
-		DBase:            dbase,
-		DistributionsDir: "../../distributions",
-	})
-	defer func() {
-		err := srv.Shutdown(context.Background())
-		require.NoError(t, err)
-	}()
-	defer tokenSrv.Close()
-	payloads := []struct {
-		composerStatus composer.ComposeStatus
-		imageStatus    ImageStatus
+func TestGetComposeStatusErrorResponse(t *testing.T) {
+	testCases := []struct {
+		statusCode       int
+		checkImageStatus bool
+		expectedBody     string
 	}{
-		{
-			composerStatus: composer.ComposeStatus{
-				ImageStatus: composer.ImageStatus{},
-			},
+		{http.StatusNotFound,
+			true,
+			"",
+		},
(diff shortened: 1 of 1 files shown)


lzap (approved):

internal/v1/handler_test.go
```diff
@@ -203,6 +203,77 @@ func TestGetComposeStatusNotFoundResponse(t *testing.T) {
...
+			composerStatus: composer.ComposeStatus{
+				ImageStatus: composer.ImageStatus{},
+			},
+		},
+	}
+	for idx, payload := range payloads {
+		fmt.Printf("TT payload %d\n", idx)
+		composerStatus = payload.composerStatus
+
+		respStatusCode, body := tutils.GetResponseBody(t, fmt.Sprintf("http://localhost:8086/api/image-builder/v1/composes/%s",
```
schuellerf, line 267: maybe use the `srv.URL` here to, to be more flexible
```suggestion
		respStatusCode, body := tutils.GetResponseBody(t, fmt.Sprintf(srv.URL + "/api/image-builder/v1/composes/%s",
```

schuellerf (changes requested):
Raising the coverage is super cool 🥳📈
as this test is pretty similar to `TestGetComposeStatusNotFoundResponse` maybe you can use the pattern `testCases := []struct {…` and `for _, tc := range testCases {` as implemented for example in `TestWithoutOsbuildComposerBackend` or similar here too and unify
`TestGetComposeStatusNotFoundResponse` with `TestGetComposeStatusInternalServerErrorResponse` in one test function (having all common code only once

internal/v1/handler_test.go
schuellerf: the reason for `srv.URL` is that we might want to make the port dynamic to avoid crashes just due to port overlap… (or better: I'd like to make the port dynamic soonish :-D - like in [osbuild-composer tests](https://github.com/osbuild/osbuild-composer/blob/main/cmd/osbuild-worker-executor/main_test.go#L56) )

mgold1234:
@schuellerf in `TestGetComposeStatusNotFoundResponse` 
I have in the apiSrv `w.WriteHeader(http.StatusNotFound)`
and in ` TestGetComposeStatusInternalServerErrorResponse` I have in the apiSrv `w.WriteHeader(http.StatusInternalServerError)` I am not sure I can handle that the same way as TestGetComposeStatusNotFoundResponse

internal/v1/handler_test.go
mgold1234: perefct idea

internal/v1/handler_test.go
schuellerf, line 1: I think it could work similar to this to replace both functions (I did't check if it really works):

```
func TestGetComposeStatusResponse(t *testing.T) {
	testCases := []struct {
		statusCode       int
		checkImageStatus bool
		expectedBody     string
	}{
		{http.StatusNotFound,
			true,
			"",
		},
		{http.StatusInternalServerError,
			false,
			"Failed querying compose status",
		},
	}
	for _, tc := range testCases {
		ctx := context.Background()
		composeId := uuid.New()
		var composerStatus composer.ComposeStatus
		apiSrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			if "Bearer" == r.Header.Get("Authorization") {
				w.WriteHeader(http.StatusUnauthorized)
				return
			}
			require.Equal(t, "Bearer accesstoken", r.Header.Get("Authorization"))
			w.Header().Set("Content-Type", "application/json")
			w.WriteHeader(tc.statusCode)
			err := json.NewEncoder(w).Encode(composerStatus)
			require.NoError(t, err)
		}))
		defer apiSrv.Close()

		dbase, err := dbc.NewDB()
		require.NoError(t, err)
		cr := ComposeRequest{
			Distribution: "rhel-9",
			Customizations: &Customizations{
				// Since we are not calling handleCommonCompose but inserting directly to DB
				// there is no point in using plaintext passwords
				// If there is plaintext password in DB there is problem elsewhere (eg. CreateBlueprint)
				Users: &[]User{
					{
						Name:     "user",
						SshKey:   common.ToPtr("ssh-rsa AAAAB3NzaC2"),
						Password: common.ToPtr("$6$password123"),
					},
				},
			}}

		crRaw, err := json.Marshal(cr)
		require.NoError(t, err)
		err = dbase.InsertCompose(ctx, composeId, "000000", "user000000@test.test", "000000", cr.ImageName, crRaw, (*string)(cr.ClientId), nil)
		require.NoError(t, err)
		srv, tokenSrv := startServer(t, &testServerClientsConf{ComposerURL: apiSrv.URL}, &ServerConfig{
			DBase:            dbase,
			DistributionsDir: "../../distributions",
		})
		defer func() {
			err := srv.Shutdown(context.Background())
			require.NoError(t, err)
		}()
		defer tokenSrv.Close()
		payloads := []struct {
			composerStatus composer.ComposeStatus
			imageStatus    ImageStatus
		}{
			{
				composerStatus: composer.ComposeStatus{
					ImageStatus: composer.ImageStatus{},
				},
			},
		}
		for idx, payload := range payloads {
			fmt.Printf("TT payload %d\n", idx)
			composerStatus = payload.composerStatus

			respStatusCode, body := tutils.GetResponseBody(t, fmt.Sprintf(srv.URL+"/api/image-builder/v1/composes/%s", composeId), &tutils.AuthString0)
			require.Equal(t, tc.statusCode, respStatusCode)
			var result ComposeStatus
			err := json.Unmarshal([]byte(body), &result)
			require.NoError(t, err)
			if tc.checkImageStatus {
				require.Equal(t, payload.imageStatus, result.ImageStatus)
			}
			if tc.expectedBody != "" {
				require.Contains(t, body, tc.expectedBody)
			}
		}
	}
}

```

schuellerf (approved):
You can rewrite to make less code duplication or we'll take it as is for now, so I'll approve and you can choose

ezr-ondrej (approved):

ezr-ondrej:
The unit tests are failing now, apart of that, nice job! :tada:

schuellerf (changes requested):
so I guess `TestGetComposeStatusNotFoundResponse()` is not obsolete and can be removed?
Looks nice once the code duplication is removed!

schuellerf (approved):
Cool, thanks!

schuellerf (approved):

PR #1310, opened 2024-08-19
templates/image-builder: add CHANNEL as parameter so it get's filled

schuellerf:
CHANNEL has to be specified as parameter to get filled
otherwise it's literally `${CHANNEL}` :-|
followup of #1299


achilleas-k (approved):

schuellerf:
Sidenote: it was pretty hard to find as it's badly documented and I hope that's the correct implementation but considering this wide pattern it's really ok to explicitly set the filename
https://github.com/codecov/uploader/blob/main/src/helpers/files.ts#L161

PR #1315, opened 2024-08-26
distributions: regenerate OpenSCAP configs

schutzbot:
Update OpenSCAP configs


lzap:
What on Earth is this...

...but is green. :-D

lzap (approved):
Only distro jsons changed, tests are green.

PR #1318, opened 2024-08-27
tests: add coverage to `GetCloneStatus` unit test

mgold1234:
* this commit add unit test for `GetCloneStatus` handling 404 notFound response cloneEntry doesnt found
* increase the coverage from 48% to 54.1%

### DIFF
--- a/internal/v1/handler_test.go
+++ b/internal/v1/handler_test.go
@@ -673,6 +673,21 @@ func TestGetCloneStatus(t *testing.T) {
 	require.Equal(t, "us-east-2", awsUS.Region)
 }
 
+func TestGetCloneEntryNotFoundResponse(t *testing.T) {
+	id := uuid.New().String()
+	srv, tokenSrv := startServer(t, &testServerClientsConf{}, nil)
+	defer func() {
+		err := srv.Shutdown(context.Background())
+		require.NoError(t, err)
+	}()
+	defer tokenSrv.Close()
+
+	respStatusCode, body := tutils.GetResponseBody(t, srv.URL+fmt.Sprintf("/api/image-builder/v1/clones/%s",
+		id), &tutils.AuthString0)
+	require.Equal(t, http.StatusNotFound, respStatusCode)
+	require.Contains(t, body, "Clone not found")
+}
+
 func TestValidateSpec(t *testing.T) {
 	spec, err := GetSwagger()
 	require.NoError(t, err)


schuellerf (approved):
LGTM

internal/v1/handler_test.go
```diff
@@ -651,6 +651,33 @@ func TestGetCloneStatus(t *testing.T) {
 	require.Equal(t, "us-east-2", awsUS.Region)
 }
 
+func TestGetCloneEntryNotFoundResponse(t *testing.T) {
+	id := uuid.New().String()
+	apiSrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
mvo5, line 656: AFAICT we do not not an apiSrv for this test, it seems to be purely working on the DB from image-builder (in the clones endpoint) so I think the entire block can be removed:
```go
diff --git a/internal/v1/handler_test.go b/internal/v1/handler_test.go
index b22bc2ad..0f007e34 100644
--- a/internal/v1/handler_test.go
+++ b/internal/v1/handler_test.go
@@ -653,19 +653,8 @@ func TestGetCloneStatus(t *testing.T) {
 
 func TestGetCloneEntryNotFoundResponse(t *testing.T) {
        id := uuid.New().String()
-       apiSrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-               if "Bearer" == r.Header.Get("Authorization") {
-                       w.WriteHeader(http.StatusUnauthorized)
-                       return
-               }
-               require.Equal(t, "Bearer accesstoken", r.Header.Get("Authorization"))
-               w.Header().Set("Content-Type", "application/json")
-               w.WriteHeader(http.StatusNotFound)
-               fmt.Fprint(w, "404 during tests")
-       }))
-       defer apiSrv.Close()
 
-       srv, tokenSrv := startServer(t, &testServerClientsConf{ComposerURL: apiSrv.URL}, nil)
+       srv, tokenSrv := startServer(t, &testServerClientsConf{}, nil)
        defer func() {
                err := srv.Shutdown(context.Background())
                require.NoError(t, err)
```

mgold1234: thank you :)

mvo5 (approved):
Thank you, looks fine.

schuellerf (approved):

PR #1320, opened 2024-08-28
codecov.yml: add comments to the PR

schuellerf:
… to make codecov more visible

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,4 +6,8 @@ coverage:
         threshold: 5%
 codecov:
   require_ci_to_pass: no
-comment: no
+comment:
+    behavior: new
+    layout: "condensed_header, files, condensed_footer"
+    hide_project_coverage: TRUE
+


schuellerf:
@lucasgarfield @regexowl that comment is not what we want I guess, but it might work as you expected once this is merged :-D

regexowl:
Sounds good to me! Looking forward to see coverage reports again :rocket: 

I've dug through frontend a bit - this is how the codecov config looks like: https://github.com/osbuild/image-builder-frontend/blob/main/codecov.yml and this is how the comments looked like when the codecov in frontend worked: https://github.com/osbuild/image-builder-frontend/pull/987#issuecomment-1458115047

Not sure if it's helpful in any way, but hey - maybe for reference or so :sweat_smile:

ezr-ondrej:
I've never find it useful :stuck_out_tongue:

codecov.yml
```diff
@@ -6,4 +6,8 @@ coverage:
         threshold: 5%
 codecov:
   require_ci_to_pass: no
-comment: no
+comment:
+    behavior: new
+    layout: "condensed_header, files, condensed_footer"
+    hide_project_coverage: TRUE
```
mgold1234, line 12: why we should hide porject coverage here?

schuellerf: Because on each PR this is too much information and will distract people. If every PR increases or keeps the percentage of coverage - that's a win, I guess.

schuellerf:
> I've never find it useful 😛

Tests in general or coverage reports 😅🙃
I'm motivated to make them useful

mgold1234 (approved):
/lgtm

PR #1321, opened 2024-08-28
Makefile: pull new ubi-container always

schuellerf:
in my setup it took me a while to figure out that I had an old golang version as my cached container was old.
It should not hurt to always check if it's new


lzap (approved):
Should not hurt the CI.

lzap:
Conflicts... what were the odds!

schuellerf:
duplicated by #1323

PR #1323, opened 2024-08-29
Makefile/Dockerfile-ubi: implement logging git hash

schuellerf:
reporting the git hash in the logs should ease the search which version exactly is deployed

any thoughts on this (or how to really verify the deployed version or if the deployment is still ongoing)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,7 @@ check-api-spec:
 
 .PHONY: ubi-container
 ubi-container:
+	if [ -f .git ]; then echo "You seem to be in a git worktree - build will fail here"; exit 1; fi
 	podman build --pull=always -t osbuild/image-builder -f distribution/Dockerfile-ubi .
 
 .PHONY: generate-openscap-blueprints
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ check-api-spec:
 
 .PHONY: ubi-container
 ubi-container:
-	podman build -t osbuild/image-builder -f distribution/Dockerfile-ubi .
+	podman build --pull=always -t osbuild/image-builder -f distribution/Dockerfile-ubi .
 
 .PHONY: generate-openscap-blueprints
 generate-openscap-blueprints:
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"runtime/debug"
 
 	"github.com/osbuild/image-builder/internal/oauth2"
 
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -15,9 +15,12 @@ help:
 	@echo "    push-check:         Replicates the github workflow checks as close as possible"
 	@echo "                        (do this before pushing!)"
 
+VERSION="NotUsed"
+COMMIT_HASH=$(shell git rev-parse HEAD)
+
 .PHONY: image-builder
 image-builder:
-	go build -o image-builder ./cmd/image-builder/
+	go build -ldflags "-X main.Version=$(VERSION) -X main.GitHash=$(COMMIT_HASH)" -o image-builder ./cmd/image-builder/
```
mvo5, line 23: go includes the git revision automatically during the build (if build from git) so you can just read it via `debug.ReadBuildInfo()`. See https://github.com/osbuild/bootc-image-builder/blob/main/bib/cmd/bootc-image-builder/main.go#L550 for an example. With this no changes to go build/go install should be necessary.

Makefile
```diff
@@ -15,9 +15,12 @@ help:
 	@echo "    push-check:         Replicates the github workflow checks as close as possible"
 	@echo "                        (do this before pushing!)"
 
+VERSION="NotUsed"
```
mvo5, line 18: (nitpick) I personally would not include this yet if it is always set to "NotUsed" and only introduce it when it is used (YAGNI and all that :)

mvo5 (commented):
I like the idea of including the git hash in the logs, we should probably do this everywhere. A quick suggestion about `debug.ReadBuildInfo` inline that may simplify things.

cmd/image-builder/main.go
```diff
@@ -28,6 +29,21 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func gitRev() string {
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		logrus.Warn("cannot read build info")
```
mvo5, line 35: (nipitck) I personally would avoid logrus in such a lowlevel helper and do that in the caller as I assume this code will be copied around/extracted into shared helpers and might be used in places where we use less logrus (bib for example). But feel free to ignore, it's very much a personal opinion and I see the benefits.

mvo5 (approved):
Thank you! I like that we log the git version, I think this is a great help.

cmd/image-builder/main.go
```diff
@@ -26,6 +27,22 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func gitRev() string {
+	info, ok := debug.ReadBuildInfo()
```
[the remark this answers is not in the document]
schuellerf: For debugging the build-workflow I think this will be interesting once either of those happen (to know which one it is)

cmd/image-builder/main.go
```diff
@@ -26,6 +27,22 @@ import (
...
+		logrus.Warn("cannot read build info")
+		return "unknown"
+	}
+	for _, bs := range info.Settings {
+		logrus.Warnf("got %s:%s", bs.Key, bs.Value)
+		if bs.Key == "vcs.revision" {
+			return bs.Value
+		}
+	}
+	logrus.Warn("vcs.revision not found in debug.ReadBuildInfo()")
```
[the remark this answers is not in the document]
schuellerf: see above - just to be prepared on changes in `debug.ReadBuildInfo()` and find out which one failed

cmd/image-builder/main.go
schuellerf: good input, we could as well change the signature to return `string,error` I'll revise and reuse the function in composer, …

PR #1324, opened 2024-08-29
v1/internal: refactor `GetComposeStatus` improve error handling 

mgold1234:
this commit improve error handling and code structure
* Split GetComposeStatus into smaller, more focused functions
* Enhance error handling with more specific HTTP status codes *Add dedicated functions for handling different response scenarios
* Implement parseAndRedactComposeRequest for better separation of concerns
* Improve code readability and maintainability

### DIFF
--- a/internal/v1/handler.go
+++ b/internal/v1/handler.go
@@ -238,35 +238,11 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
 		return err
 	}
 
-	resp, err := h.server.cClient.ComposeStatus(composeId)
-	if err != nil {
-		return err
-	}
-	defer closeBody(ctx, resp.Body)
-
-	if resp.StatusCode == http.StatusNotFound {
-		body, err := io.ReadAll(resp.Body)
-		if err != nil {
-			return err
-		}
-		// Composes can get deleted in composer, usually when the image is expired
-		return echo.NewHTTPError(http.StatusNotFound, string(body))
-	} else if resp.StatusCode != http.StatusOK {
-		httpError := echo.NewHTTPError(http.StatusInternalServerError, "Failed querying compose status")
-		body, err := io.ReadAll(resp.Body)
-		if err != nil {
-			ctx.Logger().Errorf("Unable to parse composer's compose response: %v", err)
-		} else {
-			_ = httpError.SetInternal(fmt.Errorf("%s", body))
-		}
-		return httpError
-	}
-
 	var composeRequest ComposeRequest
-	err = json.Unmarshal(composeEntry.Request, &composeRequest)
-	if err != nil {
-		return err
+	if err := json.Unmarshal(composeEntry.Request, &composeRequest); err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, "Failed to unmarshal compose request").SetInternal(err)
 	}
+
 	if composeRequest.Customizations != nil && composeRequest.Customizations.Users != nil {
 		users := *composeRequest.Customizations.Users
 		for i := range users {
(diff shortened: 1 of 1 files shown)


schuellerf (changes requested):
While this makes the functions smaller and more readable
it duplicates the number of requests (calling `h.getComposeByIdAndOrgId(ctx, composeId)` twice now)?
If I'm right - this should be optimized to avoid duplicating the number of external requests

internal/v1/handler.go
```diff
@@ -298,6 +291,26 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
...
+	return echo.NewHTTPError(http.StatusNotFound, string(body))
+}
+
+func (h *Handlers) handleErrorResponse(ctx echo.Context, resp *http.Response) error {
+	httpError := echo.NewHTTPError(http.StatusInternalServerError, "Failed querying compose status")
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		ctx.Logger().Errorf("Unable to parse composer's compose response: %v", err)
+	} else {
+		_ = httpError.SetInternal(fmt.Errorf("%s", body))
```
schuellerf, line 309: I'm not the go expert - why would you assign something to `_` instead of just skipping `_ = `

internal/v1/handler.go
```diff
@@ -298,6 +291,26 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
...
+	}
+	// Composes can get deleted in composer, usually when the image is expired
+	return echo.NewHTTPError(http.StatusNotFound, string(body))
+}
+
+func (h *Handlers) handleErrorResponse(ctx echo.Context, resp *http.Response) error {
+	httpError := echo.NewHTTPError(http.StatusInternalServerError, "Failed querying compose status")
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		ctx.Logger().Errorf("Unable to parse composer's compose response: %v", err)
```
mvo5, line 307: I know this is a pre-existing string but it's a bit misleading, here it's not about `parse` but about reading.

internal/v1/handler.go
```diff
@@ -299,6 +292,26 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
 	return ctx.JSON(http.StatusOK, status)
 }
 
+func (h *Handlers) handleNotFoundResponse(resp *http.Response) error {
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, "Failed to read response body").SetInternal(err)
+	}
+	// Composes can get deleted in composer, usually when the image is expired
```
mvo5, line 300: This comment is specific about `composes` but the function is now very generic, it can handle any NotFoundResponses and amend them with the body of the reply. So it seems to me the comment should be moved into GetComposeStatus() maybe here https://github.com/osbuild/image-builder/pull/1324/files#diff-7ca1815ab9c73f0b2b01afdd43fd976e6a77d5fcec07e7f5076251c43ff0dd4dR262 ?

internal/v1/handler.go
```diff
@@ -299,6 +292,26 @@ func (h *Handlers) GetComposeStatus(ctx echo.Context, composeId uuid.UUID) error
...
+	return echo.NewHTTPError(http.StatusNotFound, string(body))
+}
+
+func (h *Handlers) handleErrorResponse(ctx echo.Context, resp *http.Response) error {
+	httpError := echo.NewHTTPError(http.StatusInternalServerError, "Failed querying compose status")
```
mvo5, line 305: The error message is about `compose status` but the function name is very generic so I would suggest to either make the function name more specific (maybe `handleErrorResponseForComposeStatus`) *or* (probably better) pass something like a "errorContext" or "errorDescription" into the helper, so something like:
```go
func (h *Handlers) handleErrorResponse(ctx echo.Context, errorDescription string, resp *http.Response) error {
	httpError := echo.NewHTTPError(http.StatusInternalServerError, errorDescription)
...
```
(and the same for the ctx.Logger() below) and of course the caller now needs to pass `"Failed querying compose status"` or if a different function uses this helper the caller they can use the appropriate message.

This way the helper can be reused in all places that need to handle error responses.

mvo5 (commented):
Thank you and sorry that it took me so long to look at this! This looks nice, some tiny naming/code organisation suggestions inline for your consideration.

internal/v1/handler.go
mgold1234: Thank you I think I should remove it

mvo5 (commented):
Thank you for this refactor, I have two small suggestions inline, but if you prefer we could also do them as a followup (but I included a suggested diff so maybe/hopefully it's not much trouble).

internal/v1/handler.go
```diff
@@ -455,7 +467,7 @@ func (h *Handlers) GetComposeMetadata(ctx echo.Context, composeId uuid.UUID) err
 		}
 		return echo.NewHTTPError(http.StatusNotFound, string(body))
 	} else if resp.StatusCode != http.StatusOK {
-		httpError := echo.NewHTTPError(http.StatusInternalServerError, "Failed querying compose status")
+		httpError := echo.NewHTTPError(http.StatusInternalServerError, fmt.Sprintf("Failed querying compose status"))
```
mvo5, line 470: AFAICT the Sprintf is not needed here. Maybe also switch to "Failed querying compose metadata" or similar (as the call that fails above is "h.server.cClient.ComposeMetadata(composeId)"?

mgold1234: I missed the removing of 'Sprintf' word, sorry about that

internal/v1/handler.go
mvo5: AFAICT this still an open topic, no? Maybe: `ctx.Logger().Errorf("Unable to read error response data: %v", err)` ?

mvo5 (commented):
This is nice, thank you! One tiny comment left but everything else looks very nice now

schuellerf (approved):
LGTM, thanks!

mvo5 (approved):
Thank you!

PR #1325, opened 2024-08-30
codecov.yml: change codecov to avoid new comments

schuellerf:
Generating a new comment seems fine, but I guess generates too many E-Mail which might get annoying.

So we'll try this behavior and the footer also seems not necessary


croissanne (approved):

mgold1234 (commented):
/lgtm

PR #1327, opened 2024-09-02
cmd/image-builder/main: restructure gitRev for reusability

schuellerf:
This should improve the usability of `gitRev()` to be more generic.
Although this returns a string in case of errors but that's on purpose (see code-comment)

followup of #1323 

---

@mvo5 please check
also please check if I should create a new error class or just wrap and unwrap the string like I did as it's not really important here.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"runtime/debug"
 
@@ -29,19 +30,18 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func gitRev() string {
+// gitRev returns the gitHash of the current running binary
+func gitRev() (string, error) {
 	info, ok := debug.ReadBuildInfo()
 	if !ok {
-		logrus.Warn("cannot read build info")
-		return "unknown"
+		return "", errors.New("cannot read build info")
 	}
 	for _, bs := range info.Settings {
 		if bs.Key == "vcs.revision" {
-			return bs.Value
+			return bs.Value, nil
 		}
 	}
-	logrus.Warn("vcs.revision not found in debug.ReadBuildInfo()")
-	return "unknown"
+	return "", errors.New("vcs.revision not found in debug.ReadBuildInfo()")
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


cmd/image-builder/main.go
```diff
@@ -29,19 +30,21 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func gitRev() string {
+// gitRev returns the gitHash. In case of errors
+// the string "unknown" **and** an error with details
```
mvo5, line 34: (nitpick) It is a quite unusual in go to return a "valid" value and an error at the same time. The usual pattern is to return the "empty" value when an error happens. Maybe just:
```go
diff --git a/cmd/image-builder/main.go b/cmd/image-builder/main.go
index bb66635f..7da4086b 100644
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -30,21 +30,18 @@ import (
        "github.com/sirupsen/logrus"
 )
 
-// gitRev returns the gitHash. In case of errors
-// the string "unknown" **and** an error with details
-// about the failure is returned.
+// gitRev returns the gitHash of the current running binary
 func gitRev() (string, error) {
-       const unknownHash = "unknown"
        info, ok := debug.ReadBuildInfo()
        if !ok {
-               return unknownHash, errors.New("cannot read build info")
+               return "", errors.New("cannot read build info")
        }
        for _, bs := range info.Settings {
                if bs.Key == "vcs.revision" {
                        return bs.Value, nil
                }
        }
-       return unknownHash, errors.New("vcs.revision not found in debug.ReadBuildInfo()")
+       return "", errors.New("vcs.revision not found in debug.ReadBuildInfo()")
 }
 
 func main() {
@@ -81,9 +78,8 @@ func main() {
 
        gitRev, err := gitRev()
        if err != nil {
-               // just warn about the error details
-               // gitRev will contain a valid string nevertheless
                logrus.Warn(err.Error())
+               gitRev = "unknown"
        }
 
        logrus.Infof("Starting image-builder from Git Hash: %s", gitRev)
```
which has the same effect but feel a little bit more idiomatic ? Then the comments can also be simplified as the code will do what most developers would expect, eliminating the need for the comments :-) Sorry, I know this is quite nitpicky

schuellerf: I also thought about this, but then the text "unknown" might be (will be) inconsistent between implementations!? 
(if we would reuse the function in more projects/binaries)
which (also being nitpicky) would not be super nice :-/ 🤷‍♂️

mvo5: Up to you, it's a tradeoff but personally still prefer this over non-idiomatic code

lzap (approved):

PR #1328, opened 2024-09-02
internal/v1/server_test.go: cancel tests on failed test server

schuellerf:
The current implementation just creates a "failed test" entry (e.g. if the port is in use) and continues testing

I had the problem again of some other debug service running and the error (of port in use) was lost in console output

### DIFF
--- a/internal/v1/server_test.go
+++ b/internal/v1/server_test.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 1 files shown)


internal/v1/server_test.go
```diff
@@ -207,7 +208,9 @@ func startServer(t *testing.T, tscc *testServerClientsConf, conf *ServerConfig)
 	URL := "http://" + addr
 	go func() {
 		err = echoServer.Start(addr)
-		require.Equal(t, err, http.ErrServerClosed)
+		if !errors.Is(err, http.ErrServerClosed) {
+			panic(fmt.Errorf("starting test server failed %w", err))
```
croissanne, line 212: can none of the testing functions be of help here? Like https://pkg.go.dev/testing#B.Fatalf or something? might be nicer than a panic.

schuellerf: In my tests with `t.Fatalf()` the code still tries to connect to the wrong server (the one who has the port bound)
So does not stop the tests fast enough

schuellerf:
Just start an `nc -l 127.0.0.1 8086` this causes strange behavior - I only got it working with panic (in a human readable manner)

croissanne (approved):

PR #1330, opened 2024-09-03
clients/compliance: use minor version instead of tailoring id

croissanne:
There's only a single tailoring item per rhel minor version per policy, and
    you can index the tailorings by id or by rhel minor version.

---

The tailoring id was hardcoded and the test didn't catch it because it was the same id as the hardcoded one in the test :facepalm:


kingsleyzissou (approved):

kingsleyzissou (approved):
Oh nice, seems to have cleaned up the function nicely too

PR #1343, opened 2024-09-16
v1/api: Export empty subscription on existing subscription

avitova:
In case the Subscription field is nil, we leave it nil. Otherwise, we overwrite it with an empty subscription.
This should help us on import to detect if there was a subscription, and therefore prefill the fields according to that.
I hope this makes sense.

### DIFF
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -301,7 +301,9 @@ func (h *Handlers) ExportBlueprint(ctx echo.Context, id openapi_types.UUID) erro
 		return err
 	}
 
-	blueprint.Customizations.Subscription = nil
+	if blueprint.Customizations.Subscription != nil {
+		blueprint.Customizations.Subscription = &Subscription{}
+	}
 	blueprintExportResponse := BlueprintExportResponse{
 		Name:           blueprintEntry.Name,
 		Description:    blueprintEntry.Description,
--- a/internal/v1/handler_blueprints_test.go
+++ b/internal/v1/handler_blueprints_test.go
@@ -816,6 +816,7 @@ func TestHandlers_GetBlueprint(t *testing.T) {
 	require.Equal(t, blueprint.ImageRequests, result.ImageRequests)
 	require.Equal(t, blueprint.Distribution, result.Distribution)
 	require.Equal(t, blueprint.Customizations.Packages, result.Customizations.Packages)
+	require.Nil(t, blueprint.Customizations.Subscription)
 	// Check that the password returned is redacted
 	for _, u := range *result.Customizations.Users {
 		require.Nil(t, u.Password)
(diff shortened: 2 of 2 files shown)


amirfefer (commented):
Thanks @avitova,

Maybe I'm missing something here, but the exported blueprint's subscription is initialized like this:

```json
{
  "activation-key": "",
  "base-url": "",
  "insights": false,
  "organization": 0,
  "server-url": ""
}
```
When importing it into the wizard, this triggers the `subscribed` [flag](https://github.com/osbuild/image-builder-frontend/pull/2434/files#diff-c859a462b63358defddb3cf67d9f3608f5e489b13cc5e56f0b7936dd2641a9dbR305), but all the values are essentially "zero" or empty., This could lead to an invalid state in the wizard since these values don't represent a valid subscription configuration.

amirfefer:
If the goal here is to have an empty {} subscription in the exported blueprint
```json
{
"customizations":
  "subscription": {}
}
```
then, the `Subscription` struct should have `omitempty` for each field.

lzap (approved):
I also think that omitting should be preferred over empty dictionary in this case.

avitova:
So the initial plan was to handle the nil values of the struct on frontend, and set default values instead. But Amir is right, that this smells.
I looked into the possibilities. Currently, we have almost all the fields in Subscription set as required, therefore there is no omitempty in the json structure definition for Subscription. In order to be able to marshal the Subscription structure as Subscription{}, we would have to:
a) make all of the values inside Subscription as "not required", which we do not want for sure
b) create a separate SubscriptionExport structure that would be part of CustomizationExport
```
BlueprintExportResponse {
    CustomizationExport {
        SubscriptionExport {
            everythingTheSameAsSubscriptionButPointers,
        },
        everythingElseSameAsCustomizations,
    }
}
```
^ sounds like tons of duplication, and a huge overkill for this purpose; smells as well

c) let's do the following in [internal/v1/handler_blueprints.go](https://github.com/osbuild/image-builder/pull/1343/files#diff-7e4feba1f6c48c70d1b9dd9561967bfeb38cff5a11fca3714ffb8d83eae1529aL310)
```
		Metadata: BlueprintMetadata{
			ExportedAt: time.Now().UTC().String(),
			ParentId:   &id,
			Subscribed: blueprint.Customizations.Subscription != nil,
		},
```

What do you think? / Any suggestions? Did I miss something?

lzap:
In RHEL context, "don’t register" is only useful for quick testing, perhaps for us as engineers and designers. But in real production use, every RHEL must be registered, thus it makes sense to have "Register to:" with default customer activation key as the default option during import.

Activation keys cannot be exported as they are "secret" information, thus, any sort of auto-population is not possible there. And I see very little value in pre-selecting "Do not register" when that was the option used in the source blueprint. However, if you really need it, I suggest to put this information into blueprint metadata, a special structure that is stored with every blueprint and leave Subscription off the blueprint structure for every single export.

avitova:
Okay, I am closing this pull request, as it seems like we came to the conclusion that this has very low value for customers. :) Thank you for your inputs. I believe this is a smart decision.

PR #1345, opened 2024-09-17
floorplan: Export files customization

ochosi:
This should allow us to track the usage of the first boot script feature.

I looked into the json request and what we're interested in the most is probably the path. While we can almost rule out that a lot of customers are using the files customization (because overall API usage is low) and therefore our data wouldn't be skewed too much, it'd be great to limit the results we account for to `/etc/systemd/system/custom-first-boot.service`

Or would it be more efficient/reliable to use the services customization to check for `"custom-first-boot"` in `enabled`?

### DIFF
--- a/templates/image-builder.yml
+++ b/templates/image-builder.yml
@@ -196,6 +196,8 @@ objects:
           c.request->'customizations'->'filesystem' as filesystem,
           c.request->'customizations'->'payload_repositories' as payload_repositories,
           c.request->'customizations'->'openscap' as openscap,
+          c.request->'customizations'->'files' as files,
+          c.request->'customizations'->'services' as services,
           jsonb_array_length(c.request->'customizations'->'users') as num_users
         from
           composes c


croissanne (dismissed):

templates/image-builder.yml
```diff
@@ -196,6 +196,7 @@ objects:
           c.request->'customizations'->'filesystem' as filesystem,
           c.request->'customizations'->'payload_repositories' as payload_repositories,
           c.request->'customizations'->'openscap' as openscap,
+          c.request->'customizations'->'files' as files,
```
croissanne, line 199: You might also want `c.request->'customizations'->'services' as services`, the first boot service will also show up in there.

ochosi: Yeah, that's what I was wondering about - is it more efficient to just rely on the `services` bit.
I guess we can try to dump both for now. I'll update the PR.

croissanne (approved):

PR #1349, opened 2024-09-23
Upgrade to Go 1.21.13

lzap:
Dependa bots latest update `gosec` is complaining about old Go version, let’s bump it:

https://github.com/osbuild/image-builder/pull/1348

Also bumping some dependencies too.


ezr-ondrej:
gosec has some new rules apparently... just to make file perms explicit it seems.

lzap:
Yeah I was already looking into it yesterday but got disturbed. Let’s see...

croissanne:
Is this version already in ubi-9?

cmd/oscap/main.go
```diff
@@ -233,7 +233,7 @@ func generateJson(dir, datastreamDistro, profileDescription, profile string) {
 	}
 	// hack to add an empty line at the end of the file for nicer diffs
 	bArray = append(bArray, '\n')
-	err = os.WriteFile(path.Join(dir, "customizations.json"), bArray, os.ModePerm)
+	err = os.WriteFile(path.Join(dir, "customizations.json"), bArray, 0640)
```
ezr-ondrej, line 236: Lets just disable the rule here? sounds bit too strict and seems it might be tricky to verify 0600 would work as well here.

lzap: @croissanne can you comment? I am not sure what this does, shall I set 666 here?

lzap:
So it now passes with 0600 for customization, should be fine according to @croissanne 

Tho there is a snag on CI:

```
go: go.mod requires go >= 1.21.13 (running go 1.21.11; GOTOOLCHAIN=local)
```

lzap:
Ok I mixed two problems together accidentally. What I am trying to solve is this problem:

```
github.com/oapi-codegen/oapi-codegen/v2@v2.4.0 requires go >= 1.21.13 (running go 1.21.9; ../../../go.mod sets toolchain go1.21.9)
```

This is unrelated to that, I am closing this for now, `gosec` is a separate issue that will be solved once Go toolchain is updated later on.

PR #1350, opened 2024-09-23
v1/handlers: Return 404 correctly when Blueprint not found (HMS-4243)

ezr-ondrej:
Return correct status on non-existent blueprint for blueprint/<id>/compose endpoint.


mvo5 (commented):
I wonder how hard it would be to have a test that checks for the 404 behavior?

mvo5 (approved):
Thank you! This looks fine, I looked a bit what it would take to add a unit test for the update case too and maybe totally overthought this, I put my ideas into the PR but it seems fine as is. I'm not super familiar with image-builders code base though so take my +1 more like a 0.5 :)

lzap (approved):

mvo5 (approved):

ezr-ondrej:
Two acks, CI green, small PR, I'll self merge this :)

PR #1354, opened 2024-09-25
handlers: extract `makeTestServer()` test helper

mvo5:
While reviewing PR#1350 I noticed that there is some repetition that could be extracted into a common helper. This commit does this and adds a new `makeTestServer()` that creates the DB, starts the db and token servers and passes a cleanup/shutdown helper back.

### DIFF
--- a/internal/v1/handler_blueprints_test.go
+++ b/internal/v1/handler_blueprints_test.go
@@ -20,13 +20,7 @@ import (
 	"github.com/osbuild/image-builder/internal/tutils"
 )
 
-func TestHandlers_CreateBlueprint(t *testing.T) {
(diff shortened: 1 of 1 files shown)


lzap (dismissed):

mvo5:
@croissanne added you mostly for awareness - I think this is okay (and the PR got a +1 already) but I I really like you being in the loop :)

croissanne (changes requested):
LGTM but seems like the linter found a tiny thing?

mvo5:
> LGTM but seems like the linter found a tiny thing?

Indeed, there was drift because the PR was old and new code with db_srv got added nice, should be fixed now :)

PR #1355, opened 2024-09-29
build: bump oapi-codegen to 2.4.1

lzap:
Since the project relaxed the hardcoded Go requirement, we can now update. I propose to keep the pin and thus to keep updating this generator manually. Or alternatively, we can just remove the hack completely.

Quite honestly, I am not huge fan of `go generate` and I would suggest to get rid of it completely calling `oapi-codegen` directly. This is what works for us quite nicely in provisioning app (this is from makefile but you get the idea):

* [installation target](https://github.com/RHEnVision/provisioning-backend/blob/main/mk/deps.mk#L23)
* [generation target](https://github.com/RHEnVision/provisioning-backend/blob/main/mk/clients.mk#L18)


mgold1234 (dismissed):
/lgtm

ezr-ondrej:
I'd love to get rid of the hack.
To minimize the dependencies, my vote goes to not directly depend on it. If it's not used in production code, it should not be packaged IMO :)

lzap:
> I'd love to get rid of the hack.

Before we do that, I will say that the hack makes generated content more deterministic. Instead of people regenerating with effectively "random" version, we can upgrade this and create commits which will only show upgrade changes. These generator upgrades did introduce some bugs in provisioning in the past.

lzap:
How about this?

tools/prepare-source.sh
```diff
@@ -2,20 +2,20 @@
...
 
-# this is the official way to get a different version of golang
-# see https://go.dev/doc/manage-install
+# Install Go SDK
 go install golang.org/dl/go$GO_VERSION@latest
 $GO_BINARY download
 
 # Ensure dev tools are installed
-which goimports || $GO_BINARY install golang.org/x/tools/cmd/goimports@latest
+which goimports >/dev/null || GOBIN=$TOOLS_PATH $GO_BINARY install golang.org/x/tools/cmd/goimports@latest
```
croissanne, line 15: Will which still work here? $TOOLS_PATH wouldn't be part of the user's path.

croissanne (commented):
Overall I'm not sure what the real gain is here in changing up the tooling, I don't feel strongly either way however. But each change we'll have will need to update the individual go generate directives.

Anyway LGTM, but one comment on the use of `which`.

lzap:
> But each change we'll have will need to update the individual go generate directives.

I was testing this and as far as I can tell, the tool picks the tool which is installed. I hope this will work, but it is hard to test now that there is only one release that we can use due to Go stack incompatiblity. The next update will tell. But I manually removed the version from all comment headers and the shell script does not add it back now! Looks good.

I need to figure out how to silence shellcheck on this line tho, it wants me to enquote it:

```
TOOLS_PATH="$(realpath $(dirname $0)/../bin)
```

tools/prepare-source.sh
lzap: Oh good catch! Rebasing.

lzap:
Requesting re-review @mgold1234 and @croissanne thank you.

lzap:
@croissanne can we get this one sorted before holidays? I know I squashed some things into single PR, if you do not like the `./bin` pattern this is fine I can remove this from the patch. but the kin-openapi dependency is pinned and we want to get it unpinned so this is not forgotten.

tools/prepare-source.sh
```diff
@@ -2,20 +2,20 @@
 set -eux
 
 GO_VERSION=1.22.0
+OAPI_VERSION=2.4.1
+
 GO_BINARY=$(go env GOPATH)/bin/go$GO_VERSION
-OAPI_VERSION=2.3.0
+TOOLS_PATH="$(realpath "$(dirname "$0")/../bin")"
```
croissanne, line 8: hmm, should we move this under `tools/bin`, that way it's clear these are just dependencies for tools? `/bin` could be or become the folder that we actually build into.

croissanne (commented):
I'm ok with this but it needs a small update for shellcheck

```
./tools/prepare-source.sh:15:9: note: Double quote to prevent globbing and word splitting. [SC2086]
```
Also maybe a small explanation in the commit message about the changes in tools?

lzap:
Done:

```
build: bump oapi-codegen to 2.4.1 and fix tools bin 

This change unpins oapi code generator via the sed hack and changes the
installation directory to tools/bin so we all use the same version of 
the CLI when calling the script. It also installs a specific version 
which is independent to `go.mod` so we will have to bump it from time to
time. This should bring better consisntency in regenerating sources.
```

croissanne (approved):
thanks!

PR #1361, opened 2024-10-07
OpenSCAP customizations endpoint (HMS-4803)

kingsleyzissou:
This is a proof of concept for getting OpenSCAP customizations in the v1 handler rather than having generated hard-code content.


kingsleyzissou:
Result:

```
curl -H "x-rh-identity: $IDHEADER" -H 'Content-Type: application/json' 172.31.0.40:8086/api/image-builder/v1/oscap/centos-9/xccdf_org.ssgproject.content_profile_cis/customizations
{
  "filesystem": [
    {
      "min_size": 1073741824,
      "mountpoint": "/home"
    },
    {
      "min_size": 1073741824,
      "mountpoint": "/tmp"
    },
    {
      "min_size": 3221225472,
      "mountpoint": "/var"
    },
    {
      "min_size": 1073741824,
      "mountpoint": "/var/log"
    },
    {
      "min_size": 10737418240,
      "mountpoint": "/var/log/audit"
    },
    {
      "min_size": 1073741824,
      "mountpoint": "/var/tmp"
    }
  ],
  "kernel": {
    "append": "audit_backlog_limit=8192 audit=1"
  },
  "openscap": {
    "profile_description": "This profile defines a baseline that aligns to the \"Level 2 - Server\" configuration from the Center for Internet Security® Red Hat Enterprise Linux 9 Benchmark™, v2.0.0, released 2024-06-20.  This profile includes Center for Internet Security® Red Hat Enterprise Linux 9 CIS Benchmarks™ content.",
    "profile_id": "xccdf_org.ssgproject.content_profile_cis",
    "profile_name": "CIS Red Hat Enterprise Linux 9 Benchmark for Level 2 - Server"
  },
  "packages": [
    "aide",
    "sudo",
    "libpwquality",
    "systemd-journal-remote",
    "firewalld",
    "nftables",
    "libselinux",
    "audit-libs",
    "audit"
  ],
  "services": {
    "enabled": [
      "auditd",
      "crond",
      "firewalld",
      "systemd-journald"
    ],
    "masked": [
      "cups",
      "nfs-server",
      "rpcbind",
      "avahi-daemon",
      "autofs",
      "bluetooth",
      "nftables"
    ]
  }
}
```

internal/v1/handler.go
```diff
@@ -843,9 +843,22 @@ func (h *Handlers) GetOscapProfiles(ctx echo.Context, distribution Distributions
...
 	if err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, err)
 	}
+
+	if distroFile == nil {
+		return echo.NewHTTPError(http.StatusBadRequest, fmt.Errorf("Unable to process distro file"))
+	}
+
+	// TODO: get tailoring customizations with the compliance client
+	customizations, err := processRequest(string(profile), distroFile.OscapDatastream, nil)
```
kingsleyzissou, line 858: Unfortunately this means a tonne of I/O inside the handler.

I have done a POC for an aws lambda as an alternative, if we want to think about that

internal/oscap/oscap.go
```diff
@@ -78,88 +76,87 @@ func GetProfileDescription(profile string, datastream string) string {
...
+		// we don't need to process this any further,
+		// since the xml file will just end up blank
+		// and would cause issues later down the line
+		return nil, nil
 	}
 
-	var packages []string
-	for _, bpPackage := range bp.Packages {
-		packages = append(packages, bpPackage.Name)
+	jsonFile, err := os.CreateTemp("", "tailoring.json")
```
kingsleyzissou, line 87: `autotailor` doesn't seem to be in the `RHEL-9 ubi` so maybe we could just pass the `xml` tailoring in directly for this. Otherwise we could maybe pull in the binary from `fedora` in the `Dockerfile`, feels a bit gross though

internal/v1/handler.go
kingsleyzissou: This is not as fleshed out, but here is the poc for the aws lambda:
https://github.com/osbuild/image-builder/compare/main...kingsleyzissou:image-builder:oscap-lambda?expand=1

internal/oscap/oscap.go
lzap: CreateTemp does open file for rw and returns a handle, the handle is not closed and then file is deleted which only means it is not available from the tree but since it is opened it looks like a file handle leak to me.

internal/v1/oscap.go
```diff
@@ -85,35 +59,117 @@ func OscapProfiles(distribution Distributions) (DistributionProfileResponse, err
...
+
+func processRequest(profile string, datastream string, tailoring *string) (*Customizations, error) {
+	// the generated blueprint doesn't contain the profile
+	// description, so we have to run the oscap tool to get
+	// this information
+	description := oscap.GetProfileDescription(profile, datastream)
+
+	var file *os.File
+	if tailoring != nil {
+		file, err := oscap.GenXMLTailoringFile(tailoring, datastream)
```
lzap, line 156: Ditto, the file is never closed thus kernel will keep the file handle until restart.

lzap:
Dropped few random comments, I am not too familiar with the OpenSCAP code in IB but overall looks good.

The only concern is if Fedora can get some "too new" and incompatible updates that would break RHEL systems. But it is workardoundable I guess.

internal/v1/oscap.go
kingsleyzissou: Whoops, thanks for catching this

kingsleyzissou:
Thanks for the review!

> The only concern is if Fedora can get some "too new" and incompatible updates that would break RHEL systems. But it is workardoundable I guess.

Yeah this is totally valid and something to keep in mind. I think we might end up waiting on this a bit before deciding if this is really the solution we want to go with.

PR #1362, opened 2024-10-07
v1: empty recommendation is warning now

lzap:
This error is cluttering glitchtip there are over 2k events and counting. This does not need to be an error:

obrazek


croissanne (approved):
thanks!

PR #1373, opened 2024-10-16
pr_check: don't remove resources for notifications

jrusz:
They probably set their limits wrong for ephemeral so don't strip them otherwise the deployment fails.


lzap (approved):
All green, merge at will.

PR #1378, opened 2024-10-18
templates/image-builder: point default IMAGE param to Konflux

croissanne:
The default container registry changed when switching to Konflux for container builds.


ezr-ondrej (approved):
Nice! Thanks

PR #1381, opened 2024-10-22
v1/api: Add exported custom repositories

avitova:
Currently, the custom_repositories field only stores ids. When exporting a blueprint and sharing with someone from different org, we can not reuse ids, because from different org, users would not be able to see it. For that reason, we need to export more details about the repositories on export.

[https://issues.redhat.com/browse/HMS-4811](https://issues.redhat.com/browse/HMS-4811)

Edit: There is a very long discussion, the summary is in this comment (https://github.com/osbuild/image-builder/pull/1381#issuecomment-2451696913)

### DIFF
--- a/internal/common/testdata/exported_blueprint.json
+++ b/internal/common/testdata/exported_blueprint.json
@@ -1,0 +1,38 @@
+{
+  "content_sources": [
+    {
+      "gpg_key": "some-gpg-key",
+      "name": "payload",
+      "url": "http://snappy-url/snappy/baseos"
+    }
+  ],
+  "customizations": {
+    "custom_repositories": [
+      {
+        "baseurl": [
+          "http://snappy-url/snappy/baseos"
+        ],
+        "gpgkey": [
+          "some-gpg-key"
+        ],
+        "id": "2531793b-c607-4e1c-80b2-fbbaf9d12790",
+        "name": "payload"
+      }
+    ],
+    "packages": [
+      "nginx"
+    ],
+    "users": [
+      {
+        "name": "user"
+      }
+    ]
+  },
+  "description": "desc",
+  "distribution": "centos-9",
+  "metadata": {
+    "exported_at": "2013-06-13 00:00:00 +0000 UTC",
+    "parent_id": "f43a4ec2-5447-4a25-8a62-e258ff11a2d9"
+  },
+  "name": "blueprint"
+}
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -1069,6 +1069,14 @@ components:
           $ref: '#/components/schemas/Customizations'
         metadata:
           $ref: '#/components/schemas/BlueprintMetadata'
+        content_sources:
+          type: array
+          items:
+            type: object
+          additionalProperties: true
+          description: |
+            List of custom repositories including all the repository details needed in order
+            to recreate the repositories.
     BlueprintMetadata:
       required:
         - parent_id
@@ -1568,6 +1576,7 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/CustomRepository'
+          description: List of custom repositories.
         openscap:
           $ref: '#/components/schemas/OpenSCAP'
         filesystem:
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -331,32 +331,28 @@ func (h *Handlers) ExportBlueprint(ctx echo.Context, id openapi_types.UUID) erro
 	}
 	defer closeBody(ctx, exportedRepositoriesResp.Body)
 
-	if exportedRepositoriesResp.StatusCode != http.StatusOK {
-		if exportedRepositoriesResp.StatusCode != http.StatusUnauthorized {
-			body, err := io.ReadAll(exportedRepositoriesResp.Body)
-			if err != nil {
-				return err
-			}
-			ctx.Logger().Warnf("Unable to export custom repositories: %s", body)
-		}
+	if exportedRepositoriesResp.StatusCode == http.StatusUnauthorized {
 		return fmt.Errorf("Unable to fetch custom repositories, got %v response", exportedRepositoriesResp.StatusCode)
 	}
 
-	if exportedRepositoriesResp.Body != nil {
-		bodyBytes, err := io.ReadAll(exportedRepositoriesResp.Body)
-		if err != nil {
-			return fmt.Errorf("Unable to export custom repositories: %w", err)
-		}
+	if exportedRepositoriesResp.Body == nil {
+		ctx.Logger().Warnf("Unable to export custom repositories, empty body")
+		return ctx.JSON(http.StatusOK, blueprintExportResponse)
+	}
 
-		if len(bodyBytes) != 0 {
-			// Saving the custom repositories in the object format
-			var result []map[string]interface{}
-			err = json.Unmarshal(bodyBytes, &result)
-			if err != nil {
-				return fmt.Errorf("Unable to export custom repositories: %w, %s", err, string(bodyBytes))
-			}
-			blueprintExportResponse.ContentSources = &result
+	bodyBytes, err := io.ReadAll(exportedRepositoriesResp.Body)
+	if err != nil {
+		return err
+	}
+
+	if len(bodyBytes) != 0 {
+		// Saving the custom repositories in the object format
+		var result []map[string]interface{}
+		err = json.Unmarshal(bodyBytes, &result)
+		if err != nil {
+			return fmt.Errorf("Unable to export custom repositories: %w, %s", err, string(bodyBytes))
 		}
+		blueprintExportResponse.ContentSources = &result
 	}
 
 	return ctx.JSON(http.StatusOK, blueprintExportResponse)
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -4,11 +4,14 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"regexp"
 	"strconv"
 	"time"
+
+	"github.com/osbuild/image-builder/internal/clients/content_sources"
 
 	"github.com/jackc/pgerrcode"
 	"github.com/jackc/pgx/v5/pgconn"
@@ -312,6 +315,50 @@ func (h *Handlers) ExportBlueprint(ctx echo.Context, id openapi_types.UUID) erro
 			ParentId:   &id,
 		},
 	}
+
+	repoUUIDs := []string{}
+	if blueprint.Customizations.CustomRepositories != nil {
+		for _, repo := range *blueprint.Customizations.CustomRepositories {
+			repoUUIDs = append(repoUUIDs, repo.Id)
+		}
+	}
+
+	exportedRepositoriesResp, err := h.server.csClient.BulkExportRepositories(ctx.Request().Context(), content_sources.ApiRepositoryExportRequest{
+		RepositoryUuids: common.ToPtr(repoUUIDs),
+	})
+	if err != nil {
+		return err
+	}
+	defer closeBody(ctx, exportedRepositoriesResp.Body)
+
+	if exportedRepositoriesResp.StatusCode != http.StatusOK {
+		if exportedRepositoriesResp.StatusCode != http.StatusUnauthorized {
+			body, err := io.ReadAll(exportedRepositoriesResp.Body)
+			if err != nil {
+				return err
+			}
+			ctx.Logger().Warnf("Unable to export custom repositories: %s", body)
+		}
+		return fmt.Errorf("Unable to fetch custom repositories, got %v response", exportedRepositoriesResp.StatusCode)
+	}
+
+	if exportedRepositoriesResp.Body != nil {
+		bodyBytes, err := io.ReadAll(exportedRepositoriesResp.Body)
+		if err != nil {
+			return fmt.Errorf("Unable to export custom repositories: %w", err)
+		}
+
+		if len(bodyBytes) != 0 {
+			// Saving the custom repositories in the object format
+			var result []map[string]interface{}
+			err = json.Unmarshal(bodyBytes, &result)
+			if err != nil {
+				return fmt.Errorf("Unable to export custom repositories: %w, %s", err, string(bodyBytes))
+			}
+			blueprintExportResponse.ContentSources = &result
+		}
+	}
+
 	return ctx.JSON(http.StatusOK, blueprintExportResponse)
 }
 
--- a/internal/v1/handler_blueprints_test.go
+++ b/internal/v1/handler_blueprints_test.go
@@ -7,9 +7,12 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"runtime"
+	"slices"
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/osbuild/image-builder/internal/clients/content_sources"
 
 	"github.com/google/uuid"
 	"github.com/osbuild/image-builder/internal/clients/composer"
@@ -20,13 +23,36 @@ import (
 	"github.com/osbuild/image-builder/internal/tutils"
 )
 
-func makeTestServer(t *testing.T) (dbase db.DB, srvURL string, shutdown func()) {
+type BlueprintExportResponseUnmarshal struct {
+	ContentSources []content_sources.ApiRepositoryExportResponse `json:"content_sources,omitempty"`
+	Customizations Customizations                                `json:"customizations"`
+	Description    string                                        `json:"description"`
+	Distribution   Distributions                                 `json:"distribution"`
+	Metadata       BlueprintMetadata                             `json:"metadata"`
+	Name           string                                        `json:"name"`
+}
+
+func makeTestServer(t *testing.T, apiSrv *string, csSrv *string) (dbase db.DB, srvURL string, shutdown func()) {
 	dbase, err := dbc.NewDB()
 	require.NoError(t, err)
 
-	db_srv, tokenSrv := startServer(t, &testServerClientsConf{}, &ServerConfig{
+	db_srv, tokenSrv := startServer(t, &testServerClientsConf{
+		ComposerURL: func() string {
+			if apiSrv != nil {
+				return *apiSrv
+			}
+			return ""
+		}(),
+		CSURL: func() string {
+			if csSrv != nil {
+				return *csSrv
+			}
+			return ""
+		}(),
+	}, &ServerConfig{
 		DBase:            dbase,
 		DistributionsDir: "../../distributions",
+		CSReposURL:       "https://content-sources.org",
 	})
 	return dbase, db_srv.URL, func() {
 		err := db_srv.Shutdown(context.Background())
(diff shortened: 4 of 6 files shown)


internal/v1/api.yaml
```diff
@@ -1069,6 +1070,10 @@ components:
           $ref: '#/components/schemas/Customizations'
         metadata:
           $ref: '#/components/schemas/BlueprintMetadata'
+        custom_repositories:
+          type: array
+          items:
+            $ref: '#/components/schemas/CustomRepositoryExported'
```
avitova, line 1076: I have decided to add a new column here. We are reusing the Customizations field in the API for BlueprintExportResponse, and I did not like creating a whole new "CustomizationsExport" struct that would be the same, just with different type for Custom Repositories.
Other option we have, is to make the Customizations.CustomRepositories into
`
      oneOf:
            $ref: '#/components/schemas/CustomRepository'
            $ref: '#/components/schemas/CustomRepositoryExported'
`
But honestly, I do not like that solution either. I think that we can handle one additional field here. Feel free to express your thoughts on this.

ezr-ondrej: I don't really mind the additional field here, but few thoughts
1) we now have one custom_repositories under customizations and one outside, I think it is quite confusing
2) it changes the blueprint structure, so now it won't work in the import, will it?

avitova: 1. Yes, I agree, we could do something like `custom_repositories_exported`. And we can even set the pointer to custom repositories to nil, in order to not have two custom_repositories. But I do not think it really matters.
2. It should not break anything, we ignore any extra fields. The idea was to make the import as easy as possible without creating complications. If we ever get to the point where export will be stabilized, I think we can start to be a bit more strict for import files.

croissanne:
I don't fully understand why we're adding this? Could you add that in the commit message?

internal/v1/api.yaml
lzap: Have you considered adding it to `BlueprintMetadata` component which already holds parent id and export time and date?

Another option is to simply extend the blueprint used in image builder, it already has a reference to OpenSCAP hosted resource anyway so there is the pattern already.

@croissanne asked for explanation below, so I suggest to discuss this with her.

internal/v1/api.yaml
```diff
@@ -2031,6 +2036,42 @@ components:
...
+          type: string
+        origin:
+          description: "Origin of the repository"
+          type: string
+        snapshot:
+          description: "Enable snapshotting and hosting of this repository"
+          type: boolean
+        url:
+          description: "URL of the remote yum repository"
+          type: string
```
lzap, line 2074: I wonder if image builder should store this in JSON blob simply not caring about the contents. Since the purpose of this is simply being able to recreate the same repository in a different organization, we could simply store version and JSON blob and as long as CS import/export endpoint handles the data correctly we could simply pass it in and out.

Take this as food for thought, this is fine tho we will have to maintain this here.

internal/v1/api.yaml
avitova: I did consider it. We could also put it as a separate field into customizations, something like "CustomRepositoriesExported". However, the point of the new field outside existing ones is that we will not need this field anywhere else other than in the export structure. This will just appear in the file that the users will be able to download anywhere else.
So I do not want the normal "blueprint" structure to have a new field because it would not be used anywhere. It will be a temporary field that lives only in the json file.
1) We export custom repos, including the response from /repositories/bulk_export/ into the export json
2) On import, we call the /repositories/bulk_import/ with this field in order for custom services to create these repositories for us. We can build the "custom_repositories: [xxx, xxx]" field because the account will finally have the given repositories saved. Once we do that, we throw away the CustomRepositoriesExported.

internal/v1/api.yaml
avitova: I think this is a relevant point, and we could possibly do that too. This could be easier even in cases when CS decides to change the structure of the ApiRepositoryExportResponse endpoint. However, the most readable way to send a request is again initializing an ApiRepositoryExportRequest struct, and passing in all the data. So I think we would have to unmarshal the JSON anyway.

internal/v1/api.yaml
lzap: Makes sense. LGTM

internal/v1/api.yaml
lzap: Fair enough.

lzap (dismissed):
WFM @croissanne ?

avitova:
I ended up thinking a lot about what Lukáš suggested about keeping just the json structure as is. I did a quick [check here](https://github.com/osbuild/image-builder/pull/1381/files#diff-7e4feba1f6c48c70d1b9dd9561967bfeb38cff5a11fca3714ffb8d83eae1529aR352) only to make sure that we do not include values that are not needed. So, the field would not be required; in case we actually get some values that make sense from content sources, we add them to the structure.

croissanne (dismissed):
Thank you!

I think there's some confusion about what kind of data custom repositories actually contains. Yes the ID is the uuid for the content_sources database, but it will always contain either a baseurl, mirrorlist or metalink propertly.

This is how we create custom repositories from the UI: https://github.com/osbuild/image-builder-frontend/blob/0bebb87ec0420bb0bdd91d1677a70bcb52a3184a/src/Components/CreateImageWizard/steps/Repositories/components/Utilities.ts#L8-L28.

As you can see, while the ID matches, there's all the information needed there to import properly. Including GPG related configurations if defined.


So what I would suggest is, export the data as-is. And when importing, if it finds the ID in the content sources api, great! If not, recreate the repo for the user in the content-sources api (we might need to make clear that blueprint import needs content-sources write rights configured in rbac), and then update the uuid during the import to match the new one.

But since this PR only considers exports, I don't think we need to do anything special at all?

croissanne:
@avitova here's an example of what's in the DB https://gist.github.com/croissanne/bad023411f3bb65d7e7e391427eeb43c

avitova:
Thank you for the comment, Sanne. I misinterpreted the field during the documentation, so I changed that.
However, what we were trying to solve here was the ability to import custom repositories fully. Two scenarios can happen if we export it as-is and later import it. This is due to the fact that for each URL, only one record can exist:
1) We import, the customer does not have custom repo with the given URL, so whatever we create does not influence the customer as much.
2) The user already has a repo with a given URL but has some restrictions for architecture or versions, probably for a reason. Here, we wanted to keep it as non-intrusive as possible. The import endpoint should modify the existing repo, which is why we can not use simply bulk_create endpoint. tor to make changes that modify the custom repo as little as possible, we need the information such as distribution_arch, distribution_versions, etc. Are these saved in our DB? Maybe I am missing something.

internal/v1/api.yaml
```diff
@@ -1069,6 +1069,12 @@ components:
           $ref: '#/components/schemas/Customizations'
         metadata:
           $ref: '#/components/schemas/BlueprintMetadata'
+        custom_repositories_details:
+          type: string
+          additionalProperties: true
+          description: |
+            List of custom repositories including all the repository details needed in order
+            to recreate the repositories.
```
lzap, line 1077: Are you aiming for JSON in JSON string or arbitrary JSON object (without schema)? Because this looks like the former, from what I understand arbitrary key-value pairs are represented via

```
type: object
additionalProperties: true
```

Which should be more readable and easier to work with than JSON-encoded strings. Can you paste how such export looks like?

croissanne:
> Thank you for the comment, Sanne. I misinterpreted the field during the documentation, so I changed that. However, what we were trying to solve here was the ability to import custom repositories fully. Two scenarios can happen if we export it as-is and later import it. This is due to the fact that for each URL, only one record can exist:
> 
>     1. We import, the customer does not have custom repo with the given URL, so whatever we create does not influence the customer as much.
> 
>     2. The user already has a repo with a given URL but has some restrictions for architecture or versions, probably for a reason. Here, we wanted to keep it as non-intrusive as possible. The import endpoint should modify the existing repo, which is why we can not use simply bulk_create endpoint. tor to make changes that modify the custom repo as little as possible, we need the information such as distribution_arch, distribution_versions, etc. Are these saved in our DB? Maybe I am missing something.

These are indeed missing, but maybe we can just take the arch / distro from the other parameters in the blueprint being imported? I think they should be able to edit it after.

lzap:
> These are indeed missing, but maybe we can just take the arch / distro from the other parameters in the blueprint being imported? If we could just infer to be close enough, and then I think they should be able to edit it after.

For what benefits tho?

Exporting everything maps nicely, in fact, this is a generic approach we can use for all of other services. Who knows what other fields we will not be able to infer. This is just a blob that stores next to the blueprint, it is not being stored in the image-builder database so there is nothing to worry about in terms of data duplication.

croissanne:
> > These are indeed missing, but maybe we can just take the arch / distro from the other parameters in the blueprint being imported? If we could just infer to be close enough, and then I think they should be able to edit it after.
> 
> For what benefits tho?
> 
> Exporting everything maps nicely, in fact, this is a generic approach we can use for all of other services. Who knows what other fields we will not be able to infer. This is just a blob that stores next to the blueprint, it is not being stored in the image-builder database so there is nothing to worry about in terms of data duplication.

Will repositories fail to import if the blob is not there? I guess we'd only support blueprints exported with the ib api? But right maybe my idea was too many heuristics.

internal/v1/api.yaml
```diff
@@ -1069,6 +1069,12 @@ components:
           $ref: '#/components/schemas/Customizations'
         metadata:
           $ref: '#/components/schemas/BlueprintMetadata'
+        custom_repositories_details:
+          type: string
```
ezr-ondrej, line 1073: Should this be `type: object`?

lzap:
> Will repositories fail to import if the blob is not there? I guess we'd only support blueprints exported with the ib api?

The idea behind this "special" blueprint export / import endpoint is that customers can clone self-contained blueprints. So the workflow is: this export endpoint -> JSON file -> new import endpoint (TBD). With the idea of creating some curated JSON files as an experiment for Red Hat supported blueprints.

So yes, this is a special export just for this feature. Tho hopefully if we are able to find a common blueprint schema at some point and we would go for my idea of "extensions" this extra data could live exactly there.

Alternatively, we could create a new field in customizations-custom_repositories that could carry the extra information (UUID). Have you explored that before, @avitova ?

avitova:
There has been a very long discussion, and I feel like we might get into a loop very soon. I will quickly summarize the purpose of this PR, as well as explain my choices here. I adjusted the PR based on some of yours opinions, make sure what you think now. A summary of what was going on:

### Why is this PR even a thing?
If a **Person A** from **Org A** exports their **Blueprint A**, provide the exported file of it to a **Person B** from **Org B**, we want to create as much of a true copy of Blueprint A on import, as possible.
Now, Custom repositories are specific to an org, so repos **Org A** has can be different than what **Org B** has.
=> We want to have a way to import, and also handle mismatches in case **Org B** has the repo, but with different restrictions.

### Why did I decide to use the endpoint and not use existing data in the blueprint?
So, the Customizations.CustomRepositories do not have some of the fields - such as restriction for OS and restriction for architecture. As Sanne said, we could inherit that from the rest of the blueprint - but that just does not give us the full information, we are kind of "guessing" what is there. If we are okay with that, we could do this. But then also, as Lukas said, there could be more fields that we are unable to "guess" or that would make it complicated.
Also, as Ondra said in one of the discussions, the schema for custom repositories could change. That is one of the reasons why having the blop is more beneficial than my previous solution, which was creating a new structure. [(Lukas suggested this too)](https://github.com/osbuild/image-builder/pull/1381#discussion_r1812485942)

### Now, why did I create a separate field and did not save it in Customizations?
Well, because we will not use these data anywhere else than in export, and if we change Customizations, the field will potentially be everywhere, and might even be misused. Keeping it only in BlueprintExportResponse, sounds safe.

internal/v1/api.yaml
avitova: Done, this is now an array of objects. :partying_face:

internal/v1/api.yaml
```diff
@@ -1069,6 +1069,14 @@ components:
           $ref: '#/components/schemas/Customizations'
         metadata:
           $ref: '#/components/schemas/BlueprintMetadata'
+        custom_repositories_details:
```
lzap, line 1072: One observation - this export is closer to console dot than the IB blueprint, this could be named content_sources or something like that.

internal/v1/handler_blueprints.go
```diff
@@ -312,6 +315,56 @@ func (h *Handlers) ExportBlueprint(ctx echo.Context, id openapi_types.UUID) erro
...
+			return fmt.Errorf("Unable to export custom repositories: %w", err)
+		}
+
+		if len(bodyBytes) != 0 {
+			// Checking the contents of content sources response
+			var exportedRepositories []content_sources.ApiRepositoryExportResponse
+			err = json.Unmarshal(bodyBytes, &exportedRepositories)
+			if err != nil {
+				return fmt.Errorf("Unable to export custom repositories: %w, %s", err, string(bodyBytes))
+			}
```
lzap, line 357: I do not think we should be in the business of validating against schema, this service should not literally care about what is in there. If it does not import back = error and ask for re-export.

lzap (changes requested):
Looks good, few nitpicks.

But I have a request - would you mind creating a test that literally serializes the export as a JSON file and then compares is against a JSON fixture checked in the git repository? As long as you pretty-print keys-sorted file, the file should be exactly the same.

The reason for this is I would love to see how the file actually looks, I am scared to death by string-encoded JSONs and I believe they are not here, I just want to make sure they are never there. This patch makes a template for other services in the future, people will be copy-pasting the code including the tests so we can make it bullet-proof that the files looks okay.

avitova:
Okay, I managed to figure the tests out, Lukas:)

internal/common/testdata/exported_blueprint.json
```diff
@@ -0,0 +1 @@
+{"content_sources":[{"gpg_key":"some-gpg-key","name":"payload","url":"http://snappy-url/snappy/baseos"}],"customizations":{"custom_repositories":[{"baseurl":["http://snappy-url/snappy/baseos"],"gpgkey":["some-gpg-key"],"id":"2531793b-c607-4e1c-80b2-fbbaf9d12790","name":"payload"}],"packages":["nginx"],"users":[{"name":"user"}]},"description":"desc","distribution":"centos-9","metadata":{"exported_at":"2024-11-01 10:00:00 +0000 UTC","parent_id":"f43a4ec2-5447-4a25-8a62-e258ff11a2d9"},"name":"blueprint"}
```
lzap, line 1: I was hoping for a formatted JSON but this is okay :-)

internal/v1/handler_blueprints.go
```diff
@@ -27,6 +30,7 @@ var (
 	blueprintNameRegex         = regexp.MustCompile(`\S+`)
 	customizationUserNameRegex = regexp.MustCompile(`\S+`)
 	blueprintInvalidNameDetail = "The blueprint name must contain at least two characters."
+	now                        = time.Now
 )
```
lzap, line 34: Oh here is a bug - this time is now a fixed time. In fact it is the time when pod was started and it will never be updated. Package-level variables are initialized during initialization of the process.

I see what you are trying to do, can't you just keep the code unchanged but in the test itself before you store or compare the JSON simply modify it and set the time to a fixture? Like a birthday of your faviourite movie star or something :-)

You can parse arbitrary JSON into an empty interface map in Go pretty easily, then you can manipulate it too. I think it is easier than creating a full stub for this handler.

internal/v1/handler_blueprints.go
```diff
@@ -308,10 +312,54 @@ func (h *Handlers) ExportBlueprint(ctx echo.Context, id openapi_types.UUID) erro
...
+		if len(bodyBytes) != 0 {
+			// Saving the custom repositories in the object format
+			var result []map[string]interface{}
+			err = json.Unmarshal(bodyBytes, &result)
+			if err != nil {
+				return fmt.Errorf("Unable to export custom repositories: %w, %s", err, string(bodyBytes))
+			}
+			blueprintExportResponse.ContentSources = &result
+		}
+	}
```
lzap, line 361: I suggest to concatenate all three if statement into one, then the read all code will not be duplicated and there will be only one level of identation.

internal/v1/handler_blueprints_test.go
```diff
@@ -837,16 +864,124 @@ func TestHandlers_GetBlueprint(t *testing.T) {
...
+
+func compareOutputExportBlueprint(t *testing.T, jsonResponse []byte) {
+	file, err := os.CreateTemp(t.TempDir(), "exported_blueprint_temp.*.json")
+	require.NoError(t, err)
+
+	jsonFile, err := os.Create(file.Name())
+	require.NoError(t, err)
+
+	_, err = jsonFile.Write(jsonResponse)
+	require.NoError(t, err)
```
lzap, line 889: There is no need to create a temporary file, Go's `Reader` and `Writer` interface is really powerful. Just write it into a bytes buffer or string, then you can use the same string in the `JSONEq` function below.

internal/v1/handler_blueprints_test.go
```diff
@@ -837,16 +864,124 @@ func TestHandlers_GetBlueprint(t *testing.T) {
...
+	fixtureData, err := os.ReadFile(exportedBlueprintFile)
+	require.NoError(t, err)
+
+	generatedData := marshalAndUnmarshalPrint(t, jsonResponse)
+	require.NoError(t, err)
+
+	expectedData := marshalAndUnmarshalPrint(t, fixtureData)
+	require.NoError(t, err)
+
+	require.JSONEq(t, string(expectedData), string(generatedData))
```
lzap, line 901: Nitpick: If this does not print the full JSON on error, can you also print it? In case it is off in the future so it is easy to update it. Ideally nicely formatted. All you need to do is to use `assert` instead of `require` and check for error.

lzap (changes requested):
Looks good, tho there is a small bug introduced now.

internal/v1/handler_blueprints_test.go
avitova: Okay, it used to print in base64, but now it should be more precise.

internal/v1/handler_blueprints.go
avitova: Thanks for catching this

ezr-ondrej (approved):
This looks great to me :+1:

internal/common/testdata/exported_blueprint.json
```diff
@@ -0,0 +1,38 @@
...
+    ]
+  },
+  "description": "desc",
+  "distribution": "centos-9",
+  "metadata": {
+    "exported_at": "2013-06-13 00:00:00 +0000 UTC",
+    "parent_id": "f43a4ec2-5447-4a25-8a62-e258ff11a2d9"
+  },
+  "name": "blueprint"
+}
```
lzap, line 38: Looks great.

internal/v1/api.yaml
```diff
@@ -1069,6 +1069,14 @@ components:
           $ref: '#/components/schemas/Customizations'
         metadata:
           $ref: '#/components/schemas/BlueprintMetadata'
+        content_sources:
```
lzap, line 1072: I like this name, short and on point.

lzap (approved):
Excellent work, I am really happy with the result. It is a good approach until we hopefully find a common blueprint format that would have some way of storing additional data, or would accept hosted data by itself.

avitova:
Thank you for the great discussion, guys! It was definitely productive!

PR #1385, opened 2024-10-28
v1: expose groups in user customization (HMS-4901)

croissanne:
When adding a user, immediately add it to the specified groups.

---

It should be fine to drop the default as the edge-api always adds the groups already
https://github.com/RedHatInsights/edge-api/blob/622244f04989159709332e67788c04296989e431/pkg/clients/imagebuilder/client.go#L353-L358

### DIFF
--- a/internal/v1/handler_compose_image.go
+++ b/internal/v1/handler_compose_image.go
@@ -751,13 +751,12 @@ func (h *Handlers) buildCustomizations(ctx echo.Context, cr *ComposeRequest, d *
 	if cust.Users != nil {
 		var users []composer.User
 		for _, u := range *cust.Users {
-			groups := &[]string{"wheel"}
 			u.RedactPassword()
 			users = append(users, composer.User{
 				Name:     u.Name,
 				Key:      u.SshKey,
 				Password: u.Password,
-				Groups:   groups,
+				Groups:   u.Groups,
 			})
 		}
 		res.Users = &users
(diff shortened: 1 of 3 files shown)


internal/v1/handler_compose_image.go
```diff
@@ -751,13 +751,12 @@ func (h *Handlers) buildCustomizations(ctx echo.Context, cr *ComposeRequest, d *
 	if cust.Users != nil {
 		var users []composer.User
 		for _, u := range *cust.Users {
-			groups := &[]string{"wheel"}
```
mvo5, line 754: I guess I'm stating the obvious here but I will do it anyway. Unless there is somewhere adding "wheel" to u.Group automatically this is a behavior change - I guess it's okay because the frontend is aware and takes care of this and we have no other consumers? Might still be nice to mention it in the commit message maybe?

lzap (changes requested):
This is a chance to improve the API docs.

I am assuming that the code currently works that for a user named `test` a primary group `test` is added and now the new field allows customization of secondary (also known as supplementary) groups.

I suggest to document all of that, details do matter when it comes to security.

@mvo5 has a point, tho I think it is okay since the patch is actually removing a group. If this was the opposite, then it would have been much bigger problem.

internal/v1/handler_compose_image.go
croissanne: Fair point, I'll do that :)

lzap (approved):

PR #1388, opened 2024-10-30
test: rhsm unregister fallback call

lzap:
It seems that network error can cause the whole pipeline to fail just because an unsuccessful registration. Let’s try it one more time if that happens.

### DIFF
--- a/test/cases/api.sh
+++ b/test/cases/api.sh
@@ -219,8 +219,9 @@ function instanceCheck() {
   set -eu
   [[ "$subscribe_org_id" == "org ID: $API_TEST_SUBSCRIPTION_ORG_ID" ]]
 
-  # Unregister subscription
-  $_ssh sudo subscription-manager unregister
+  # Unregister subscription (try one more time if it fails)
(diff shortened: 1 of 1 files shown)


test/cases/api.sh
```diff
@@ -219,8 +219,9 @@ function instanceCheck() {
   set -eu
   [[ "$subscribe_org_id" == "org ID: $API_TEST_SUBSCRIPTION_ORG_ID" ]]
 
-  # Unregister subscription
-  $_ssh sudo subscription-manager unregister
+  # Unregister subscription (try one more time if it fails)
+  $_ssh sudo subscription-manager unregister || \
+  	$_ssh sudo subscription-manager unregister
```
jrusz, line 224: What about adding a little bit of wait before trying again? Something like:
```suggestion
  $_ssh sudo subscription-manager unregister || \
  	sleep 5 && $_ssh sudo subscription-manager unregister
```

jrusz (approved):
LGTM, thanks!

ezr-ondrej (approved):

PR #1398, opened 2024-11-12
Makefile: implement running DB tests locally

schuellerf:
In the endeavor to implement pruning the customer data after two years (see https://issues.redhat.com/browse/HMS-4244 )
We first implement running the DB test locally by spinning up a container with postgres.
This is as close as possible to the implementation in the github action and enables to develop more DB tests and execute them.

Also changed here: `image-builder-db-test` is not compiled and run, but just executed as test, as we don't really need the binary

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,7 +118,7 @@ db-tests: dev-prerequisites
 	    PGHOST=localhost \
 	    PGPORT=$$($(CONTAINER_EXECUTABLE) inspect -f '{{ (index .NetworkSettings.Ports "5432/tcp" 0).HostPort }}' image-builder-test-db) \
 	    TERN_MIGRATIONS_DIR=internal/db/migrations-tern \
-	    sh -c './tools/dbtest-run-migrations.sh ; ./tools/dbtest-entrypoint.sh'
+	    ./tools/dbtest-entrypoint.sh
 	# we'll leave the image-builder-test-db container running
 	# for easier inspection is something fails
 	@echo "The database is available for inspection at"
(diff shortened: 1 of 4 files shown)


croissanne (approved):

Makefile
```diff
@@ -83,3 +81,57 @@ push-check: generate build unit-tests
...
+db-tests: dev-prerequisites
+	-$(CONTAINER_EXECUTABLE) stop image-builder-test-db 2>/dev/null || echo "DB already stopped"
+	-$(CONTAINER_EXECUTABLE) rm image-builder-test-db 2>/dev/null || echo "DB already removed"
+	$(CONTAINER_EXECUTABLE) run -d \
+      --name image-builder-test-db \
+      --env POSTGRES_USER \
+      --env POSTGRES_PASSWORD \
+      --env POSTGRES_DB \
+      --publish :5432 \
+      postgres:12
```
lzap, line 113: Idea: if there is a way to provide `fsync=off` configuration to postgres, this significantly increases performance for any kind of testing.

schuellerf: although the tests are really quick - I'll definitely try this!

lzap: Just for the record, I did mean an option in `postgresql.conf` not sure it is ever exposed for this container:

https://www.postgresql.org/docs/current/runtime-config-wal.html#GUC-FSYNC

schuellerf: I tried it in #1401. you can set this in `postgresql.conf` or pass as `-c` command line argument
https://github.com/osbuild/image-builder/pull/1401/commits/dfa1096d61e3b2288c49fa6213d0ecd60bb4e89f
but after postgres 13 (I think) the behavior changed and impacts this special case where I try to "test vacuum"
but providing `-c fsync=off` seems to change behavior in general.
Just does not work for the tests I'm intending to run
If it turns out that we can't really test our "vacuum implementation" I'll re-enable and check again if it's faster

lzap: No need to spend much time on the `fsync` thing it was just a minor idea. Saw the cleanup PR looks great.

PR #1400, opened 2024-11-13
Add rhel 10 beta (HMS-4990)

croissanne:
Adds rhel 10 beta in all the necessary places.


distributions/rhel-10-beta/rhel-10-beta.json
```diff
@@ -0,0 +1,54 @@
...
+      "check_gpg": true
+    }, {
+      "id": "google-compute-engine",
+      "baseurl": "https://packages.cloud.google.com/yum/repos/google-compute-engine-el9-x86_64-stable",
+      "rhsm": false,
+      "image_type_tags": ["gcp"],
+      "check_gpg": false
```
schuellerf, line 29: why not use the GPG key and set checking to `true` it's the same URL as rhel-9-beta anyway for now
https://github.com/osbuild/image-builder/blob/main/distributions/rhel-9-beta/rhel-9-beta.json#L28

ondrejbudai: I wonder if these are actually needed at the moment. @thozza do you know?

ondrejbudai (approved):
Not sure about the EL9 GCP repos. Otherwise, this looks good! :)

distributions/rhel-10-beta/rhel-10-beta.json
thozza: Yes, because the el10 gce image still installs `google-osconfig-agent` from the repo.

thozza: WRT the GPG check, it can't be enabled before https://issuetracker.google.com/issues/360905189 is resolved. This is el10 specific issue.

thozza (approved):
LGTM

PR #1401, opened 2024-11-13
Improve speed of db tests

schuellerf:
As a followup of #1398 this improves the speed of `make db-tests`.
Thanks @lzap for the hint!

(to be merged after #1398)


schuellerf:
When I want to test the "maintenance job" including vacuum, I need to have fsync enabled. So I'd prefer to close this to enable vacuum tests.

lzap:
Had no idea, interesting.

PR #1406, opened 2024-11-18
internal/v1: export snapshot date

avitova:
I started a conversation about importing and exporting snapshot dates. Our team thought that it would be convenient to export the snapshot date. We might have a follow-up discussion about importing it as well, but there are some issues that we need to handle. 
In the wizard, we currently do not limit how much in the history the user can pick the snapshot date, so maybe we can just import it as is on the front end. @croissanne ? But then, users also import custom repositories, and they would most likely not have that snapshot from the history.
The discussion's outcome was to keep the feature without importing the snapshot date, but we will keep that option for the future. What do we think? We can mention somewhere that this value is not being imported.

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -1077,6 +1077,11 @@ components:
           description: |
             List of custom repositories including all the repository details needed in order
             to recreate the repositories.
+        snapshot_date:
+          type: string
+          description: |
+            Importing the snapshot date will not yet be supported. It is exported for informative reasons.
+            The format is YYYY-MM-DD.
     BlueprintMetadata:
       required:
         - parent_id
--- a/internal/common/testdata/exported_blueprint.json
+++ b/internal/common/testdata/exported_blueprint.json
@@ -6,6 +6,7 @@
       "url": "http://snappy-url/snappy/baseos"
     }
   ],
+  "snapshot_date": "2012-12-20",
   "customizations": {
     "custom_repositories": [
       {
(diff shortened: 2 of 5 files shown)


ezr-ondrej (approved):
Thanks @avitova ! I think it's great :+1: Would love to hear @croissanne opinion before merge tho :)

internal/v1/api.yaml
```diff
@@ -1077,6 +1077,11 @@ components:
           description: |
             List of custom repositories including all the repository details needed in order
             to recreate the repositories.
+        snapshot_date:
+          type: string
+          description: |
+            Importing the snapshot date will not yet be supported. It is exported for informative reasons.
+            The format is YYYY-MM-DD.
```
lzap, line 1084: Isntt there a date type in OpenAPI JSON Schema?

lzap (approved):

internal/v1/api.yaml
avitova: Might be, but I would probably leave this in the same format that we have in the image requests `snapshot_date` field for consistency.

lzap:
@croissanne any comments? Thanks.

croissanne (approved):
lgtm! ty. In the FE you can indeed choose whatever snapshot date, but it will pick the closest snapshot it has. So if you enter 1970 as a date, it'll pick whatever oldest snapshot it has.

avitova:
Thanks!

PR #1411, opened 2024-11-22
v1/api: Add on prem boolean to blueprints metadata

avitova:
This way we might be able to track how much are people using the on prem blueprints using.

[HMS-4980](https://issues.redhat.com/browse/HMS-4980)

### DIFF
--- a/internal/common/testdata/exported_blueprint.json
+++ b/internal/common/testdata/exported_blueprint.json
@@ -32,7 +32,8 @@
   "distribution": "centos-9",
   "metadata": {
     "exported_at": "2013-06-13 00:00:00 +0000 UTC",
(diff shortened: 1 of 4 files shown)


ezr-ondrej:
Can you also add it to the [floorist query](https://github.com/osbuild/image-builder/blob/675fc79b42ee5fac67d3b8ca2f5b8422a0f72db8/templates/image-builder.yml#L213) pls? :)

avitova:
Woah! You are quick! Yeah, that is a good point. I did that:) I think we agreed on this, if you feel like there are better ways, feel free to comment.

ezr-ondrej (approved):
the name is bit confusing, but I don't have a better one, so let's get it in :+1:

avitova:
Amazing, thanks!

PR #1434, opened 2024-12-10
v1: rework tests into separate package

croissanne:
- separate test pkg
- separate pkg for mocks
- added content sources mock which enforces correct behaviour
- make db accessible to tests, avoiding the need to create new db in each test


internal/v1/server.go
```diff
@@ -286,3 +286,8 @@ func (s *Server) getDistro(ctx echo.Context, distro Distributions) (*distributio
 	}
 	return d, nil
 }
+
+// used to access DB directly in tests
+func getDB(s *Server) db.DB {
```
lzap, line 291: Nitpick: [Avoid getters/setters](https://go.dev/doc/effective_go#Getters).

lzap (approved):

lzap:
FYI I already contacted Stephen about the ephemeral issue with CS:

```
15:43:04 2024-12-10 15:43:04 [   ERROR] [           thread-46] [clowdapp/content-sources-backend] timed out waiting for resource to be ready, details:   clowdapp/content-sources-backend not ready, status conditions:
15:43:04   - DeploymentsReady: False (Deployments are not yet ready)
15:43:04   - ReconciliationFailed: False
15:43:04   - ReconciliationSuccessful: True
15:43:04   deployment/content-sources-backend-task-worker not ready, status conditions:
15:43:04   - Available: False (Deployment does not have minimum availability.)
15:43:04   - Progressing: False (ReplicaSet "content-sources-backend-task-worker-55f48cf4db" has timed out progressing.)
```

@avitova anything that catches an eye in this refactoring?

avitova:
I do not see anything that would seem odd. But I am a bit curious why we want to do this.

lzap:
Can't speak for Sanne, but there are two approaches you can take with testing. Keep it in the same package and then accessing internals, useful for testing unexported functionality. And pulling it out of a package, then you can only test public API. Useful to ensure you have proper stubs and interfaces, tho I agree that keeping tests in the same package is something that had grown for me lately. No hard feelings here, tho, accessing DB like that is a bit shady.

internal/v1/server.go
croissanne: Right, but this is just to have a function to export to the test package. Alternatively i'll just make the db on the server struct public.

croissanne: found a nicer way

croissanne:
> I do not see anything that would seem odd. But I am a bit curious why we want to do this.

Mostly to increase separation between the tests and the v1 package. As we keep bolting stuff onto v1, it's good to pull stuff out now and then to make it easier to read.

Furthermore I think the single content sources mock reduces the amount of mocking we have to do in each test function. Other dependencies like provisioning could get the same treatment.

I'm currently adding support for upload repositories in content sources, and I just found it quite hard to expand the tests again in the same fashion, where I'd just bolt on more stuff without a little restructuring.

internal/common/testdata/exported_blueprint.json
```diff
@@ -1,8 +1,8 @@
 {
   "content_sources": [
     {
-      "gpg_key": "some-gpg-key",
-      "name": "payload",
```
mvo5, line 5: (nitpick) here the name changes from "payload" -> "baseos" which is good but the commit message is a bit terse and just says " v1/tests: don't declare constants which are only used once which is not the full picture of this commit AFAICT

internal/v1/handler_blueprints_test.go
```diff
@@ -34,20 +34,14 @@ type BlueprintExportResponseUnmarshal struct {
...
-func makeTestServer(t *testing.T, apiSrv *string, csSrv *string) (dbase db.DB, srvURL string, shutdown func(t *testing.T)) {
+func makeTestServer(t *testing.T, apiSrv *string) (dbase db.DB, srvURL string, shutdown func(t *testing.T)) {
 	srv := startServer(t, &testServerClientsConf{
 		ComposerURL: func() string {
 			if apiSrv != nil {
 				return *apiSrv
 			}
 			return ""
 		}(),
-		CSURL: func() string {
```
mvo5, line 48: Same comment as above, the commit message and the content seem to not align here. Maybe just add something like `and move csSrv directly into testServer struct` or something.

mvo5 (commented):
Thank you! I love this PR, tiny nitpick about the commit message of the last commit for your consideration - but everything looks so much nicer after this refactor \o/

lzap (approved):

croissanne:
I did a weird rebase at some point and the last two commits got squashed. I separated them again.

mvo5 (approved):
Thank you

lzap (approved):

PR #1443, opened 2024-12-17
v1: support specifying snapshotted, payload and custom repositories by uuid

croissanne:


internal/clients/content_sources/client.go
```diff
@@ -48,26 +50,79 @@ func (csc *ContentSourcesClient) request(method, url string, headers map[string]
...
+	err = json.NewDecoder(resp.Body).Decode(&repos)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to parse repositories: %v", err)
+	}
+
+	return repos, nil
+}
+
+func (csc *ContentSourcesClient) GetRepositories(ctx context.Context, repoURLs []string, repoIDs []string, external bool) (RepositoryByID, error) {
+	result := RepositoryByID{}
```
lzap, line 103: The map could be preallocated since the result capacity is already known, just a nitpick.

internal/clients/content_sources/client.go
```diff
@@ -48,26 +50,79 @@ func (csc *ContentSourcesClient) request(method, url string, headers map[string]
...
+		if err != nil {
+			return nil, err
+		}
+		for _, repo := range *repos.Data {
+			result[*repo.Uuid] = repo
+		}
+	}
+
+	return result, nil
 }
```
lzap, line 126: Not huge fan of putting so much logic into a HTTP client. Ideally, the whole `client.go` could be auto-generated from OpenAPI, I remember we sligthly discussed it on Slack and I forgot what was the reason not to generate it. But there is a lot of boiler plate that could be generated.

internal/v1/handler_compose_image.go
```diff
@@ -216,44 +216,24 @@ func buildRepositories(arch *distribution.Architecture, imageType ImageTypes) []
...
-	}
-
-	var csRepos content_sources.ApiRepositoryCollectionResponse
-	err = json.NewDecoder(resp.Body).Decode(&csRepos)
-	if err != nil {
-		return nil, nil, fmt.Errorf("Unable to parse repositories: %v", err)
-	}
 
-	for _, repo := range *csRepos.Data {
+	repoUUIDs := []string{}
```
lzap, line 231: Could be preallocated too.

lzap (dismissed):
LGTM I haven’t tried this myself tho.

lzap:
Not sure if you meant to fix those nitpicks, I see no changes. It needs a rebase once again FYI.

croissanne:
> Not sure if you meant to fix those nitpicks, I see no changes. It needs a rebase once again FYI.

still working on the integration test

internal/clients/content_sources/client.go
croissanne: Currently it's a balance between putting stuff in `v1` and `clients`. We could probably autogenerate some stuff and then have some logic around it in `clients`.

PR #1418, opened 2024-11-25
vendor: remove the vendored dependencies

mvo5:
Similar to what we did in `images` in PR#861 (thanks to @achilleas-k)
this commit removes the vendoring from image-builder. We don't
package image-builder so just like with images we don't need the
vendor dir (go.{mod,sum} is fine to re-create the environment).

C.f. https://github.com/osbuild/images/pull/1041 and https://github.com/osbuild/image-builder/pull/1402#issuecomment-2498333606

PR #1452, opened 2025-01-07
internal: Support both DateOnly and RFC3339 format for snapshot dates

regexowl:
This adds back support for DateOnly format of snapshot dates.


regexowl:
We'll need to support both DateOnly and RFC3339 date format at least temporarily to not break older blueprints. Hope this is not too ugly. Tried to add a test case as well.

lzap (approved):
LGTM

Please do not merge before https://github.com/osbuild/image-builder/pull/1418 we want to get rid of the vendor directory, yay! We will ask you to rebase on top of that change, easy merge just delete `vendor/` alltogether.

PR #1481, opened 2025-02-14
internal: add cacerts into customizations

avitova:


croissanne (changes requested):
Api looks ok to me, but this will need an update in ```func (h *Handlers) buildCustomizations(ctx echo.Context, cr *ComposeRequest, d *distribution.DistributionFile) (*composer.Customizations, error)` as well.

avitova:
Good point; thanks, Sanne! We mainly need the type for using it on frontend now, but this might have been easily forgotten!

lzap:
You need https://github.com/osbuild/image-builder-crc/pull/1482/files#diff-423d98a85dc577c76a5be2a77222ea131b8fd747bdeb534b221a0e00e1b665a1R21 to get merged in order to re-run prepare source to fix this PR. Sorry, I may have introduced this.

With Go 1.24 all these problems will go away with its `go get -tool` command.

lzap (approved):
Other than that, LGTM.

lzap:
Rebased, the CICD should be fixed now.

ezr-ondrej:
> Api looks ok to me, but this will need an update in `func (h *Handlers) buildCustomizations(ctx echo.Context, cr *ComposeRequest, d *distribution.DistributionFile) (*composer.Customizations, error)` as well.

This doesn't seem to have been done yet, right @avitova ? I've rebased for you here, but this point seem to still stand.

ezr-ondrej:
I thinks a small test for posting compose (and/or blueprint) with ca cert included and verifying it was saved might make sense, right? :)

avitova:
Yeah, that's a good point. I just tried to add a simple certificate. I think in case we want to do the validation on the frontend, this should be enough, right? We know that the POST compose can be done. The test basically follows what we do for other fields. Let me know if you think something is missing here.

ezr-ondrej (approved):
Exactly what I had in mind! :) Thank you! :pray:

PR #1482, opened 2025-02-14
Adapt to rename

croissanne:
Various areas of the code have to be updated. The go module, the workflows etc.


lzap (approved):

tools/prepare-source.sh
```diff
@@ -18,7 +18,7 @@ test -e "$TOOLS_PATH/goimports" || GOBIN=$TOOLS_PATH $GO_BINARY install golang.o
 GOBIN=$TOOLS_PATH $GO_BINARY generate -x -mod=mod ./...
 
 # ... the code is formatted correctly, ...
-goimports -w internal cmd
+"$TOOLS_PATH/goimports" -w internal cmd
```
lzap, line 21: Please drop this change.

I just realized there is a problem with the script when goimports is missing. Let me fix that in a different patch.

schuellerf (approved):
Thanks alot! LGTM

PR #1485, opened 2025-02-18
clients/content_sources: fix baseurl for upload repositories

croissanne:
Whenever snapshot URLs are used, they need to be used with the internal URL, as it's content hosted by content sources.

### DIFF
--- a/internal/clients/content_sources/util.go
+++ b/internal/clients/content_sources/util.go
@@ -2,15 +2,24 @@ package content_sources
 
 import (
 	"fmt"
+	"net/url"
 )
 
-func GetBaseURL(repo ApiRepositoryResponse) (string, error) {
+func GetBaseURL(repo ApiRepositoryResponse, csReposURL *url.URL) (string, error) {
 	if repo.Origin == nil {
 		return "", fmt.Errorf("unable to read origin from repository %s", *repo.Uuid)
 	}
 	switch *repo.Origin {
 	case "upload":
-		return *repo.LatestSnapshotUrl, nil
+		if csReposURL == nil {
+			return "", fmt.Errorf("upload repositories require a content sources URL")
+		}
+		// Snapshot URLs need to be replaced with the internal mtls URL
+		repoURL, err := url.Parse(*repo.LatestSnapshotUrl)
+		if err != nil {
+			return "", err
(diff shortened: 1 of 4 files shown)


lzap (dismissed):

.golangci.yml
```diff
@@ -1,9 +1,7 @@
...
-exclude-dirs:
-  - vendor
 linters:
   enable:
   - bodyclose
-run:
-  skip-dirs:
+issues:
+  exclude-dirs:
     - vendor
```
lzap, line 7: I removed this in completely since `vendor/` is gone, not big deal if this comes first I can rebase and remove again:

https://github.com/osbuild/image-builder-crc/pull/1476

internal/clients/content_sources/util.go
```diff
@@ -2,15 +2,24 @@ package content_sources
...
-func GetBaseURL(repo ApiRepositoryResponse) (string, error) {
+func GetBaseURL(repo ApiRepositoryResponse, csReposURL *url.URL) (string, error) {
 	if repo.Origin == nil {
 		return "", fmt.Errorf("unable to read origin from repository %s", *repo.Uuid)
 	}
 	switch *repo.Origin {
 	case "upload":
-		return *repo.LatestSnapshotUrl, nil
+		if csReposURL == nil {
+			return "", fmt.Errorf("upload repositories require a content sources URL")
```
lzap, line 15: Nit pick: returning dynamic errors lead to inability to unwrap them and check for values or types. AWS SDK does this and the end result is parsing error strings to find root cause. This simple trick is idiomatic Go to solve this:

```
// exported error value
var ErrMissingURL = errors.New("upload repositories require a content sources URL")

// ...
return "", ErrMissingURL
```

lzap (approved):

.golangci.yml
lzap: Apologies, the other PR got merged, please `rebase -Xtheirs main`, cheers.

internal/clients/content_sources/util.go
croissanne: I agree, but here the err value isn't checked, just the fact that there is an error. And I think this are all handled as 500s by default.

PR #1489, opened 2025-02-20
v1/api: support adding content template to a blueprint

xbhouse:
To add support for adding a content template to a blueprint, this PR does a few things:

- Adds a new field (`template`) to the blueprint customizations, which holds the template UUID
- Saves the template UUID and snapshot date from the template to the blueprint on creation
- Resolves custom repositories in the template at compose time
- Adds another content sources mock for the unit tests

Ticket here: https://issues.redhat.com/browse/HMS-5319

Open question(s):

- How should Red Hat repositories in the template be handled? Currently I've filtered out and added only the custom repos, I'm not sure if this is what we want though :)

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -1597,6 +1597,9 @@ components:
           items:
             $ref: '#/components/schemas/CustomRepository'
           description: List of custom repositories.
+        template:
+          type: string
+          description: ID of template.
         openscap:
           $ref: '#/components/schemas/OpenSCAP'
         filesystem:
--- a/internal/v1/handler_compose_image.go
+++ b/internal/v1/handler_compose_image.go
@@ -74,7 +74,7 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
 	}
 
 	var repositories []composer.Repository
-	if composeRequest.ImageRequests[0].SnapshotDate != nil {
+	if composeRequest.ImageRequests[0].SnapshotDate != nil && composeRequest.Customizations.Template == nil {
 		repoURLs := []string{}
 		for _, r := range arch.Repositories {
 			// Assume that non-rh repositories that are defined in the distributions file will not be snapshotted,
@@ -456,6 +456,85 @@ func (h *Handlers) buildCustomRepositories(ctx echo.Context, custRepos []CustomR
 	return res, nil
 }
 
+func (h *Handlers) buildTemplateRepositories(ctx echo.Context, templateID string) ([]composer.Repository, []composer.CustomRepository, error) {
+	var rhRepoIDs []string
+	var customRepoIDs []string
+	template, err := h.server.csClient.GetTemplateByID(ctx.Request().Context(), templateID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("Unable to retrieve template: %v", err)
+	}
+	if template.RepositoryUuids != nil {
+		for _, id := range *template.RepositoryUuids {
+			// Check if repo is Red Hat
+			rhRepoMap, err := h.server.csClient.GetRepositories(ctx.Request().Context(), []string{}, []string{id}, false)
+			if err != nil {
+				return nil, nil, fmt.Errorf("Unable to retrieve Red Hat repositories: %v", err)
+			}
+			// Separate Red Hat repos from custom
+			if len(rhRepoMap) > 0 {
+				rhRepoIDs = append(rhRepoIDs, id)
+			} else {
+				customRepoIDs = append(customRepoIDs, id)
+			}
+		}
+	} else {
+		return nil, nil, fmt.Errorf("Template %v has no repositories", templateID)
+	}
+
+	repoMap, err := h.server.csClient.GetRepositories(ctx.Request().Context(), []string{}, customRepoIDs, true)
+	if err != nil {
+		return nil, nil, fmt.Errorf("Unable to retrieve external repositories: %v", err)
+	}
+
+	var repositories []composer.Repository
+	var customRepositories []composer.CustomRepository
+	if template.Snapshots != nil {
+		for _, snap := range *template.Snapshots {
+			if slices.Contains(customRepoIDs, *snap.RepositoryUuid) {
+				repo, ok := repoMap[*snap.RepositoryUuid]
+				if !ok {
+					return repositories, customRepositories, fmt.Errorf("Returned snapshot %v unexpected repository id %v", *snap.Uuid, *snap.RepositoryUuid)
+				}
+
+				composerRepo := composer.Repository{
+					Baseurl: common.ToPtr(h.server.csReposURL.JoinPath(*snap.RepositoryPath).String()),
+					Rhsm:    common.ToPtr(false),
+				}
+
+				if repo.GpgKey != nil && *repo.GpgKey != "" {
+					composerRepo.Gpgkey = repo.GpgKey
+				}
+				if composerRepo.Gpgkey != nil && *composerRepo.Gpgkey != "" {
+					composerRepo.CheckGpg = common.ToPtr(true)
+				}
+				composerRepo.ModuleHotfixes = repo.ModuleHotfixes
+				composerRepo.CheckRepoGpg = repo.MetadataVerification
+				repositories = append(repositories, composerRepo)
+
+				// Don't enable custom repositories, as they require further setup to be useable.
+				customRepo := composer.CustomRepository{
+					Id:      *snap.RepositoryUuid,
+					Name:    repo.Name,
+					Baseurl: &[]string{*snap.Url},
+					Enabled: common.ToPtr(false),
+				}
+				if repo.GpgKey != nil && *repo.GpgKey != "" {
+					customRepo.Gpgkey = &[]string{*repo.GpgKey}
+					customRepo.CheckGpg = common.ToPtr(true)
+				}
+				customRepo.ModuleHotfixes = repo.ModuleHotfixes
+				customRepo.CheckRepoGpg = repo.MetadataVerification
+				customRepositories = append(customRepositories, customRepo)
+			}
+		}
+	} else {
+		return nil, nil, fmt.Errorf("Template %v has no snapshots", templateID)
+	}
+
+	ctx.Logger().Debugf("Resolved repository snapshots from template: %v", repositories)
+	return repositories, customRepositories, nil
+}
+
 func (h *Handlers) buildUploadOptions(ctx echo.Context, ur UploadRequest, it ImageTypes) (composer.UploadOptions, composer.ImageTypes, error) {
 	var uploadOptions composer.UploadOptions
 	switch ur.Type {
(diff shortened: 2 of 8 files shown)


internal/v1/handler_compose_image.go
```diff
@@ -74,7 +74,7 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
 	}
 
 	var repositories []composer.Repository
-	if composeRequest.ImageRequests[0].SnapshotDate != nil {
+	if composeRequest.ImageRequests[0].SnapshotDate != nil && composeRequest.Customizations.Template == nil {
```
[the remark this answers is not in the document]
xbhouse: if i understand correctly, this portion is for building the snapshots when custom repos are added directly. when a template is added to the request, the custom repos are extracted from the template (via buildTemplateRepositories) so this logic is not needed. 

i can add a name to the condition like `customReposChosen := composeRequest.ImageRequests[0].SnapshotDate != nil && composeRequest.Customizations.Template == nil` ? or just `hasSnapshotDate` and `hasTemplate` if that makes it more clear? i'm horrible with names :)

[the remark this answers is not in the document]
xbhouse: @croissanne i could definitely be misunderstanding how this works, does this seem right?

internal/v1/handler_compose_image.go
```diff
@@ -456,6 +456,85 @@ func (h *Handlers) buildCustomRepositories(ctx echo.Context, custRepos []CustomR
...
+				rhRepoIDs = append(rhRepoIDs, id)
+			} else {
+				customRepoIDs = append(customRepoIDs, id)
+			}
+		}
+	} else {
+		return nil, nil, fmt.Errorf("Template %v has no repositories", templateID)
+	}
+
+	repoMap, err := h.server.csClient.GetRepositories(ctx.Request().Context(), []string{}, customRepoIDs, true)
```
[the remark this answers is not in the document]
xbhouse: good point! it doesn't :) added a check for that

internal/v1/handler_compose_image.go
```diff
@@ -456,6 +456,85 @@ func (h *Handlers) buildCustomRepositories(ctx echo.Context, custRepos []CustomR
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("Unable to retrieve external repositories: %v", err)
+	}
+
+	var repositories []composer.Repository
+	var customRepositories []composer.CustomRepository
+	if template.Snapshots != nil {
+		for _, snap := range *template.Snapshots {
+			if slices.Contains(customRepoIDs, *snap.RepositoryUuid) {
+				repo, ok := repoMap[*snap.RepositoryUuid]
```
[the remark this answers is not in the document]
xbhouse: snapshots should always have both a repo uuid and repo path, but to be extra careful i added a check in case they don't for some reason

internal/v1/api.yaml
```diff
@@ -1601,6 +1601,9 @@ components:
           items:
             $ref: '#/components/schemas/CustomRepository'
           description: List of custom repositories.
+        template:
+          type: string
+          description: ID of template.
```
croissanne, line 1606: I'd actually make this part of ImageRequest, not of its customisations.

Also maybe say that you can't have both a snapshot date and a template specified? Maybe even oneOf? Not sure.

internal/v1/handler_compose_image.go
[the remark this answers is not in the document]
croissanne: @avitova  is correct, we sould accept either a snapshot date or a content template, but not both.

internal/v1/handler_compose_image.go
```diff
@@ -456,6 +456,85 @@ func (h *Handlers) buildCustomRepositories(ctx echo.Context, custRepos []CustomR
...
 }
 
+func (h *Handlers) buildTemplateRepositories(ctx echo.Context, templateID string) ([]composer.Repository, []composer.CustomRepository, error) {
+	var rhRepoIDs []string
+	var customRepoIDs []string
+	template, err := h.server.csClient.GetTemplateByID(ctx.Request().Context(), templateID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("Unable to retrieve template: %v", err)
+	}
+	if template.RepositoryUuids != nil {
```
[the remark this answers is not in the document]
croissanne: So the content sources API has pointers for everything (annoyingly). In general if something comes from the content sources api we just assume the value not nil (even if that is a pointer to an empty value).

internal/v1/handler_compose_image.go
```diff
@@ -467,6 +467,85 @@ func (h *Handlers) buildCustomRepositories(ctx echo.Context, custRepos []CustomR
 	return res, nil
 }
 
+func (h *Handlers) buildTemplateRepositories(ctx echo.Context, templateID string) ([]composer.Repository, []composer.CustomRepository, error) {
```
croissanne, line 470: This is what I had in mind indeed, and we should call either buildSnapshotRepositories or buildTemplateRepositories.  There might be a lot of overlap between the two however, so depending on what that looks like maybe we can refactor a little.

internal/v1/handler_compose_image.go
```diff
@@ -726,51 +805,59 @@ func (h *Handlers) buildCustomizations(ctx echo.Context, cr *ComposeRequest, d *
...
 		if err != nil {
 			return nil, err
 		}
-		res.PayloadRepositories = &payloadRepositories
-	} else if cust.PayloadRepositories != nil && len(*cust.PayloadRepositories) > 0 {
-		plrepos, err := h.buildPayloadRepositories(ctx, *cust.PayloadRepositories)
-		if err != nil {
-			return nil, err
+		res.CustomRepositories = &templateRepositories
+	} else {
```
croissanne, line 826: This isn't correct, there are 3 scenarios:

1. content template specified: use buildTemplateRepositories
2. snapshot date specified: use buildSnapshotRepositories
3. no snapshot date or content template: use the customizations directly

internal/v1/handler_post_compose_test.go
```diff
@@ -2821,6 +2821,124 @@ func TestComposeCustomizations(t *testing.T) {
...
+					{
+						Architecture: "x86_64",
+						ImageType:    v1.ImageTypesGuestImage,
+						UploadRequest: v1.UploadRequest{
+							Type:    v1.UploadTypesAwsS3,
+							Options: uo,
+						},
+						SnapshotDate: common.ToPtr("2000-01-30T00:00:00Z"),
+					},
+				},
```
croissanne, line 2841: I think specifying a template and a snapshot date shouldn't be allowed. It's either/or, as the date is already present in the template.

croissanne (changes requested):
Thank you, definitely the right direction.

internal/v1/handler_compose_image.go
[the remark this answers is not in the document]
croissanne: Indeed, the content sources api's spec results in generated go code with pointers for everything, since they don't mark values as required. But we kinda assume all of the values are set or at least not nil.

internal/v1/handler_compose_image.go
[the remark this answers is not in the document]
avitova: Yeah, but what was my point here is, that it is not clear from the code, why is that. If I was supposed to fix some bug around this area of code, I'd be quite confused, and would be quite hesitant to touch this, as I would not be sure about the meaning. So my point is, maybe we could add a comment or something, to explain this condition briefly. :thinking:

internal/v1/api.yaml
lzap: > Also maybe say that you can't have both a snapshot date and a template specified? Maybe even oneOf? Not sure.

Only if it makes sense to build images (and register systems) to particular snapshot dates? With introduction of templates, I think it makes sense to only restrict this to templates although technically snapshots are also repositories.

Also not sure, just food for thoughts.

lzap:
Apologies for late response, folks are already on it so not gonna do formal one, I will follow the work tho. Thanks!

internal/v1/handler_compose_image.go
```diff
@@ -467,6 +467,97 @@ func (h *Handlers) buildCustomRepositories(ctx echo.Context, custRepos []CustomR
...
+				customRepo.Gpgkey = &[]string{*repo.GpgKey}
+				customRepo.CheckGpg = common.ToPtr(true)
+			}
+			customRepo.ModuleHotfixes = repo.ModuleHotfixes
+			customRepo.CheckRepoGpg = repo.MetadataVerification
+			customRepositories = append(customRepositories, customRepo)
+		}
+	}
+
+	ctx.Logger().Debugf("Resolved repository snapshots from template: %v", repositories)
```
lzap, line 557: Just a sidenote, we have just integrated our https://github.com/osbuild/logging library and we are in the transition from `logrus` to `slog`. You can either use `echo` logging API, which is this line so all is good, or use `slog` to do all logging. Just avoid using `logrus` completely, just in case you add some more logging into this patch, thanks.

internal/v1/handler_compose_image.go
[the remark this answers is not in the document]
xbhouse: yep makes sense and i agree with you :) adding a clarifying comment here would be helpful, i'll do that

internal/v1/handler_compose_image.go
xbhouse: hmm in that case, i think maybe i've also missed overwriting the payload repositories (i'm not actually sure what these are :thinking: ) if a template if a specified. i don't think custom repos would be specified if a template is, is that the same for payload repos?

internal/v1/handler_post_compose_test.go
xbhouse: that makes sense, i added it to the compose request here since i was extracting the date from the template on creation of a blueprint and adding it to the image request. but that's probably not necessary

internal/v1/handler_compose_image.go
```diff
@@ -467,6 +471,97 @@ func (h *Handlers) buildCustomRepositories(ctx echo.Context, custRepos []CustomR
...
+		return nil, nil, fmt.Errorf("Template %v has no repositories", templateID)
+	}
+	for _, id := range *template.RepositoryUuids {
+		// Check if repo is Red Hat
+		rhRepoMap, err := h.server.csClient.GetRepositories(ctx.Request().Context(), []string{}, []string{id}, false)
+		if err != nil {
+			return nil, nil, fmt.Errorf("Unable to retrieve Red Hat repositories: %v", err)
+		}
+		// Separate Red Hat repos from custom
+		if len(rhRepoMap) > 0 {
```
xbhouse, line 492: if there are Red Hat repositories in the template, should they also be considered "custom repositories"?

lzap: I'd say no, @croissanne ?

croissanne: Yea I would think not, but I'm not entirely sure actually. @jlsherrill wdyt?

jlsherrill: I would think they would not be custom repositories

internal/v1/handler_compose_image.go
```diff
@@ -74,7 +78,8 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
 	}
 
 	var repositories []composer.Repository
-	if composeRequest.ImageRequests[0].SnapshotDate != nil {
+	// We should only build repository snapshots in the case where there is a snapshot date and no content template
+	if composeRequest.ImageRequests[0].SnapshotDate != nil && composeRequest.ImageRequests[0].ContentTemplate == nil {
```
croissanne, line 82: I think a condition needs to be added here to handle a non-nil content template:

1. content template is set => all repositories should come from the template;
2. else if snapshotdate is set => the snapshots should be resolved like is already the case;
3. else the default behaviour.

internal/v1/handler_compose_image.go
```diff
@@ -705,12 +800,25 @@ func validateComposeRequest(cr *ComposeRequest) error {
...
+		}
+		res.CustomRepositories = &templateRepositories
+	}
+
 	cust := cr.Customizations
 	if cust == nil {
+		// If no customizations requested, just return the repository snapshots resolved from the content template if one was specified
+		if res.CustomRepositories != nil {
+			return res, nil
+		}
```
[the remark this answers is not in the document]
xbhouse: i think i need some clarification on the difference between custom repos, repos, and payload repos 😅 i understand what custom repos are. what are the image request repositories and payload repositories and where do they come from?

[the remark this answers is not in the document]
croissanne: Sure thing:

1. the repositories in the image request are the _base_ repositories (so for the kernel, systemd etc.) that are used to build the image, these are generally baseos and appstream in the case of rhel;
2. the payload repositories are also used during the image build, but only for stuff 'on top', so any repositories that the user might include really;
3. the custom repositories are just saved to /etc/yum.repos.d, so they're available to the user once the system is booted up, nothing is actually pulled from them during the image build.

All 3 will need to be set via the content template, albeit disabled in the case of the custom repositories.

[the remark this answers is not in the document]
croissanne: Indeed, the baseos and appstream repos from the template would be the image request repositories, and then payload and custom repositories are the same, but custom repositories have a slightly different syntax.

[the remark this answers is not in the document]
xbhouse: alrighty, i pushed up the changes to set all 3 types. let me know if that looks ok @croissanne

internal/v1/handler_compose_image.go
xbhouse: pushed up a change to set RH repos here, doing it outside of buildCustomizations (since we're not touching those here) but using the same buildTemplateRepositories method

internal/v1/handler_compose_image.go
```diff
@@ -705,12 +820,26 @@ func validateComposeRequest(cr *ComposeRequest) error {
...
+		res.PayloadRepositories = &templatePayloadRepositories
+	}
+
 	cust := cr.Customizations
 	if cust == nil {
+		// If no customizations requested, just return the repository snapshots resolved from the content template if one was specified
+		if res.CustomRepositories != nil && res.PayloadRepositories != nil {
+			return res, nil
+		}
 		return nil, nil
```
croissanne, line 840: I think this could be rewritten to make it a bit easier to read, how about:

```suggestion
	cust := cr.Customizations
	if cust == nil && cr.ImageRequests[0].ContentTemplate == nil {
		return nil, nil
	}

	res := &composer.Customizations{}
	if cr.ImageRequests[0].ContentTemplate != nil {
		templatePayloadRepositories, templateCustomRepositories, _, err := h.buildTemplateRepositories(ctx, *cr.ImageRequests[0].ContentTemplate)
		if err != nil {
			return nil, err
		}
		res.CustomRepositories = &templateCustomRepositories
		res.PayloadRepositories = &templatePayloadRepositories
	}
```

croissanne (commented):
Overall lgtm, could you squash the commits though? And add  some of the PR description into the commit message, to describe what's happening a little.

internal/v1/handler_compose_image.go
xbhouse: hmmm i see a null pointer dereference when running TestComposeCustomizations with these changes, maybe there's something wrong with the cases i added here? do [these](https://github.com/osbuild/image-builder-crc/pull/1489/files#diff-8157e75e8106573de99e0c216371cf52ffe66f95cb956487c54ceafe6eaba7a8R2824) look right?

internal/v1/handler_compose_image.go
```diff
@@ -467,6 +476,112 @@ func (h *Handlers) buildCustomRepositories(ctx echo.Context, custRepos []CustomR
...
+		// Separate Red Hat repos from custom
+		if len(rhRepoMap) > 0 {
+			rhRepoIDs = append(rhRepoIDs, id)
+		} else {
+			customRepoIDs = append(customRepoIDs, id)
+		}
+	}
+
+	ctx.Logger().Debugf("Found RH repositories from template: %v", rhRepoIDs)
+	ctx.Logger().Debugf("Found custom repositories from template: %v", customRepoIDs)
```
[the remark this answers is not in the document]
xbhouse: thanks! fixed :)

internal/v1/handler_compose_image.go
croissanne: ah, you're right, sorry misread this a bit. Updated the suggestion but willing to leave it as is. I think it's slightly easier to read if all the template logic is within the same code block.

croissanne (approved):
Thanks!

croissanne (approved):

PR #1494, opened 2025-02-24
build: ensure Go version in prep script

lzap:
There is something different on the CICD, I am unable to reproduce this failure locally:

https://github.com/osbuild/image-builder-crc/pull/1490

This is an attempt to fix that. Versions do vary, this puts them in sync.

### DIFF
--- a/tools/prepare-source.sh
+++ b/tools/prepare-source.sh
@@ -2,7 +2,7 @@
 set -eu
 
 GO_MAJOR_VER=1.22
-GO_VERSION=1.22.9
+GO_VERSION=1.22.9 # also update .github/workflows/tests.yml
 OAPI_VERSION=2.4.1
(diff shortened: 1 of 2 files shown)


tools/prepare-source.sh
```diff
@@ -13,8 +13,11 @@ if test "$LATEST" != "$GO_VERSION"; then
 fi
 
 set -x
+export GOTOOLCHAIN=go$GO_VERSION
+export GOSUMDB='sum.golang.org' # this is turned off for Go from Fedora / RHEL
+go version
```
lzap, line 18: This ensures the Go version in this script is executed with the exactly same version as intendent. The SUM thing is a Fedora custom Go configuration which brakes this locally - sumdb is turned off for Fedora.

ezr-ondrej (approved):
I'd love to read that version from a single var, but this is a great start :+1:

PR #1502, opened 2025-03-02
build: make sure tools are in go.mod

lzap:
I do not understand why PRs are failing on the prepare script: https://github.com/osbuild/image-builder-crc/actions/runs/13613898854/job/38054492007?pr=1501 (and many more this week).

The [line is already there](https://github.com/osbuild/image-builder-crc/blob/main/go.mod#L54) in main. Something is not right.

Just trying something in this PR.


lzap:
Yes seems to be passing. Still unsure why `git diff` shows that patch when the line was present, the added line ensures it is always there.

ezr-ondrej (approved):
Lets make sure

PR #1504, opened 2025-03-04
build: run 'go tidy' as the last prep step

lzap:
The last change did not help this is another idea - `go mod tidy` is not executed as the last step. This command goes through to project tree, finds all imports and cleans up `go.mod` and `go.sum` by adding missing deps or removing those which are unused. It was not executed as the last step, so some commands may modify these files.

This patch moves is to the very back to ensure consistency. I think this should help.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -54,7 +54,6 @@ require (
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mohae/deepcopy v0.0.0-20170929034955-c48cc78d4826 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
-	github.com/oapi-codegen/oapi-codegen/v2 v2.4.1 // indirect
 	github.com/perimeterx/marshmallow v1.1.5 // indirect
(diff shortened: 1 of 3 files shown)


lzap:
Although Konflux deps updates do work now thanks to the hack I made earlier, I think this is a cleaner approach.

lzap:
I think I should explain a little better when is going on and why this PR is important.

* The script checks for existence of `oapi-codegen` and `goimports`.
* It performs `go mod tidy`
* Then when one of the tools is missing, it will `go install` them.
* This operation adds a new dependency to the `go.mod` and `go.sum` but only if the tool was missing.
* These tools are always missing on CI and almost always are present on developer machines.
* This creates inconsistency: running this script locally creates a different `go.{mod,sum}` files than on CI.
* This fix runs `go mod tidy` at the very end to cleanup the unwanted dependencies added by `go install`.

I am going to rebase this one, I think we need this in. Only `image-builder-crc` suffers from this, other projects do not appear to have conditional `go install` stuff.

lzap:
I simplified the prepare script quite a bit, I do not think installing the binaries is necessary since both `go generate` and `go run .../goimports` can be done directly.

croissanne (approved):

ezr-ondrej:
Thank you @lzap for dealing with this and simplifying the script 🧡

PR #1508, opened 2025-03-12
v1: add more rhel 9 and 10 nightlies

croissanne:
Used for internal generation of CI images.


tkoscieln (commented):
LGTM. Could we do the same for RHEL 10.0 Nightly and 10.1 Nightly?

internal/v1/api.go
```diff
@@ -91,6 +91,8 @@ const (
 	Rhel93        Distributions = "rhel-93"
 	Rhel94        Distributions = "rhel-94"
 	Rhel95        Distributions = "rhel-95"
+	Rhel96Nightly Distributions = "rhel-9.6-nightly"
+	Rhel97Nightly Distributions = "rhel-9.7-nightly"
```
tkoscieln, line 95: Shouldn't this be changed for RHEL 10 nightly as well?

internal/v1/api.yaml
```diff
@@ -1116,6 +1116,8 @@ components:
         - rhel-8.10
         - rhel-9
         - rhel-9-nightly
+        - rhel-9.6-nightly
+        - rhel-9.7-nightly
```
tkoscieln, line 1120: Shouldn't this be changed for RHEL 10 nightly as well?

tkoscieln (changes requested):

internal/v1/api.yaml
croissanne: indeed

tkoscieln (approved):

tkoscieln:
Ty :confetti_ball:

PR #1513, opened 2025-03-18
refactor: replace logrus with slog

lzap:
This replaces all dummy `logrus` statements with `slog` statements. No changes are done to all logging statements other than log messages start with lower case. I like this one more because Go errors are supposed to start like that so people are not confused and it generally does not look messy like:

```go
slog.Error("Some error")
return fmt.Errorf("some error")
```

Additionally, there was a test that was testing for a logging statement, it was not thread-safe and was not needed - generally it is unnecessary to create tests for logging statements, unless your goal is writing a logging library which is not the case.


internal/oauth2/lazy_token.go
```diff
@@ -52,9 +51,9 @@ func (lt *LazyToken) acquireNewToken(ctx context.Context, forceRefresh bool) (st
 		lt.AccessToken = tokenRes.AccessToken
 		lt.Expiration = time.Now().Add(time.Duration(tokenRes.ExpiresIn) * time.Second)
 
-		logrus.WithContext(ctx).Infof("Acquired new token with expiration at: %s", lt.Expiration)
+		slog.DebugContext(ctx, "acquired new token", "expiration", lt.Expiration)
 	} else {
-		logrus.WithContext(ctx).Infof("AccessToken reused, which expires at %s", lt.Expiration)
+		slog.DebugContext(ctx, "reusing token", "expiration", lt.Expiration)
 	}
```
lzap, line 57: Changing from info to debug, I think it is pretty noisy one this one.

internal/oauth2/lazy_token_test.go
```diff
@@ -61,23 +46,11 @@ func TestLazyToken(t *testing.T) {
...
 	token, err = lazyToken.Token(ctx)
 	require.NoError(t, err)
 	require.Equal(t, "mock-token-1", token)
 
-	field = collector.Last()["msg"]
-	assert.Equal(t, 1, len(collector.All()))
-	assert.Contains(t, field, "AccessToken reused")
-	assert.NotContains(t, field, "mock-token-1")
-	collector.Reset()
-
```
lzap, line 80: Dropped all log-related testing, the test itself is good it checks via counter if tokens are reused or not.

internal/v1/server_test.go
```diff
@@ -121,9 +123,6 @@ type testServer struct {
 func startServer(t *testing.T, tscc *testServerClientsConf, conf *v1.ServerConfig) *testServer {
 	ctx := context.Background()
 
-	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelDebug}))
-	slog.SetDefault(logger)
-
```
lzap, line 126: Moved these two lines up to the `TestMain` which is called earlier.

cmd/image-builder-maintenance/db.go
```diff
@@ -114,23 +113,32 @@ func (d *maintenanceDB) LogVacuumStats(ctx context.Context) (int64, error) {
...
-				"vacuum_count":      vc,
-				"autovacuum_count":  avc,
-				"last_vacuum":       lvc,
-				"last_autovacuum":   lavc,
-				"analyze_count":     ac,
-				"autoanalyze_count": aac,
-				"last_analyze":      lan,
-				"last_autoanalyze":  laan,
-			}).Info("Vacuum and analyze stats for table")
+			attrs := []any{
```
ezr-ondrej, line 116: Just **out of interest**: Do we need to build an array of attributes or is it possible to inline it to With and this is just personal choice?

internal/clients/composer/client.go
```diff
@@ -33,7 +33,7 @@ var contentHeaders = map[string]string{"Content-Type": "application/json"}
 
 func NewClient(conf ComposerClientConfig) (*ComposerClient, error) {
 	if conf.URL == "" {
-		logrus.Warn("Composer URL not set, client will fail")
+		slog.Warn("composer URL not set, client will fail")
```
ezr-ondrej, line 36: Isn't log message supposed to be Upper cased?

internal/clients/recommendations/client.go
```diff
@@ -35,7 +35,7 @@ var contentHeaders = map[string]string{"Content-Type": "application/json"}
 
 func NewClient(conf RecommendationsClientConfig) (*RecommendationsClient, error) {
 	if conf.URL == "" {
-		logrus.Warn("Recommendation URL not set, client will fail")
+		slog.Warn("recommendation URL not set, client will fail")
```
ezr-ondrej, line 38: Again, log message upper/lower

ezr-ondrej (approved):
Just curious why you're lower casing the log messages, apart of that, looks great to me 👍 
Thanks for working on this! 🧡

cmd/image-builder-maintenance/db.go
lzap: It cannot be done, it is limitation of the Go compiler one cannot mix variadic arguments with normal arguments. I In other words, this is not possible, at least to my knowledge:

```
fmt.Println("This", "is", "okay")
	
slice := []string{"is", "not"}
fmt.Println("But", "this", slice...) // compiler error
```

Playground: https://go.dev/play/p/ewHtdzD6pm8

internal/clients/composer/client.go
lzap: I propose to make all logging lower case, but good point: I will starte a thread on Slack before we merge this.

croissanne (approved):

PR #1514, opened 2025-03-19
v1: redact satellite token and certs (HMS-4590)

avitova:
This helps to prevent users from sharing tokens and certificates.

My question is...should we try to redact this for get as well? It might make sense, but just the idea of having to regenerate the satellite command and upload certificates every time you edit the blueprint sounds a bit annoying?

I have the rest of the frontend-side work ready, so I thought that I could prepare the backend-side while we are waiting for iqe modifications.

### DIFF
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -37,6 +37,8 @@ type BlueprintBody struct {
 	ImageRequests  []ImageRequest `json:"image_requests"`
 }
 
+type BlueprintBodyOption func(*BlueprintBody)
+
 func (u *User) CryptPassword() error {
 	// Prevent empty and already hashed password  from being hashed
 	if u.Password == nil || len(*u.Password) == 0 || crypt.PasswordIsCrypted(*u.Password) {
@@ -79,6 +81,10 @@ func (bb *BlueprintBody) RedactPasswords() {
 			(*bb.Customizations.Users)[i].RedactPassword()
 		}
 	}
+}
+
+func (bb *BlueprintBody) RedactCertificates() {
+	bb.Customizations.Cacerts = nil
 }
 
 // Merges Password or SshKey from other User struct to this User struct if it is not set
@@ -147,22 +153,45 @@ func BlueprintFromAPI(cbr CreateBlueprintRequest) (BlueprintBody, error) {
 }
 
 // Util function used to create Blueprint sctruct from DB entry (READ)
-func BlueprintFromEntry(be *db.BlueprintEntry) (BlueprintBody, error) {
+func BlueprintFromEntry(be *db.BlueprintEntry, options ...BlueprintBodyOption) (BlueprintBody, error) {
 	var result BlueprintBody
 	err := json.Unmarshal(be.Body, &result)
 	if err != nil {
 		return BlueprintBody{}, err
 	}
+
+	for _, option := range options {
+		option(&result)
+	}
+
 	return result, nil
 }
 
-func BlueprintFromEntryWithRedactedPasswords(be *db.BlueprintEntry) (BlueprintBody, error) {
-	result, err := BlueprintFromEntry(be)
-	if err != nil {
-		return BlueprintBody{}, err
+func WithRedactedPasswords() BlueprintBodyOption {
+	return func(bp *BlueprintBody) {
+		bp.RedactPasswords()
 	}
-	result.RedactPasswords()
-	return result, nil
+}
+
+func WithRedactedCertificates() BlueprintBodyOption {
+	return func(bp *BlueprintBody) {
+		bp.RedactCertificates()
+	}
+}
+
+func WithRedactedFiles(paths []string) BlueprintBodyOption {
+	return func(bp *BlueprintBody) {
+		if bp.Customizations.Files != nil {
+			files := slices.DeleteFunc(*bp.Customizations.Files, func(file File) bool {
+				return slices.Contains(paths, file.Path)
+			})
+			if len(files) == 0 {
+				bp.Customizations.Files = nil
+			} else {
+				bp.Customizations.Files = &files
+			}
+		}
+	}
 }
 
 func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
@@ -270,7 +299,10 @@ func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID, params 
 		return err
 	}
 
-	blueprint, err := BlueprintFromEntryWithRedactedPasswords(blueprintEntry)
+	blueprint, err := BlueprintFromEntry(
+		blueprintEntry,
+		WithRedactedPasswords(),
+	)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


internal/v1/handler_blueprints.go
```diff
@@ -162,6 +166,32 @@ func BlueprintFromEntryWithRedactedPasswords(be *db.BlueprintEntry) (BlueprintBo
...
+func BlueprintFromEntryRedactedForExport(be *db.BlueprintEntry) (BlueprintBody, error) {
+	blueprint, err := BlueprintFromEntryWithRedactedPasswords(be)
+	if err != nil {
+		return BlueprintBody{}, err
+	}
+
+	blueprint.RedactCertificates()
+	var files []File
+	if blueprint.Customizations.Files != nil {
+		for _, file := range *blueprint.Customizations.Files {
```
mvo5, line 178: (super nitpick) we could use `slices.DeleteFunc` here (and `slices.contains`) [0] which that it would be slightly more compact. But the exiting version is fine of course and this one might be a bit too obscure (I do like the slices package though :)

```go
...
	blueprint.RedactCertificates()
	redactPaths := []string{
		"/etc/systemd/system/register-satellite.service",
		"/usr/local/sbin/register-satellite",
	}
	if blueprint.Customizations.Files != nil {
		newFiles := slices.DeleteFunc(*blueprint.Customizations.Files, func(file File) bool {
			return slices.Contains(redactPaths, file.Path)
		})
		blueprint.Customizations.Files = &newFiles
	}

	return blueprint, nil
```

[0] I think at least, it was added in go 1.21 but maybe DeleteFunc was added later, I did not double check.

mvo5 (commented):
Thanks, this looks nice! I add some tiny suggestions inline for your consideration but all optional this is nice already.

internal/v1/handler_blueprints.go
avitova: Interesting point, though this still requires that we do:
`	if blueprint.Customizations.Files == nil || len(files) == 0 {
		blueprint.Customizations.Files = nil
}
`
If we do not want to export "files": [] in the JSON.

I usually prefer writing these kinds of things myself, rather than using packages that are a bit of a black box, but if you or anyone else reviewing will have a strong preference, I do not mind changing this.

internal/v1/handler_blueprints.go
```diff
@@ -299,7 +329,7 @@ func (h *Handlers) ExportBlueprint(ctx echo.Context, id openapi_types.UUID) erro
 		return err
 	}
 
-	blueprint, err := BlueprintFromEntryWithRedactedPasswords(blueprintEntry)
+	blueprint, err := BlueprintFromEntryRedactedForExport(blueprintEntry)
```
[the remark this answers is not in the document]
lzap: Since this is now the second time we have this BlueprintWithSomething, I think it is time to think about generalizing it. There are many options, one that immediately comes to my mind is to modify the `BlueprintFromEntry` function with a variadic arguments so it can be used fiexibly:

```go
blueprint, err := BlueprintFromEntry(blueprintEntry, WithRedactedPassword, WithRedactedXXX)
```

This is somewhat similar to the [with options pattern for Go](https://medium.com/@omid.jn/the-option-pattern-in-go-9edf55e88320).

internal/v1/handler_blueprints_test.go
```diff
@@ -730,6 +730,65 @@ func TestHandlers_BlueprintFromEntryWithRedactedPasswords(t *testing.T) {
...
+					"path": "/usr/local/sbin/register-satellite"
+				  }
+				],
+				"distribution": "centos-9"}}`)
+		be := &db.BlueprintEntry{
+			Body: body,
+		}
+		result, err := v1.BlueprintFromEntryRedactedForExport(be)
+		require.NoError(t, err)
+		require.Nil(t, result.Customizations.Cacerts)
```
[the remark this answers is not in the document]
lzap: Just to make sure you guys are on the same page, `len()` in Go is defined in a way that it returns zero for `nil` values: https://go.dev/play/p/DKYx9iCqP17

I am assuming `require.Len` does the same in which case I agree with @mvo5 this is nicer.

lzap (approved):
I agree with @mvo5 on everything plus I think I do not like `BlueprintWithRedacted` functions and I suggest to generalize it somehow.

internal/v1/handler_blueprints.go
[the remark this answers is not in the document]
avitova: I love this idea, thanks, I will implement this, Lukas. :)

avitova (commented):
Working on the update:)

lzap:
Let us know for a re-review. A pointer to slice is not different from any other pointer - it needs to be dereferenced when passed to `len` but I am sure you figured that out already :-)

avitova:
Thanks for explaining what was meant by this all, I got confused:)

However, now we will export `"cacerts": {}`, but I believe that should not be such a problem. :) Fixed

internal/v1/handler_blueprints.go
```diff
@@ -149,22 +155,45 @@ func BlueprintFromAPI(cbr CreateBlueprintRequest) (BlueprintBody, error) {
...
+		if bp.Customizations.Files != nil {
+			files := slices.DeleteFunc(*bp.Customizations.Files, func(file File) bool {
+				return slices.Contains(paths, file.Path)
+			})
+			if len(files) == 0 {
+				bp.Customizations.Files = nil
+			} else {
+				bp.Customizations.Files = &files
+			}
+		}
```
lzap, line 195: I would probably drop the intermediate value `files` deleting directly so the else branch could be dropped, but this is fine.

lzap (approved):

lzap (approved):

avitova:
I do not think that these errors are relevant. :shrug:

internal/v1/handler_blueprints_test.go
```diff
@@ -730,14 +736,94 @@ func TestHandlers_BlueprintFromEntryWithRedactedPasswords(t *testing.T) {
...
+				  }
+				],
+				"distribution": "centos-9"}}`)
+		be := &db.BlueprintEntry{
+			Body: body,
+		}
+		result, err := v1.BlueprintFromEntry(
+			be,
+			v1.WithRedactedPasswords(),
+			v1.WithRedactedCertificates(),
```
mvo5, line 813: (nitpick) It would be nice to add tiny tests for WithRedactedCertificates and WithRedactedFiles to be tested in isolation with a small test case (I personally would add them but this is not my codebase so … up-to-others :)

internal/v1/handler_blueprints.go
mvo5: Fwiw, I agree (but also think its not very important)

mvo5 (approved):
Thanks, I like this! One tiny suggestion but feel free to ignore.

internal/v1/handler_blueprints_test.go
avitova: Okay, good. I think I can do this in a followup if you do not mind. :)

avitova:
So are we all good here?:)

lzap:
Unless @croissanne has any objections, we are good to go, thank you.

croissanne (approved):

PR #1515, opened 2025-03-20
v1: don't redact password when starting a compose

croissanne:
This arose from a misunderstanding in osbuild/image-builder-crc#1283,
the compose request that is sent to osbuild-composer shouldn't be
considered as a user-facing endpoint, and omitting the password before
sending it off will result in no password being set at all.


lzap:
Yeah I have missed this during review.

ezr-ondrej (approved):
Thanks! 🧡

PR #1528, opened 2025-03-28
blueprints: support linting & fixing blueprints (HMS-5832)

croissanne:
- **clients/compliance: support fetching blueprint customizations**
- **v1: add linting to blueprints**
- **common: add DerefOrDefault helper function**
- **v1/mocks: expand compliance mock**
- **v1/test: add test for blueprint linting**
- v1: support fixing lint errors in blueprint

JIRA: [HMS-5832](https://issues.redhat.com/browse/HMS-5832)


kingsleyzissou (commented):
Looks good to me. I have one question though, do we have an endpoint to get the tailored blueprint and return that to the frontend during blueprint customization?
Or are we already doing that?

My thoughts are that we probably wouldn't want to create a blueprint without the customizations options and then immediately have lint conflicts

croissanne:
> Looks good to me. I have one question though, do we have an endpoint to get the tailored blueprint and return that to the frontend during blueprint customization? Or are we already doing that?
> 
> My thoughts are that we probably wouldn't want to create a blueprint without the customizations options and then immediately have lint conflicts

Oh let's maybe do that in a separate PR? I think currently it's requesting the customisations from `/oscap/{distribution}/{profile}/customizations`, could add something like `  /oscap/{policy}/customizations`.

kingsleyzissou (dismissed):

internal/common/pointers.go
```diff
@@ -10,3 +10,13 @@ func FromPtr[T any](ref *T) (value T) {
...
+
+// DerefOrDefault returns the dereferenced value of the given pointer or the
+// default value for the type if unset.
+func DerefOrDefault[T any](p *T) T {
+	var v T
+	if p != nil {
+		v = *p
+	}
+	return v
+}
```
lzap, line 22: This method does literally what the one defined above does, named `FromPtr`:

https://go.dev/play/p/1Vu4HuSa_T9

internal/v1/api.yaml
```diff
@@ -207,6 +207,15 @@ paths:
...
                 $ref: '#/components/schemas/HTTPErrorList'
+    patch:
+      summary: Apply linter fixes to blueprint
+      description: |
+        Apply fixes which should fix any lint errors in the blueprint.
+      responses:
+        200:
+          description: successful update
+        404:
+          description: blueprint was not found
```
lzap, line 218: Not fan of using HTTP PATCH on the blueprint resource for this, I encourage you to introduce a new sub-resource for this.

internal/v1/handler_blueprints.go
```diff
@@ -653,3 +793,64 @@ func (h *Handlers) DeleteBlueprint(ctx echo.Context, blueprintId openapi_types.U
...
 	return ctx.NoContent(http.StatusNoContent)
 }
+
+func (h *Handlers) PatchBlueprintsId(ctx echo.Context, id openapi_types.UUID) error {
+	userID, err := h.server.getIdentity(ctx)
+	if err != nil {
+		return err
+	}
+
+	ctx.Logger().Infof("Fetching blueprint %s", id)
```
lzap, line 803:
```suggestion
	slog.InfoContext(ctx.Request().Context(), "fetching blueprint", "id", id)
```

Please do not capitalize logs and prefer structured logging. I think echo logger does support structured fields but AFAIR it is not very easy to use, I would just prefer `slog`. Use method `Infoj` if you want to use that.

internal/v1/handler_blueprints.go
```diff
@@ -653,3 +793,64 @@ func (h *Handlers) DeleteBlueprint(ctx echo.Context, blueprintId openapi_types.U
...
+		Distribution:   blueprint.Distribution,
+		ImageRequests:  blueprint.ImageRequests,
+		Customizations: blueprint.Customizations,
+	}
+
+	body, err := json.Marshal(blueprintRequest)
+	if err != nil {
+		return err
+	}
+	desc := common.DerefOrDefault(blueprintRequest.Description)
```
lzap, line 844:
```suggestion
	desc := common.FromPtr(blueprintRequest.Description)
```

internal/v1/handler_blueprints.go
```diff
@@ -653,3 +793,64 @@ func (h *Handlers) DeleteBlueprint(ctx echo.Context, blueprintId openapi_types.U
...
+
+	err = h.server.db.UpdateBlueprint(ctx.Request().Context(), uuid.New(), blueprintEntry.Id, userID.OrgID(), blueprintRequest.Name, desc, body)
+	if err != nil {
+		ctx.Logger().Errorf("Error updating blueprint in db: %v", err)
+		if errors.Is(err, db.ErrBlueprintNotFound) {
+			return echo.NewHTTPError(http.StatusNotFound, err)
+		}
+		return err
+	}
+	ctx.Logger().Infof("Updated blueprint %s", blueprintEntry.Id)
```
lzap, line 854: Ditto - logging (here and on line 848.

lzap (approved):

lzap (changes requested):
Did not mean to approve, the utility function should definitely be dropped as it is just exact copy of the one above.

internal/common/pointers.go
croissanne: woops, right

internal/v1/handler_blueprints.go
croissanne: I removed them, they don't really contribute much anyway.

internal/common/pointers.go
lzap: I think it was me who added it. On your defence, it lacks godoc completely it should have it perhaps.

internal/v1/handler_blueprints.go
lzap: For the record, structured logging via echo logger looks ugly as heck:

```
echo.Infoj("use infoj method instead", map[string]any{"msg": "test"})
```

That is the reason why I suggest using `slog` which is much cleaner.

internal/v1/handler_blueprints.go
```diff
@@ -272,18 +272,158 @@ func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID, params
...
+				})
+			}
+			if fixup {
+				blueprint.Customizations.Kernel = common.ToPtr(common.FromPtr(blueprint.Customizations.Kernel))
+				blueprint.Customizations.Kernel.Append = common.ToPtr(fmt.Sprintf("%s %s", common.FromPtr(blueprint.Customizations.Kernel.Append), kcmd))
+			}
+		}
+	}
+	return lintErrors, nil
+}
```
lzap, line 425: I see a pattern that repeats 4-5 times if I am not mistaken, could be maybe generalized cos I think you are onto something in here and more fixups could be added later. Alternatively, the function could be split into separate subfunctions. Maybe having this in a handler is also not good idea, perhaps acceptable for now and separate package later if it grows.

lzap (approved):
Looks great, couple of possible overlooks, feel free to fix them or not.

internal/v1/handler_blueprints.go
croissanne: Hm I wonder how you'd generalise this because each separate customisation needs a different treatment. Like other than using reflection and making some guesses.

Agreed with the handler point, I think we should detangle the handlers in general a bit in future. Not sure if this is the right PR for it though.

lzap: I just saw a pattern and did not look into it deeply. Definitely did not mean to use reflection, no :-) This is fine. Thanks.

lzap (approved):

kingsleyzissou (approved):

PR #1540, opened 2025-04-02
refactor: make use of generated clients for composer

lzap:
I have always wondered why client code is not generated, this fixes it. It should be pretty stragithforward change for all other OAS3 clients (not onese which are OpenAPI 3.1). Review with care, things can go wild here.

### DIFF
--- a/internal/oauth2/doer.go
+++ b/internal/oauth2/doer.go
@@ -1,0 +1,61 @@
+package oauth2
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/osbuild/logging/pkg/strc"
+)
+
+func do(client *http.Client, tokener Tokener, req *http.Request) (*http.Response, error) {
+	if client == nil {
+		panic("client must be set before calling Do()")
+	}
+
+	var bodyBytes []byte
+	var err error
+	if req.Body != nil {
+		bodyBytes, err = io.ReadAll(req.Body)
+		if err != nil {
+			return nil, err
+		}
+
+		err = req.Body.Close()
+		if err != nil {
+			return nil, err
+		}
+
+		req.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+	}
+
+	token, err := tokener.Token(req.Context())
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+
+	resp, err := strc.NewTracingDoer(client).Do(req)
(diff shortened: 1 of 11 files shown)


internal/oauth2/doer.go
```diff
@@ -0,0 +1,56 @@
...
+			req.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+		}
+		resp, err = strc.NewTracingDoer(client).Do(req)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return resp, err
+}
```
lzap, line 56: Actual token handling is now part of the Tokener itself. It implements `Doer` interface to do that.

internal/v1/handler.go
```diff
@@ -57,7 +57,7 @@ func (h *Handlers) GetVersion(ctx echo.Context) error {
 }
 
 func (h *Handlers) GetReadiness(ctx echo.Context) error {
-	resp, err := h.server.cClient.OpenAPI(ctx.Request().Context())
+	resp, err := h.server.cClient.GetOpenapi(ctx.Request().Context())
```
lzap, line 60: And the intermediate client is gone, all generated methods can be used directly.

lzap:
Resolved all conflicts, rebased. Since this is the first of a series of all client refactoring, I would appreciate a blessing from @croissanne cheers :-D

croissanne:
I'm not sure if I'm a fan of the generated client code. I think it adds a lot of code and might make debugging harder. I mean we're literally replacing 75 lines with 2300 lines, and not necessarily more readable code.

What are the benefits of the generated client?

lzap:
I was able to bring it down to `+1,160 −278` by selecting only the operations we need at the moment. The generated code has both parsed and unparsed variants and this is not configurable, this is major reason why there is more code.

> What are the benefits of the generated client?

The OpenAPI was literally created for generating server and client stubs, the biggest benefit is not having to maintain boilerplate code for example body parsing. Specifically, error handling is all generated code as it is defined in OpenAPI. This is very often a cause of bugs, at the cost of little bit more code, handling of error HTTP codes is much easier.

No hard feelings here, we already use ton of generated code for server and models.

mvo5:
A bit of a meta-comment: if that would be split into multiple commits it might be a bit easier for me (but maybe its just me). As a stawman: the auto-generation in its own commit, the auto-generated code in its own and then the support and then finally the commit that uses the new code. But again, I feel bad for asking for this as it seems like I can puzzle it together myself if I just put a little bit more effort into it :)

PR #1543, opened 2025-04-03
floorist: Add locale and timezone to query

ochosi:
In order to track the locale and timezone features in our metrics report, we need to export the usage of these customizations via Floorist.


lucasgarfield (approved):
lgtm

PR #1553, opened 2025-04-08
templates/image-builder: fix floorist query

croissanne:
Parquet doesn't deal well with empty objects:
```
Cannot write struct type 'timezone' with no child field to
Parquet. Consider adding a dummy child field.
```

Work around it by returning 'null' if all the fields in the objects are marked as optional.


ochosi (dismissed):

lzap (dismissed):
Nice, but why the capitalization of a lower-case SQL query. :-)

croissanne:
> Nice, but why the capitalization of a lower-case SQL query. :-)

oh hey, good point, fixing that. Force of habit.

ochosi (approved):

PR #1554, opened 2025-04-08
v1: fix adding RH repos from template to compose request

xbhouse:
Tested the recent [backend changes for content templates](https://github.com/osbuild/image-builder-crc/pull/1489) in the associated frontend PR and am hitting this [error](https://github.com/osbuild/image-builder-crc/blob/main/internal/v1/handler_compose_image.go#L161) when composing an image using a template. 

I'm not sure if the request is failing due to the missing base RH repositories, but they need to be added regardless.

### DIFF
--- a/internal/v1/handler_compose_image.go
+++ b/internal/v1/handler_compose_image.go
@@ -483,8 +483,6 @@ func (h *Handlers) buildCustomRepositories(ctx echo.Context, custRepos []CustomR
 func (h *Handlers) buildTemplateRepositories(ctx echo.Context, templateID string) ([]composer.Repository, []composer.CustomRepository, []composer.Repository, error) {
 	var customRepoIDs []string
 	var rhRepoIDs []string
-	rhRepoMap := content_sources.RepositoryByID{}
-	customRepoMap := content_sources.RepositoryByID{}
 
 	template, err := h.server.csClient.GetTemplateByID(ctx.Request().Context(), templateID)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


croissanne (changes requested):
lgtm but could you maybe rewrite the commit as `v1: fix adding RH repos from snapshot template to compose request`? 

Generally we aim to have `topic: title`, where `topic` is really more the section of code that's being changed, in this case the v1 api.

internal/v1/handler_compose_image.go
```diff
@@ -497,7 +496,7 @@ func (h *Handlers) buildTemplateRepositories(ctx echo.Context, templateID string
 
 	for _, id := range *template.RepositoryUuids {
 		// Check if repo is Red Hat
-		rhRepoMap, err = h.server.csClient.GetRepositories(ctx.Request().Context(), []string{}, []string{id}, false)
+		rhRepoMap, err := h.server.csClient.GetRepositories(ctx.Request().Context(), []string{}, []string{id}, false)
```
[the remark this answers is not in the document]
xbhouse: good point! reworked it a bit, let me know if that's better

croissanne (approved):

internal/v1/handler_compose_image.go
```diff
@@ -495,17 +493,23 @@ func (h *Handlers) buildTemplateRepositories(ctx echo.Context, templateID string
...
-	for _, id := range *template.RepositoryUuids {
-		// Check if repo is Red Hat
-		rhRepoMap, err = h.server.csClient.GetRepositories(ctx.Request().Context(), []string{}, []string{id}, false)
-		if err != nil {
-			return nil, nil, nil, fmt.Errorf("Unable to retrieve Red Hat repositories: %v", err)
+	rhRepoMap, err := h.server.csClient.GetRepositories(ctx.Request().Context(), []string{}, *template.RepositoryUuids, false)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("Unable to retrieve Red Hat repositories: %v", err)
+	}
+	if len(rhRepoMap) > 0 {
```
croissanne, line 500: Technically a length check isn't necessary here, `for repoID := range rhRepoMap ...` will already do what you want and just not iterate

internal/v1/handler_compose_image.go
```diff
@@ -495,17 +493,23 @@ func (h *Handlers) buildTemplateRepositories(ctx echo.Context, templateID string
...
-			rhRepoIDs = append(rhRepoIDs, id)
-		} else {
-			customRepoIDs = append(customRepoIDs, id)
+	}
+
+	customRepoMap, err := h.server.csClient.GetRepositories(ctx.Request().Context(), []string{}, *template.RepositoryUuids, true)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("Unable to retrieve custom repositories: %v", err)
+	}
+	if len(customRepoMap) > 0 {
```
croissanne, line 510: Same here, I think the length check is superfluous.

croissanne (changes requested):
Noticed one more thing, sorry :(

internal/v1/handler_compose_image.go
xbhouse: oh yes, fixed :)

croissanne (approved):
thanks!

PR #1557, opened 2025-04-10
v1/test: discard all log output

croissanne:
The test logs are actually quite hard to read in case of failures, they obscure useful information.

--- 

Similar to https://github.com/osbuild/osbuild-composer/pull/4614, might be controversial so depending on what other people think. But generally the useful information is in that which go test outputs, not the logs.

### DIFF
--- a/internal/v1/server_test.go
+++ b/internal/v1/server_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"log/slog"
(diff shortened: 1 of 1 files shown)


lzap (approved):
Oh yeah, absolutely.

The logging library also have Collector handler which you can use too and in case of failed test print the logs.

One observation tho - if you run a single file or package, this will not be executed and therefore default slog will be used, which is a debug level text on stderr I think. There is not much one can do against it AFAIK.

PR #1563, opened 2025-04-14
renovate.json: pin go version

croissanne:
Currently the bot is trying to update the go version of the whole project (see #1562), but the max go version at any given time should be the one shipped in the ubi9 go-toolset container.


bcl (approved):
Looks ok to me

PR #1564, opened 2025-04-14
v1: use internally accessible baseurl for RH repo snapshot from template

xbhouse:
We need to use the internal pulp URL for RH repo snapshots coming from a content template so composer can access it, otherwise we'll see errors reading the repository on image build


croissanne (approved):
Thanks!

PR #1565, opened 2025-04-14
tools/prepare-source: align with images' version

croissanne:
Removes go version check as it's misleading (we need to align with go-toolset ubi9). And don't use the hosts go version to use tooling like fmt, mod etc.

### DIFF
--- a/tools/prepare-source.sh
+++ b/tools/prepare-source.sh
@@ -1,18 +1,21 @@
 #!/bin/sh
 set -eu
 
-GO_MAJOR_VER=1.22
-GO_VERSION=1.22.9 # also update .github/workflows/tests.yml
+# Go version must be consistent with image-builder which uses UBI
+# container that is typically few months behind
+GO_VERSION=1.22.9
+GO_BINARY=$(go env GOPATH)/bin/go$GO_VERSION
 
-# Pin Go and toolchain versions at a reasonable version
-go get go@$GO_VERSION toolchain@$GO_VERSION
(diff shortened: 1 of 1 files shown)


lzap:
_This branch is out-of-date with the base branch_

croissanne:
> I am okay with this but the new Go toolchain feature in 1.22 does exactly this.

Wait, so the 2nd commit here is a no-op?

croissanne:
It seems to me we might still need to choose the toolchain version explicitly in the script after downloading it? https://go.dev/doc/toolchain

croissanne:
ah indeed, then I might jsut drop the 2nd commit.

lzap:
> It seems to me we might still need to choose the toolchain version explicitly in the script after downloading it? https://go.dev/doc/toolchain

I am not entirely sure what you are trying to do here, the current version should already pin to Go 1.22 and the toolbox is set to 1.22.9.

Removing that `curl` is fine, it did not work well at all since we are lagging behind the upstream version so it was rather confusing than useful.

If you just want to use `goX.Y` then you can simply just set `GOTOOLCHAIN` variable and that should do it. I haven't tried this myself but this is what the linked doc says:

_When GOTOOLCHAIN is set to <name> (for example, GOTOOLCHAIN=go1.21.0), the go command always runs that specific Go toolchain. If a binary with that name is found in the system PATH, the go command uses it. Otherwise the go command uses a Go toolchain it downloads and verifies._

PR #1568, opened 2025-04-15
v1: set rhsm to false for repo snapshots from template

xbhouse:
Sets RHSM to false for repository snapshots coming from a template. When set to true, the RHSM certs would be used instead of the service account certs which may have resulted in repository read errors when building an image with a template.


lzap:
Understood, there is a pending review requested from @croissanne let me ask her.

PR #1569, opened 2025-04-16
v1: upload request default values

lzap:
I noticed that the upload request must be always present and that API user must figure out that the cryptic `aws.s3` is what they want when building non-cloud images like `guest-image`.

This patch makes `x86_64` a default value upload request architecture and `aws.s3` for UR type. I am trying to set a standard for how to set default values by creating methods called `BuildDefaults` in a separate file to do that.

Whether to include the architecture is an open question, I just wanted to have two examples when working on this. Also the default value is somewhat weird, we might be able to use the default value of the image itself (and convert it to the proper CR architecture constant).

This is all for discussion, I do not feel strongly about all of this but I would love to start taking advantage of the OpenAPI default value feature so the API is more convenient to work with.

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -1215,18 +1215,22 @@ components:
       required:
         - architecture
         - image_type
-        - upload_request
       properties:
         architecture:
           type: string
           enum:
             - x86_64
             - aarch64
+          default: x86_64
           description: |
             CPU architecture of the image, x86_64 and aarch64 are currently supported.
         image_type:
           $ref: '#/components/schemas/ImageTypes'
         upload_request:
+          description: |
+            Upload request is optional. If not specified, the image will be
+            uploaded to the image builder S3 bucket and kept for several days for download.
+          nullable: true
           $ref: '#/components/schemas/UploadRequest'
         ostree:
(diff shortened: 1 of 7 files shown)


internal/v1/api.yaml
```diff
@@ -1215,18 +1215,22 @@ components:
       required:
         - architecture
         - image_type
-        - upload_request
```
croissanne, line 1218: Hm, why remove this as required? Image requests are useless without an upload_request.

internal/v1/defaults.go
```diff
@@ -0,0 +1,35 @@
...
+		}
+
+		if (*obj)[i].UploadRequest == nil {
+			(*obj)[i].UploadRequest = &UploadRequest{
+				Type: UploadTypesAwsS3,
+				Options: UploadRequest_Options{
+					union: json.RawMessage(`{"type": "aws.s3", "options": {}}`),
+				},
+			}
+		}
```
croissanne, line 23: I'm not sure if default values make much sense here. If you want to add default values, they'd need to be set per image type.

lzap: That is a good point, but we do have the information in the request, so this could be actually inferred.

PR #1570, opened 2025-04-16
v1: allow users without pass/key (HMS-5993)

lzap:
We want to allow creation of system users with `/sbin/nologin` shell.

Not sure how to link jiras so trying both what I have seen so far:

https://issues.redhat.com/browse/HMS-4181

/jira-epic HMS-4181

JIRA: [HMS-5993](https://issues.redhat.com/browse/HMS-5993)

### DIFF
--- a/internal/v1/handler_blueprints_test.go
+++ b/internal/v1/handler_blueprints_test.go
@@ -108,15 +108,6 @@ func TestHandlers_CreateBlueprint(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, "Invalid blueprint name", jsonResp.Errors[0].Title)
 
-	// Test customization users, user without password and key is invalid
-	body["name"] = "Blueprint with invalid user"
-	body["customizations"] = map[string]interface{}{"users": []map[string]interface{}{{"name": "test"}}}
-	statusCode, resp = tutils.PostResponseBody(t, srvURL+"/api/image-builder/v1/blueprints", body)
-	require.Equal(t, http.StatusUnprocessableEntity, statusCode)
-	err = json.Unmarshal([]byte(resp), &jsonResp)
-	require.NoError(t, err)
-	require.Equal(t, "Invalid user", jsonResp.Errors[0].Title)
-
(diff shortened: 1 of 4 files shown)


internal/v1/handler_blueprints_test.go
```diff
@@ -108,15 +108,6 @@ func TestHandlers_CreateBlueprint(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, "Invalid blueprint name", jsonResp.Errors[0].Title)
 
-	// Test customization users, user without password and key is invalid
```
avitova, line 111: Well what if we tested that the `user without password and key is **valid**` instead?

avitova (commented):
I do not have a problem with this, and I did a review on the frontend-sidel so I kinda know where this is going.
I think we do not need to throw away all the tests - we can just make them a positive scenario tests, what do you think?

internal/v1/handler_blueprints_test.go
lzap: Good point, I am putting all the tests back but this one I am thinking removing because it sort of breaks the flow of this test - it tries to create a record several times and then the final one is success. I think authentication is well covered by the other dozen of tests if you don't mind :-)

Rebased.

croissanne (approved):

lzap:
@tkoscieln hey could you help me with updating the following tests which will break when this is merged?

```
[tests.test_blueprints.test_blueprint_create[ViaREST]](https://ci.int.devshift.net/job/osbuild-image-builder-crc-pr-check/276/testReport/junit/tests/test_blueprints/test_blueprint_create_ViaREST_/)
[tests.test_blueprints.test_blueprint_update[ViaREST]](https://ci.int.devshift.net/job/osbuild-image-builder-crc-pr-check/276/testReport/junit/tests/test_blueprints/test_blueprint_update_ViaREST_/)
```

The gist of the change is - we now allow creating/editing users without any password nor ssh-key. Those tests probably checks that, we can just either remove those checks or flip it to positive cases, thanks!

PR #1577, opened 2025-04-23
v1: custom repos should be empty if using a template

xbhouse:
We shouldn't set repo configs on the system if using a content template, so this sets custom repositories to be empty in that case. Also only builds repos/repo snapshots for payload and custom repos in the case of no template.


croissanne (approved):
thanks!

xbhouse:
thanks @croissanne 💟 feel free to merge when you get a chance

PR #1591, opened 2025-05-05
v1: don't redact passwords when starting a compose from blueprint (HMS-6047)

croissanne:
Similar to #1515, passwords need to be forwarded to composer in order for them to take effect.

### DIFF
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -531,7 +531,6 @@ func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) err
 	}
 	blueprint, err := BlueprintFromEntry(
 		blueprintEntry,
-		WithRedactedPasswords(),
 	)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


avitova (dismissed):
Nice, thanks! If you have any idea on how to make sure that something like this does not happen again, feel free to share:)

avitova (approved):
Okay, thanks! I guess it would be kinda hard to test if the request with password actually ends up with an image with password. Let's get this in. :)

lzap:
> Nice, thanks! If you have any idea on how to make sure that something like this does not happen again, feel free to share:)

Simon had a good idea yesterday in Brno office, we could refactor the (legacy) compose API POST request to use the blueprint handler instead. There could be some "constant" blueprint that is not actually stored in DB that would be used for "ad-hoc" compose requests, something like that.

lucasgarfield:
> Nice, thanks! If you have any idea on how to make sure that something like this does not happen again, feel free to share:)

In my opinion, we just proper end to end tests that cover all of our customizations with a boot test. Working on it. :smile:

PR #1592, opened 2025-05-06
floorist: Add new customization options to query

regexowl:
This adds hostname, kernel and firewall to query so we're able to track them in the metrics report.


regexowl:
Should the pr_check be passing?

lzap (approved):
Tests are flaky I think.

ochosi (approved):

PR #1598, opened 2025-05-12
ansible automation platform first boot registration (HMS-4840)

kingsleyzissou:
Add customization options for the Ansible Automation Platform which then converts the options into a first boot script.

### DIFF
--- a/internal/tmpl/aap-first-boot.tmpl.service
+++ b/internal/tmpl/aap-first-boot.tmpl.service
@@ -1,0 +1,15 @@
+Description=Ansible Automations Platform first boot registration
+ConditionFileIsExecutable=/usr/local/sbin/aap-first-boot-reg
+ConditionPathExists=!/var/local/.aap-first-boot-reg-done
+Wants=network-online.target
+After=network-online.target
+After=osbuild-first-boot.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/local/sbin/aap-first-boot-reg
+ExecStartPost=/usr/bin/touch /var/local/.aap-first-boot-reg-done
+RemainAfterExit=yes
+
+[Install]
+WantedBy=basic.target
--- a/internal/tmpl/render.go
+++ b/internal/tmpl/render.go
@@ -1,0 +1,45 @@
+package tmpl
+
+import (
+	"bytes"
+	"context"
+	"embed"
+	"fmt"
+	"log/slog"
+	"text/template"
+
+	"github.com/osbuild/image-builder-crc/internal/common"
+)
+
+//go:embed *.tmpl.*
+var templateFS embed.FS
+
+var templates *template.Template
+
+func init() {
+	var err error
+	templates, err = template.ParseFS(templateFS, "*.tmpl.*")
+	if err != nil {
+		panic(err)
+	}
+}
+
+func Render(ctx context.Context, name string, params any) (*string, error) {
+	var buf bytes.Buffer
+	slog.DebugContext(ctx, "rendering template", "name", name, "params", params)
+	err := templates.ExecuteTemplate(&buf, name, params)
+	if err != nil {
+		return nil, fmt.Errorf("error executing template: %w", err)
+	}
+	slog.DebugContext(ctx, "template rendered", "output", buf.String(), "template", true)
+
+	return common.ToPtr(buf.String()), nil
+}
+
+func RenderAAPServiceUnit(ctx context.Context) (*string, error) {
+	return Render(ctx, "aap-first-boot.tmpl.service", "")
+}
+
+func RenderAAPRegistrationScript(ctx context.Context, params AAPRegistrationParams) (*string, error) {
+	return Render(ctx, "aap-first-boot-script.tmpl.txt", params)
+}
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -1749,6 +1749,8 @@ components:
           $ref: '#/components/schemas/Installer'
         cacerts:
           $ref: '#/components/schemas/CACertsCustomization'
+        aap:
+          $ref: '#/components/schemas/AAPCustomization'
     Container:
       type: object
       required:
@@ -1984,6 +1986,26 @@ components:
           example: [ '---BEGIN CERTIFICATE---\nMIIC0DCCAbigAwIBAgIUI...\n---END CERTIFICATE---' ]
           items:
             type: string
+    AAPCustomization:
+      type: object
+      x-go-name: AAP
+      additionalProperties: false
+      required:
+        - ansible_controller_url
+        - job_template_id
+        - host_config_key
+      properties:
+        ansible_controller_url:
+          type: string
+          example: "example.towerhost.net"
+        job_template_id:
+          type: integer
+          example: 38
+        host_config_key:
+          type: string
+          example: "44d7507f2ead49af5fca80aa18fd24bc"
+        tls_certificate_authority:
+          type: string
     Ignition:
       type: object
       additionalProperties: false
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -1254,6 +1254,8 @@ components:
           description: |
             ID of the content template. A content template and snapshot date cannot both be specified.
             If a content template is specified, the snapshot date used will be the one from the content template.
+        aap_registration:
+          $ref: '#/components/schemas/AAPRegistration'
     ImageTypes:
       type: string
       enum:
@@ -1465,7 +1467,28 @@ components:
           description: |
             Determines whether a valid subscription manager (candlepin) identity is required to
             access this repository. Consumer certificates will be used as client certificates when
-            fetching metadata and content.
+             fetching metadata and content.
+    AAPRegistration:
+      type: object
+      x-go-name: AAPRegistration
+      additionalProperties: false
+      required:
+        - ansible_controller_url
+        - job_template_id
+        - host_config_key
+      properties:
+        ansible_controller_url:
+          type: string
+          example: "example.towerhost.net"
+        job_template_id:
+          type: integer
+          example: 38
+        host_config_key:
+          type: string
+          example: "44d7507f2ead49af5fca80aa18fd24bc"
+        tls_certificate_authority:
+          type: string
+          x-go-type-skip-optional-pointer: true
     PackagesResponse:
       type: object
       required:
@@ -1749,8 +1772,6 @@ components:
           $ref: '#/components/schemas/Installer'
         cacerts:
           $ref: '#/components/schemas/CACertsCustomization'
-        aap:
-          $ref: '#/components/schemas/AAPCustomization'
     Container:
       type: object
       required:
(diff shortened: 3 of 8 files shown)


internal/v1/handler_compose_image.go
```diff
@@ -1256,5 +1277,53 @@ func (h *Handlers) buildCustomizations(ctx echo.Context, cr *ComposeRequest, d *
 		}
 	}
 
+	if cust.Aap != nil {
+		// TODO: this doesn't handle the TLS certificate case yet.
+		// Also, the job template id might need to go into the API url
+		data := common.ToPtr(fmt.Sprintf(
+			"#!/usr/bin/bash\ncurl -s -i --data %q %s",
```
kingsleyzissou, line 1282: This is based on:
https://github.com/datacenter/cloudcenter-content/blob/bbb41be37f2b306d5fbf28f8a5c8b2a679104752/services/ansible/request_tower_configuration.sh#L3-L8

internal/v1/api.yaml
```diff
@@ -1980,6 +1982,25 @@ components:
...
+          type: string
+          example: "example.towerhost.net"
+        job_template_id:
+          type: integer
+          example: 38
+        host_config_key:
+          type: string
+          example: "44d7507f2ead49af5fca80aa18fd24bc"
+        tls_certificate:
+          type: string
```
lzap, line 2003: Nitpick: I am assuming this is the AAP CA. While CA essentially is a certificate too, I think this should probably read `tls_certificate_authority` but it is quite long. 🤷🏻

internal/v1/handler_compose_image.go
lzap: I would probably use `text/template` and `embed` Go packages here as the resulting code and templates are generally more readable and we plan to refactor Satellite registration this way as well so a utility template package could be a great fit. I do not insist tho.

internal/v1/api.yaml
```diff
@@ -1745,6 +1745,8 @@ components:
           $ref: '#/components/schemas/Installer'
         cacerts:
           $ref: '#/components/schemas/CACertsCustomization'
+        aap:
+          $ref: '#/components/schemas/AAPCustomization'
```
lzap, line 1749: Could you use `oapi-codegen` extended fields to customize the Go type to `AAP` here? The generated `Aap` is awful :-)

internal/v1/handler_compose_image.go
```diff
@@ -1308,6 +1306,16 @@ func (h *Handlers) buildCustomizations(ctx echo.Context, cr *ComposeRequest, d *
 			},
 		}
 
+		if cust.Aap.TlsCertificate != nil {
+			if res.Cacerts != nil && res.Cacerts.PemCerts != nil {
+				res.Cacerts.PemCerts = append(res.Cacerts.PemCerts, *cust.Aap.TlsCertificate)
```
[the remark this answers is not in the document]
lzap: Yes this is all that needs to be done.

internal/v1/handler_compose_image.go
```diff
@@ -1256,5 +1277,61 @@ func (h *Handlers) buildCustomizations(ctx echo.Context, cr *ComposeRequest, d *
...
+
+		if res.Services != nil && len(*res.Services.Enabled) > 0 {
+			if res.Services.Enabled != nil && len(*res.Services.Enabled) > 0 {
+				res.Services.Enabled = common.ToPtr(append(*res.Services.Enabled, fmt.Sprintf("%s.service", AAP_FIRST_BOOT_FILENAME)))
+			}
+		} else {
+			res.Services = &composer.Services{
+				Enabled: &[]string{fmt.Sprintf("%s.service", AAP_FIRST_BOOT_FILENAME)},
+			}
+		}
```
lzap, line 1327: The build-in function `append` does work fine with `nil` slice, you can append to `nil` slice and it will allocate it automatically. The code can be much more simple in here.

internal/v1/handler_compose_image.go
```diff
@@ -1256,5 +1277,61 @@ func (h *Handlers) buildCustomizations(ctx echo.Context, cr *ComposeRequest, d *
...
+
+		if res.Files != nil && len(*res.Files) > 0 {
+			res.Files = common.ToPtr(append(*res.Files, files...))
+		} else {
+			res.Files = &files
+		}
+	}
```
lzap, line 1334: The same thing as above, just append and Go will figure it out.

lzap:
Dropped some initial comments but I am gonna drag few people in as we discussed this already: @croissanne @mvo5 @achilleas-k @schuellerf 

We touched on this topic and we concluded so far that registration should not be part of customization, but rather it should be part of the sidecar also known as image/compose request. In fact, [registration in CLI](https://github.com/osbuild/image-builder-cli/pull/201) is separate from the main blueprint.

So I would like to propose to move AAP registration to image request for CRC API as well. Opinions?

internal/v1/handler_compose_image.go
kingsleyzissou: Ah thanks, yeah this was pretty ugly. I actually didn't know you could append to a nil slice. Will sort that out

internal/v1/handler_post_compose_test.go
```diff
@@ -2961,6 +2954,12 @@ func TestComposeCustomizations(t *testing.T) {
 							Type:    v1.UploadTypesAwsS3,
 							Options: uo,
 						},
+						AAPRegistration: &v1.AAPRegistration{
```
kingsleyzissou, line 2957: @lzap is this what you had in mind?

I can squash these commits into the previous ones if this is the route we want to go

achilleas-k:
> We touched on this topic and we concluded so far that registration should not be part of customization, but rather it should be part of the sidecar also known as image/compose request.

Ah, yes.  My concerns and thoughts on these things usually stem from the desire to separate things that have different _reusability cycles_ than he blueprint, i.e. things that might be reused in different blueprints or conversely things that might change between builds even when the blueprint is reused.  This doesn't necessarily need to map directly to how we separate things in the CRC API, but things would be nicer for everyone if we started moving towards a 1-to-1 mapping between the two interfaces.

internal/v1/handler_post_compose_test.go
lzap: Yes, tho, we need to hear from others.

lzap:
> started moving towards a 1-to-1 mapping between the two interfaces

Yes, since we did a ton of work on the common blueprint and also I am suffering the pain of making it `oapi-codegen` based, then V2 should really be 1:1 blueprint with some additional data.

internal/v1/api.yaml
```diff
@@ -1461,7 +1463,27 @@ components:
...
+          type: string
+          example: "example.towerhost.net"
+        job_template_id:
+          type: integer
+          example: 38
+        host_config_key:
+          type: string
+          example: "44d7507f2ead49af5fca80aa18fd24bc"
+        tls_certificate_authority:
+          type: string
```
lzap, line 1486: Nitpick: this could have `x-go-type-skip-optional-pointer` and then it will be just string, when empty it will be omitted as `omitempty` is added by default by the generator.

internal/v1/handler_post_compose_test.go
```diff
@@ -2946,6 +2941,100 @@ func TestComposeCustomizations(t *testing.T) {
...
+Type=oneshot
+ExecStart=/usr/local/sbin/aap-first-boot-reg
+ExecStartPost=/usr/bin/touch /var/local/.aap-first-boot-reg-done
+RemainAfterExit=yes
+
+[Install]
+WantedBy=basic.target
+WantedBy=multi-user.target
+WantedBy=graphical.target
+`),
```
lzap, line 3003: Can you drop both basic and graphical? This is a leftover I am going to fix in the UI.

internal/v1/handler_compose_image.go
```diff
@@ -24,6 +26,29 @@ import (
...
+ExecStart=/usr/local/sbin/aap-first-boot-reg
+ExecStartPost=/usr/bin/touch /var/local/.aap-first-boot-reg-done
+RemainAfterExit=yes
+
+[Install]
+WantedBy=basic.target
+WantedBy=multi-user.target
+WantedBy=graphical.target
+`
+)
```
lzap, line 50: Can you drop both basic and graphical? This is a leftover I am going to fix in the UI.

I also suggest to make use of `embed` package and extract this into `aap-first-boot.tmpl.service` so we will get syntax hightlighting benefits. It could have been generalized via attributes in the future as well into a generic template for all firstboot templates.

internal/v1/handler_compose_image.go
```diff
@@ -1256,5 +1279,74 @@ func (h *Handlers) buildCustomizations(ctx echo.Context, cr *ComposeRequest, d *
 		}
 	}
 
+	if aap := cr.ImageRequests[0].AAPRegistration; aap != nil {
+		tmpl := `#!/usr/bin/bash
+curl -s -i --data "host_config_key={{ .HostConfigKey }}" {{ .ServerAddress }}`
+
+		t, err := template.New("aap-script").Parse(tmpl)
```
lzap, line 1286: This could be extracted into a separate package together with templates. I did this on a different project and it worked well: https://github.com/foresterorg/forester/tree/main/internal/tmpl

lzap:
Dropped more random comments, I do not insist on any of these, the code looks good. Can we confirm the image request is the right spot @croissanne ? It aligns with the CLI and common blueprint nicely, I am really hoping for 1:1 customization for V2.

internal/v1/handler_compose_image.go
kingsleyzissou: Cool, I've added a separate package and just reused most of what you had there

lucasgarfield (approved):
lgtm, admin merging to unblock frontend work

lzap:
Sorry I missed the final review, GH will not tell if one just pushes or marks as ready for review I think, I missed it.

internal/tmpl/aap-first-boot.tmpl.service
```diff
@@ -0,0 +1,15 @@
...
+After=osbuild-first-boot.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/local/sbin/aap-first-boot-reg
+ExecStartPost=/usr/bin/touch /var/local/.aap-first-boot-reg-done
+RemainAfterExit=yes
+
+[Install]
+WantedBy=basic.target
```
lzap, line 15: So much nicer, it is missing the first line but I already saw a fix... :-)

internal/tmpl/render.go
```diff
@@ -0,0 +1,45 @@
...
+	return common.ToPtr(buf.String()), nil
+}
+
+func RenderAAPServiceUnit(ctx context.Context) (*string, error) {
+	return Render(ctx, "aap-first-boot.tmpl.service", "")
+}
+
+func RenderAAPRegistrationScript(ctx context.Context, params AAPRegistrationParams) (*string, error) {
+	return Render(ctx, "aap-first-boot-script.tmpl.txt", params)
+}
```
lzap, line 45: Nailed it, this is great and extensible for the future. Love how you debug it as well, tho, it may sometimes contain secrets or tokens which is possibly a problem. I would also only do one log records instead of two but this is just tiny nitpick.

lzap (commented):
Great.

PR #1600, opened 2025-05-12
v1: send template uuid/name in compose request (HMS-6297)

rverdile:
If a template is being used, the template name or UUID is sent in the compose request for system registration. If the RHEL version is >= 9.6, the name is used for registration. Any older version uses the UUID for registration.

JIRA: [HMS-6297](https://issues.redhat.com/browse/HMS-6297)

### DIFF
--- a/internal/v1/handler_post_compose_test.go
+++ b/internal/v1/handler_post_compose_test.go
@@ -2892,6 +2892,9 @@ func TestComposeCustomizations(t *testing.T) {
 							Id:       mocks.RepoPLID,
 						},
 					},
+					Subscription: &v1.Subscription{
+						Insights: true,
+					},
 				},
 				ImageRequests: []v1.ImageRequest{
 					{
@@ -2905,6 +2908,7 @@ func TestComposeCustomizations(t *testing.T) {
 					},
 				},
 			},
+			// content template with 2 rhel repos with registration
 			composerRequest: composer.ComposeRequest{
 				Customizations: &composer.Customizations{
 					CustomRepositories: nil,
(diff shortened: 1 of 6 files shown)


rverdile:
Depends on these composer changes: https://github.com/osbuild/osbuild-composer/pull/4712

lzap:
I think this should go into the compose request and not customization, @croissanne ?

lzap:
My point is that this is a console.redhat.com related stuff that should not go into customization, definitely not the common blueprint folks were not liking the idea during our meetings.

internal/v1/handler_post_compose_test.go
```diff
@@ -2921,6 +2925,16 @@ func TestComposeCustomizations(t *testing.T) {
...
+					Subscription: &composer.Subscription{
+						ActivationKey:       "",
+						BaseUrl:             "",
+						Insights:            true,
+						Rhc:                 common.ToPtr(false),
+						Organization:        "0",
+						ServerUrl:           "",
+						InsightsClientProxy: common.ToPtr(""),
+						TemplateUuid:        common.ToPtr(mocks.TemplateID),
+					},
```
lzap, line 2937: Is this hunk related?

jlsherrill:
This is definitely more 1), its meant to live with the blueprint and be re-used for every image build.  That said, it doesn't make sense to export and import this information with the blueprint (which i don't think is being done here).  

I'm curious why other RHC configuration options belong here but not this one?

lzap:
> @lzap I'm curious why other RHC configuration options belong here but not this one?

Insights-only related data should not be part of customization, I am working on a V2 blueprint format, schema and possibly CRC API V2 that will not have such fields and it will have to be stored separately. Take this as a more future-proofing of the change, if we choose to put it there we can refactor it later, I mean there will likely be V1 to V2 mapper anyway so no hard feelings.

internal/v1/handler_post_compose_test.go
rverdile: yes, this is needed for the test to pass

lzap (approved):

rverdile:
rebased to resolve merge conflict

croissanne (approved):

croissanne (approved):

croissanne:
It makes sense here to have it in the image request as the content template chosen in ibcrc also determines the repositories being sent to composer. So it's not a "pure" customisation.

PR #1601, opened 2025-05-13
New RHEL ga releases (HMS-6149)

croissanne:
Rhel 10 and 9.6

---

HMS-6149

HMS-6150


croissanne:
HMS-6149

croissanne:
HMS-6150

distributions/rhel-95/rhel-9.6
```diff
@@ -0,0 +1 @@
+rhel-9.6
```
lzap, line 1: This looks like a symlink, since the numbers do not match I wonder if this is some leftover?

croissanne: ah yes, that's not supposed to be there

internal/distribution/distroregistry_test.go
```diff
@@ -30,11 +30,14 @@ func TestDistroRegistry_List(t *testing.T) {
 		"rhel-93",
 		"rhel-94",
 		"rhel-95",
+		"rhel-9.6",
```
kingsleyzissou, line 33: Quick question out of curiosity. How come we change to a dot release here?

lzap (dismissed):

kingsleyzissou (approved):

croissanne:
still waiting for quay to be back online so konflux works

croissanne:
https://status.redhat.com/incidents/k7kvfvgfrbdf

internal/distribution/distroregistry_test.go
kingsleyzissou: Thanks. Ya, thought the change in composer happened a while back. Makes sense to drop the aliases... but obviously that can be done in a different PR at a later stage

ondrejbudai (approved):
LGTM, thanks!

PR #1602, opened 2025-05-14
oscap: add missing rhel9 and rhel10 distros

kingsleyzissou:
Enable oscap customizations for some missing rhel9 distros and for rhel10


kingsleyzissou:
I'm just generating the customizations now. Otherwise we could generate them with a workflow after this PR has landed.

croissanne (approved):

croissanne:
> I'm just generating the customizations now. Otherwise we could generate them with a workflow after this PR has landed.

whichever you prefer

kingsleyzissou:
Cool, I'll trigger the workflow, because the permissions get borked when I try it locally

lzap (approved):

PR #1604, opened 2025-05-15
v1: fix merging users when updating blueprint

croissanne:
Only merge users if there are existing users in the blueprint.


achilleas-k (approved):
!!

PR #1605, opened 2025-05-15
v1: aap service file missing header

kingsleyzissou:
There was a copy & paste error and the `[Unit]` header was missing from the service file.


croissanne (approved):
that'll do it!

PR #1606, opened 2025-05-16
v1: app change aap controller url example

kingsleyzissou:
This is a minor fix but the example uses a hostname and not the full url.

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -1483,7 +1483,7 @@ components:
       properties:
         ansible_controller_url:
           type: string
-          example: "example.towerhost.net"
+          example: "https://aap-gw.example.com"
         job_template_id:
           type: integer
           example: 38


lzap (approved):

internal/v1/api.yaml
```diff
@@ -1483,7 +1483,7 @@ components:
       properties:
         ansible_controller_url:
           type: string
-          example: "example.towerhost.net"
+          example: "https://example.towerhost.net"
```
ochosi, line 1486:
```suggestion
          example: "https://aap-gw.example.com"
```

This was a suggestion by someone on the #osbuild channel. Otherwise this seems like an improvement, too!

ochosi (dismissed):

internal/v1/api.yaml
kingsleyzissou: Cool, I'll update it quickly. I just took it from one of their scripts 😅

lzap (approved):

PR #1607, opened 2025-05-16
v1: pass template name in compose request for RHEL 10

rverdile:
This fixes a small issue the registering on boot for RHEL 10, when building from a template. When registering to insights, systems buit with RHEL 10 images (or higher) must use the template name.


lucasgarfield (approved):
lgtm

PR #1608, opened 2025-05-20
[wip] cmd/oscap: scap-security-guide source of truth

kingsleyzissou:
This PR updates the OpenSCAP customization generation to use the upstream definitions of the scap-security-guide rather than the RPM in Fedora. The upstream definitions are the source of truth for scap-security-guide.

### DIFF
--- a/.github/workflows/openscap.yml
+++ b/.github/workflows/openscap.yml
@@ -10,9 +10,18 @@ jobs:
   update-openscap:
     name: "Update OpenSCAP configs"
     if: github.repository == 'osbuild/image-builder-crc'
+    strategy:
+      matrix:
+        include:
+          - image: registry.fedoraproject.org/fedora:42
+            release: fedora
+          - image: quay.io/centos/centos:stream9
+            release: el9
+          - image: quay.io/centos/centos:stream10
+            release: el10
     runs-on: ubuntu-latest
     container:
-      image: registry.fedoraproject.org/fedora:42
+      image: ${{ matrix.image }}
 
     steps:
       - name: Install dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/openscap.yml
```diff
@@ -16,16 +16,25 @@ jobs:
...
+        run: dnf install -y tar gcc git-core go openscap gh
 
       - name: Check out code into the Go module directory
         uses: actions/checkout@v4
 
+      - name: Download & extract latest ssg
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run:  |
+         gh release download --repo ComplianceAsCode/content --pattern "*.zip" --output /tmp/ssg.zip
```
kingsleyzissou, line 28: Waiting for feedback from the OpenSCAP team on this.  It's a simpler approach which allows us to use the upstream rather than rpms in upstream.

.github/workflows/openscap.yml
```diff
@@ -35,13 +44,15 @@ jobs:
             exit "0"
           fi
 
+      # this is not ideal since it creates a PR per release
+      # with potentially 3 prs per workflow.
```
lzap, line 48: This will be pain and rebasing hell, could you explore creating artifacts from the jobs and then a separate job that depends on all, downloads all artifacts and make one PR?

lzap:
Other than that looks good.

.github/workflows/openscap.yml
kingsleyzissou: Yeah agreed, that was the next step because it would be a massive pain. I have a meeting with the OpenSCAP team next week about this, they're split about using centos for generating RHEL customizations... so I'll put the changes on pause for now.

kingsleyzissou:
Closing this for now after discussing this with the OpenSCAP team

PR #1609, opened 2025-05-21
Add distribution field to recommendation request and response models

lzap:
Introduce a distribution field in the recommendation request and response models to support additional filtering.

Refactor tests to accommodate these changes and also drop some unneeded test cases - the token has proper tests already there is no reason to test it again in the recommendation package. This is done in a separate commit.

### DIFF
--- a/internal/clients/recommendations/openapi.v3.gen.go
+++ b/internal/clients/recommendations/openapi.v3.gen.go
@@ -10,13 +10,15 @@ type GetModelVersion200Response struct {
 
 // RecommendPackageRequest defines model for RecommendPackageRequest.
 type RecommendPackageRequest struct {
+	Distribution        string   `json:"distribution"`
 	Packages            []string `json:"packages"`
 	RecommendedPackages int32    `json:"recommendedPackages"`
 }
 
 // RecommendationsResponse defines model for RecommendationsResponse.
 type RecommendationsResponse struct {
-	Packages []string `json:"packages"`
+	ModelVersion string   `json:"modelVersion"`
+	Packages     []string `json:"packages"`
 }
 
 // RefreshModel200Response defines model for RefreshModel200Response.
--- a/internal/v1/handler_recommend_package.go
+++ b/internal/v1/handler_recommend_package.go
@@ -2,7 +2,10 @@ package v1
 
 import (
 	"encoding/json"
+	"fmt"
+	"log/slog"
 	"net/http"
+	"slices"
 
 	"github.com/osbuild/image-builder-crc/internal/clients/recommendations"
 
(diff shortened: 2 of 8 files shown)


internal/clients/recommendations/openapi.v3.gen.go
```diff
@@ -10,13 +10,15 @@ type GetModelVersion200Response struct {
...
 type RecommendPackageRequest struct {
+	Distribution        string   `json:"distribution"`
 	Packages            []string `json:"packages"`
 	RecommendedPackages int32    `json:"recommendedPackages"`
 }
 
 // RecommendationsResponse defines model for RecommendationsResponse.
 type RecommendationsResponse struct {
-	Packages []string `json:"packages"`
+	ModelVersion string   `json:"modelVersion"`
```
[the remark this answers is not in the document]
lzap: The OpenAPI JSON document has been imported from recommendation project, there is nothing I can do from here.

internal/v1/handler_recommend_package.go
```diff
@@ -2,42 +2,50 @@ package v1
...
+	err := ctx.Bind(&req)
 	if err != nil {
 		return err
 	}
 
-	recommendationResponse, err := h.handleRecommendationsResponse(ctx, uploadPackageRequest)
+	resp, err := h.handleRecommendationsResponse(ctx, req)
 	if err != nil {
-		ctx.Logger().Errorf("Failed to recommendation package Response: %v", err)
+		slog.ErrorContext(ctx.Request().Context(), "Failed to recommendation package Response", "error", err)
```
kingsleyzissou, line 24: I know this is outside the scope of this PR, but since we are updating this line could we also fix the message?
Also because it's hard to tell what this is doing from the message alone. Maybe something like:
```
Failed to get package recommendation response
```

kingsleyzissou: Not a perfect message, but at least a bit better

internal/v1/handler_recommend_package.go
```diff
@@ -46,8 +54,16 @@ func (h *Handlers) handleRecommendationsResponse(ctx echo.Context, uploadPackage
 		return RecommendationsResponse{}, err
 	}
 
+	slog.InfoContext(ctx.Request().Context(), "Package recommendation",
```
kingsleyzissou, line 57: Nice!

kingsleyzissou (dismissed):
Just a minor nitpick inline. Otherwise LGTM

lzap:
Thanks, done. I noticed I suggested to keep lowercase for `slog` logging statements and wrote a linter for that, it will be enabled once there is a new linter release and we upgrade. Fixed these.

kingsleyzissou (approved):

PR #1612, opened 2025-06-02
refactor: add documentation to paginator

lzap:
Small nitpicks for the paginator, mostly docs, inconsistency for getters (some having the prefix while others dont) and useless memory consumption for count.

It is worth noting that in the review (https://github.com/osbuild/image-builder-crc/pull/1610) I expressed an opinion to avoid getters for the fields alltogether and export all the three fields. The type currently is mutable due to the fact that it does not copy the slice, so exposing those fields does not change a thing about immutability. However, type inference for generics only works with functions, so it would create it a little more verbose which I do not like:

```go
p := common.Paginator[int]{values, limit, offset}
```

Therefore I am not including such a change.

### DIFF
--- a/internal/common/pagination.go
+++ b/internal/common/pagination.go
@@ -2,13 +2,15 @@ package common
 
 import "fmt"
 
+// Paginator is a generic type that provides pagination functionality.
+// It allows you to paginate through a slice of items of any type T.
 type Paginator[T any] struct {
 	items  []T
-	count  int
 	limit  int
 	offset int
 }
 
+// NewPaginator creates a new Paginator instance with the provided items, limit, and offset.
 func NewPaginator[T any](items []T, limit, offset int) (*Paginator[T], error) {
 	if limit <= 0 {
 		return nil, fmt.Errorf("limit cannot be less than or equal to zero: %d", limit)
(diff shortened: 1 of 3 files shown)


internal/common/pagination.go
```diff
@@ -19,43 +21,46 @@ func NewPaginator[T any](items []T, limit, offset int) (*Paginator[T], error) {
...
 
-// GetPage returns the items for the current page based on the limit and offset.
+// Page returns the items for the current page based on the limit and offset.
 // It returns an empty slice if the offset is greater than or equal to the total count.
-func (p *Paginator[T]) GetPage() []T {
+func (p *Paginator[T]) Page() []T {
 	start := p.offset
 	end := start + p.limit
-	if start >= p.count {
+	if start >= len(p.items) {
```
thozza, line 49: Nitpick: you could just call `p.Count()`

internal/common/pagination.go
```diff
@@ -19,43 +21,46 @@ func NewPaginator[T any](items []T, limit, offset int) (*Paginator[T], error) {
...
 		return []T{}
 	}
-	if end > p.count {
-		end = p.count
+	if end > len(p.items) {
+		end = len(p.items)
```
thozza, line 53: ditto `p.Count()`

internal/common/pagination.go
```diff
@@ -19,43 +21,46 @@ func NewPaginator[T any](items []T, limit, offset int) (*Paginator[T], error) {
...
 	}
 	return p.items[start:end]
 }
 
 // LasetPageOffset returns the last page offset based on the total count and page size defined by the limit.
 func (p *Paginator[T]) LastPageOffset() int {
-	return CalculateLastPageOffset(p.count, p.limit)
+	return CalculateLastPageOffset(len(p.items), p.limit)
```
thozza, line 60: ditto `p.Count()`

thozza (approved):
Thanks.

In general, LGTM, I've added one nitpick.

lzap:
Done.

thozza (approved):
Amazing, thanks!

PR #1613, opened 2025-06-02
cmd/oscap: format configs to be readable

achilleas-k:
Use json.MarshalIndent() to make the test configuration files readable and easier to diff when they're updated.


mvo5 (approved):
Thank you

lzap (approved):

PR #1629, opened 2025-06-25
ci: introduce gobump gha

lzap:
Since gobump proven to be useful in `images`, I am proposing it for composer as well. We would drop other dependency bots and only use gobump. It bumps weekly.

This can only be tested after GHA exists and then it can be triggered against `upstream` branch manually to fine-tune it. It is the exact copy from `images` job: https://github.com/osbuild/images/blob/main/.github/workflows/gobump.yml

Also filed: https://github.com/osbuild/osbuild-composer/pull/4758

### DIFF
--- a/.github/workflows/gobump.yml
+++ b/.github/workflows/gobump.yml
@@ -1,0 +1,22 @@
+---
+name: "Updates Go dependencies via gobump"
+
+on:  # yamllint disable-line rule:truthy
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


lzap:
Rebased, it now uses the shared GHA directly from gobump project, it could not be used in the images project since it needs Fedora image there for tests, but for composer/crc using ubuntu image should work just fine.

PR #1637, opened 2025-07-02
Support MCP clientId for metrics

schuellerf:
For now we won't differentiate between possible local or hosted MCP deployments.
Still can be used for metrics UI / API vs MCP.

### DIFF
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -546,7 +546,9 @@ func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) err
 	}
 	composeResponses := make([]ComposeResponse, 0, len(blueprint.ImageRequests))
 	clientId := ClientId("api")
-	if ctx.Request().Header.Get("X-ImageBuilder-ui") != "" {
+	if ctx.Request().Header.Get("X-ImageBuilder-ui") == "mcp" {
+		clientId = "mcp"
+	} else if ctx.Request().Header.Get("X-ImageBuilder-ui") != "" {
 		clientId = "ui"
 	}
 	for _, imageRequest := range blueprint.ImageRequests {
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -57,6 +57,7 @@ require (
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mohae/deepcopy v0.0.0-20170929034955-c48cc78d4826 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+	github.com/oapi-codegen/oapi-codegen/v2 v2.4.1 // indirect
```
croissanne, line 60: hm, why was this necessary?

internal/v1/handler_blueprints.go
```diff
@@ -546,7 +546,9 @@ func (h *Handlers) ComposeBlueprint(ctx echo.Context, id openapi_types.UUID) err
 	}
 	composeResponses := make([]ComposeResponse, 0, len(blueprint.ImageRequests))
 	clientId := ClientId("api")
-	if ctx.Request().Header.Get("X-ImageBuilder-ui") != "" {
+	if ctx.Request().Header.Get("X-ImageBuilder-ui") == "mcp" {
+		clientId = "mcp"
+	} else {
```
croissanne, line 551: This behaviour is actually different now, the default changed from api to ui. 

It should be `else if ctx.Req… != ""`, making the default "api" again.

croissanne (changes requested):
Thanks, 1 behavioural change that I think we should avoid one q.

go.mod
schuellerf: I just ran the prepare source 🤔🤷
I'll check again in detail

internal/v1/handler_blueprints.go
schuellerf: Sorry, thanks - did it in a hurry 🫣

go.mod
schuellerf: `go generate ./...` creates this - not sure if this should be committed?
I remove the commit for now

PR #1642, opened 2025-07-07
renovate.json: disable gomod manager

croissanne:
Gobump is also enabled, and it can limit go version upgrades.


lzap:
Enabled auto-merge, the other option for my permissions is to rebase. I cannot merge right away unfortunately, which would be the ideal thing to do here.

PR #1644, opened 2025-07-08
aap: edit inputs for provisioning call

ksiekl:
The AAP registration needs slightly different input values than were expected at first.

First time diving into the CRC project and writing Go, so fingers crossed it all makes sense. No promises it’s perfect—there might be some rookie mistakes lurking :smile:


lzap:
For the context read https://issues.redhat.com/browse/HMS-8804 @kingsleyzissou

PR #1645, opened 2025-07-08
v1/oscap: handle policies without extra customizations

croissanne:
ssia, got a nil pointer error when handling a policy without extra customizations.


lzap (approved):

lzap:
Enabled automerge, update needed, cannot merge right away.

PR #1647, opened 2025-07-11
Restructure distro tests

croissanne:
distribution: use placeholder distributions in tests
    
    The v1 package already tests against the real distribution data, so
    it's safe to just expand the placeholder data a little.
    
    This results in one less place to update when the distribution data is
    changed. Now it's just the v1 package and the distribution data itself.


schuellerf (approved):

lzap (approved):
I was just confused by the changes, no need to change anything here.

PR #1650, opened 2025-07-16
templates: add patch_url to service environment

rverdile:
PATCH_URL is missing from the environment


rverdile:
the tests were complaining about a diff in the api specs so I updated those too

PR #1651, opened 2025-07-16
AAP: import/export (HMS-8836) (HMS-8837)

ksiekl:
AAP was in image_request, but since it is not an extra registration and more a firstboot, I moved it to customizations. Also edited tests for import and export.

Just one question - do I understand correctly that when exporting a bp, we want to redact certificates (I also removed config key), but when importing, we expect the user to have the certificates in the request? In my head it made sense, so hopefully that's ok :)

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -1258,8 +1258,6 @@ components:
           type: string
           description: |
             Name of the content template. Used when registering the system to Insights.
-        aap_registration:
-          $ref: '#/components/schemas/AAPRegistration'
     ImageTypes:
       type: string
       enum:
@@ -1783,6 +1781,8 @@ components:
           $ref: '#/components/schemas/Installer'
         cacerts:
           $ref: '#/components/schemas/CACertsCustomization'
+        aap_registration:
+          $ref: '#/components/schemas/AAPRegistration'
     Container:
       type: object
       required:
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -87,6 +87,12 @@ func (bb *BlueprintBody) RedactCertificates() {
 	bb.Customizations.Cacerts = nil
 }
 
+func (bb *BlueprintBody) RedactAAPRegistration() {
+	if bb.Customizations.AAPRegistration != nil {
+		bb.Customizations.AAPRegistration.HostConfigKey = ""
+	}
+}
+
 // Merges Password or SshKey from other User struct to this User struct if it is not set
 func (u *User) MergeExisting(other User) {
 	if u.Password == nil {
@@ -176,6 +182,12 @@ func WithRedactedPasswords() BlueprintBodyOption {
 func WithRedactedCertificates() BlueprintBodyOption {
 	return func(bp *BlueprintBody) {
 		bp.RedactCertificates()
+	}
+}
+
+func WithRedactedAAPRegistration() BlueprintBodyOption {
+	return func(bp *BlueprintBody) {
+		bp.RedactAAPRegistration()
 	}
 }
 
@@ -369,9 +381,11 @@ func (h *Handlers) ExportBlueprint(ctx echo.Context, id openapi_types.UUID) erro
 		blueprintEntry,
 		WithRedactedPasswords(),
 		WithRedactedCertificates(),
+		WithRedactedAAPRegistration(),
 		WithRedactedFiles([]string{
 			"/etc/systemd/system/register-satellite.service",
 			"/usr/local/sbin/register-satellite",
+			"/usr/local/sbin/aap-first-boot-reg",
 		}),
 	)
 	if err != nil {
--- a/internal/common/testdata/exported_blueprint.json
+++ b/internal/common/testdata/exported_blueprint.json
@@ -8,6 +8,12 @@
   ],
   "snapshot_date": "2012-12-20",
   "customizations": {
+    "aap_registration": {
+      "ansible_callback_url": "https://aap-gw.example.com/api/controller/v2/job_templates/42/callback/",
+      "host_config_key": "",
+      "skip_tls_verification": false,
+      "tls_certificate_authority": "-----BEGIN CERTIFICATE-----\nMIIC0DCCAbigAwIBAgIUI...\n-----END CERTIFICATE-----"
+    },
     "custom_repositories": [
       {
         "baseurl": [
--- a/internal/v1/api.go
+++ b/internal/v1/api.go
@@ -514,8 +514,9 @@ type CustomRepository struct {
 
 // Customizations defines model for Customizations.
 type Customizations struct {
-	Cacerts    *CACertsCustomization `json:"cacerts,omitempty"`
-	Containers *[]Container          `json:"containers,omitempty"`
+	AAPRegistration *AAPRegistration      `json:"aap_registration,omitempty"`
+	Cacerts         *CACertsCustomization `json:"cacerts,omitempty"`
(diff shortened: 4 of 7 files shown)


internal/v1/handler_blueprints.go
```diff
@@ -369,6 +382,7 @@ func (h *Handlers) ExportBlueprint(ctx echo.Context, id openapi_types.UUID) erro
 		blueprintEntry,
 		WithRedactedPasswords(),
 		WithRedactedCertificates(),
+		WithRedactedAAPRegistration(),
 		WithRedactedFiles([]string{
 			"/etc/systemd/system/register-satellite.service",
 			"/usr/local/sbin/register-satellite",
```
ksiekl, line 388: I was thinking that we should also probably redact the `/usr/local/sbin/aap-first-boot-reg` file, because it contains the config key. What I am not sure about is the `/etc/systemd/system/aap-first-boot-reg.service` file - should we redact that one as well?

kingsleyzissou: The service file in theory should be fine, it's really just the path to the firstboot script

internal/v1/handler.go
```diff
@@ -392,7 +392,7 @@ func parseComposeStatusError(ctx echo.Context, composeErr *composer.ComposeStatu
 		fallthrough
 	case 28: // osbuild target errors are added to details
 		if composeErr.Details != nil {
-			intfs := (*composeErr.Details).([]interface{})
+			intfs := (composeErr.Details).([]interface{})
```
kingsleyzissou, line 395: I'm trying to figure out this change, was this intentional?

kingsleyzissou (commented):
From a changes point of view it looks good, I don't have a huge amount of domain knowledge here though, so I'll wait for @lzap to review also

internal/v1/handler.go
ksiekl: Yes, the tests kept failing because of it

lzap:
Oh no this again. I am not the best one to ask if something belongs to blueprint (aka customizations) or not, I feel I was going in circles when we were discussing some similar fields. So I am going to drag @achilleas-k into the conversation. Let me CC @lucasgarfield and make a quick summary first:

* We added "firstboot" script in a way that it is completely constructed via custom files in the UI. This was our (team envision) first bigger contribution, now we know that was not the best decision.
* Then we added Satellite registration following the same design pattern - the firstboot for satellite is also constructed in the UI.
* We have realized that firstboot and satellite should have been a proper customization in blueprint processed by images. This was not yet fixed and I do not believe there is a Jira for this.
* For AAP registration, we have decided to process it in the CRC backend, again, it creates file customizations like the UI.
* One customization was properly added as a tho, it was the "cacert" which turned out to be useful to stand on its legs from the day one - I added this and it is being used in both registrations.

So we ended up with a mixture of three approaches for the three little features for CRC:

* Firstboot and Satellite handled in the UI.
* AAP handled in the CRC.
* CA cert as the only "sub-customization" that is properly handled by images library.

Now we have this secondary problem and the reason of this PR is where the registration fields do belong - is that a customization? Or is that a image request thing? I think we need this to be settled for once with a plan going forward. I am not happy about the situation I initiated and then left building UBP. Tho I have to say the clunky state of the current BP might play a role in how it ended now :-)

Since I am finishing off with the initial work of UBP and work will likely shift more onto other folks for the integration, I would like to step up and take this effort under my umbrella and lead it towards the ideal state which would be proper fisrtboot and satellite/aap registration customizations in the BP/UBP.

Shall we meet, discuss and plan this? Cheers.

lzap:
@ksiekl you brought very good points, but I would like to stop for a moment, discuss how to approach this and plan accordingly. The AAP work is part of a bigger technical gap that I think we should clear first. But to answer your concerns:

Registration data is part of image request for a reason, we thought customization is not a good place for registration-related things despite there already is a RHSM registration. I am not completely on board with this honestly.

Redacting generated firstboot files is a very good point, yes. Tho I will say if we refactor this into a proper customization handled by images library, there will be no need of this as the files will be created during manifest generation.

kingsleyzissou:
> We have realized that firstboot and satellite should have been a proper customization in blueprint processed by images. This was not yet fixed and I do not believe there is a Jira for this.

Yeah this keeps coming back to bite us. Let me create a jira for it! 

> Redacting generated firstboot files is a very good point, yes. Tho I will say if we refactor this into a proper customization handled by images library, there will be no need of this as the files will be created during manifest generation.

Also might be an opportunity for @ksiekl to look at other parts of the stack

lzap:
> Also might be an opportunity for @ksiekl to look at other parts of the stack

Absolutely, apologies for being a bit protective from composer/images blueprints work. By all means, it does not bite or anything! :-D

lzap:
I just noticed this is a PR with two commits. If you need the import/export patch in, I think this discussion can be separate, in that case just move import/export commit into a separate PR and rename this PR to "AAP: move aap registration from image_request to customization" instead so we can continue the discussion around the refactoring if you want.

ksiekl:
@lzap since import and export depends directly on the structure of the bp request, I would leave these commits together if it's ok with you. Creating a separate PR with i/e with aap in image_request now would be useless extra work if we decide to keep the aap in customizations, and I would have to rewrite it again to the same state it is in now.

I'd like to wait for the conversation about the customizations vs. image_request to be resolved and make changes accordingly.

lzap:
> My thoughts on the customization topic.

I was hoping you would comment on the registration in customization vs image request. :-D

I am going to work on firstboot customization full support because that is bothering me the most. I think it is okay to leave AAP as is perhaps put Satellite on par with this solution so at least we have a parity. Once firstboot is done, I think it allows for a nice workaround for AAP/Satellite users who would need to do on-prem registrations - just add few lines here and there and you are good to go.

lzap:
We discussed this after the meeting. Summary:

* Make AAP/Sat a proper images customization.
* Move AAP CRC API registration from image request to customization.
* Discussion about what belongs into customization should be part of the UBP effort.

internal/v1/api.yaml
```diff
@@ -1783,6 +1781,8 @@ components:
           $ref: '#/components/schemas/Installer'
         cacerts:
           $ref: '#/components/schemas/CACertsCustomization'
+        aap_registration:
+          $ref: '#/components/schemas/AAPRegistration'
```
lzap, line 1785: This is an API breaking change, will UI be ready when we merge this? It will go straight into stage.

internal/v1/handler_blueprints.go
```diff
@@ -87,6 +87,13 @@ func (bb *BlueprintBody) RedactCertificates() {
 	bb.Customizations.Cacerts = nil
 }
 
+func (bb *BlueprintBody) RedactAAPRegistration() {
+	if bb.Customizations.AAPRegistration != nil {
+		bb.Customizations.AAPRegistration.TlsCertificateAuthority = ""
```
lzap, line 92: TLS Certificate is not a secret, the secret part is a key that is not stored in this field. You may delete this line.

lzap (changes requested):
All good, just do not drop the TLS cert during secret redaction.

internal/v1/api.yaml
achilleas-k: Is it?  It's just additive, isn't it?

lzap (approved):

ksiekl:
Sorry, needed to fix the test after tls certificate change

lzap (approved):

lzap:
Leaving the merge on @kingsleyzissou thanks great work.

kingsleyzissou (approved):

PR #1671, opened 2025-08-28
api: Implement policy blueprint snapshot storage for compliance tracking

mgold1234:
Store policy-generated blueprint snapshots to track compliance policy

This would be saved when:
user creates blueprint with a policy
User updates the blueprint and changes the policy
User 'fixes' lint warnings with compliance differences

Database changes:
Add blueprint_policy_snapshots table to store policy blueprint JSON Implement CRUD operations: UpsertBlueprintPolicySnapshot, GetBlueprintPolicySnapshot, DeleteBlueprintPolicySnapshot
Add CASCADE delete constraint for automatic cleanup


lzap:
Why this was suggested to be a separate table @jlsherrill ? I am assuming this to be 1:1 relation, because doing this N:1 creates another complexity on top of versioning we already have. That is, there is a table named `blueprints` that is versioned and it has a `metadata` column which is JSON. Why not to store it there? It is currently [fully exported by the API](https://github.com/osbuild/image-builder- bcrc/blob/ded8bb178bd25945df20863a148862474fcee655/internal/v1/api.yaml#L1136) which I see as a plus, but it can be redacted if needed. The data is stored as JSONB this means it is efficiently encoded and compressed. This field will never be indexed although that is something Postgres can also do in JSON fields.

Was the plan to do this as many-to-one relation with possibility to build an UI to update those "snapshots"? If that is the case, then I also think embedding this into versioned blueprint metadata is better because then we can use all the versioning UI infrastructure we already have and everything will be more simple, more linear.

lzap:
Correction, I was slightly confused [since we separated blueprint body](https://github.com/osbuild/image-builder-crc/blob/main/internal/db/migrations-tern/009_create_table_blueprint_versions.sql) some time ago from `blueprint` as `blueprint_version`. Instead of storing the data in `blueprint.metadata`, I am updating my suggestion to: let's add a new flag in `blueprint_version` that would be JSONB and put the data in so we can extend it with other possible services in the future the same way.

Technically, a one-to-one separate table is possible but it is just unnecessary extra join I think. We use JSON/JSONB quite a lot in image builder (crc, composer) and it works great.

lzap (changes requested):
Looks good, a lot of code is copied and it can be all avoided.

internal/v1/handler_blueprints.go
```diff
@@ -144,6 +146,58 @@ func (u *User) MergeForUpdate(userData []User) error {
...
+	serviceSnapshots := &db.ServiceSnapshots{
+		Compliance: &db.ComplianceSnapshot{
+			PolicyId:            compl.PolicyId,
+			PolicyBlueprintToml: policyToml,
+			CreatedAt:           time.Now(),
+		},
+	}
+
+	ctx.Logger().Infof("Built compliance snapshot: policy_id=%s, toml_size=%d bytes",
+		compl.PolicyId, len(policyToml))
```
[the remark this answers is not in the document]
lzap: Also you created way too many INFO log records, please prefer DEBUG level which is turned off on production.

I think with DEBUG level it is okay to log the full TOML for easier debugging on stage, these documents are short. Or maybe @croissanne has a preference on that.

internal/v1/api.yaml
```diff
@@ -1102,6 +1102,10 @@ components:
             Array of image requests. Having more image requests in a single blueprint is currently not supported.
         customizations:
           $ref: '#/components/schemas/Customizations'
+        service_snapshots:
+          $ref: '#/components/schemas/ServiceSnapshots'
+          description: |
+            Service snapshot information including compliance data
```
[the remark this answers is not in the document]
croissanne: Why do we need to expose this in the first place? I think this is only relevant for the linting endpoint.

internal/v1/handler_blueprints.go
```diff
@@ -144,6 +147,88 @@ func (u *User) MergeForUpdate(userData []User) error {
 	return nil
 }
 
+// Helper function to build service snapshots if compliance policy exists
+func (h *Handlers) buildServiceSnapshots(ctx echo.Context, customizations *Customizations, distribution Distributions) (*db.ServiceSnapshots, error) {
```
croissanne, line 151: Could you split out any api changes to a separate commit? It makes it a bit hard to review.

lzap:
> Why do we need to expose this in the first place? I think this is only relevant for the linting endpoint.

Is there any reason against exposing it tho? Troubleshooting is my argument, sure, we can drop the whole blob to structure log as well, that is fine too. It is not a must, but accessing database is not super easy.

internal/v1/handler_blueprints.go
```diff
@@ -283,7 +389,9 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
 		}
 		return err
 	}
-	ctx.Logger().Infof("Inserted blueprint %s", id)
+	slog.InfoContext(ctx.Request().Context(), "inserted blueprint with service snapshots",
+		"id", id)
```
lzap, line 393: This should have been debug.

internal/v1/handler_blueprints.go
```diff
@@ -788,13 +918,43 @@ func (h *Handlers) FixupBlueprint(ctx echo.Context, id openapi_types.UUID) error
 	}
 	desc := common.FromPtr(blueprintRequest.Description)
 
-	err = h.server.db.UpdateBlueprint(ctx.Request().Context(), uuid.New(), blueprintEntry.Id, userID.OrgID(), blueprintRequest.Name, desc, body)
+	slog.InfoContext(ctx.Request().Context(), "starting buildServiceSnapshots during fixup",
+		"blueprint_id", blueprintEntry.Id,
+		"distribution", blueprintRequest.Distribution,
+		"has_openscap", blueprintRequest.Customizations.Openscap != nil)
```
lzap, line 924: Consider lowering amount of debug messages to just one at the end, or two begin/end of the function. If you want to keep this one okay, but please not info level these are turned on in production.

lzap (changes requested):
Nice, mostly nitpicks around logging.

cmd/image-builder-db-test/main_test.go
```diff
@@ -460,7 +460,37 @@ func testGetBlueprintComposes(ctx context.Context, t *testing.T) {
...
+		},
+	}
+	bodyJson1, err := json.Marshal(body1)
+	require.NoError(t, err)
+
+	policyBlueprint1 := `[packages]
+                                   packages = ["vim", "git", "curl", "admin"]
+     
+                                   [customizations]
+                                   hostname = "rhel8-server"`
```
lzap, line 482: Formatting is little off, no biggie.

internal/v1/handler_blueprints.go
```diff
@@ -787,14 +894,37 @@ func (h *Handlers) FixupBlueprint(ctx echo.Context, id openapi_types.UUID) error
 		return err
 	}
 	desc := common.FromPtr(blueprintRequest.Description)
+	slog.DebugContext(ctx.Request().Context(), "starting buildServiceSnapshots during fixup",
+		"blueprint_id", blueprintEntry.Id,
+		"distribution", blueprintRequest.Distribution,
+		"has_openscap", blueprintRequest.Customizations.Openscap != nil)
```
lzap, line 900: I think I do not see a nil check for `bpr.Customizations` above, this could be possibly a nil pointer error. If not, disregard.

internal/db/db.go
```diff
@@ -62,6 +63,15 @@ type BlueprintWithNoBody struct {
 	LastModifiedAt time.Time
 }
 
+type ServiceSnapshots struct {
+	Compliance *ComplianceSnapshot `json:"compliance,omitempty"`
+}
+
+type ComplianceSnapshot struct {
+	PolicyId        uuid.UUID `json:"policy_id"`
+	PolicyBlueprint string    `json:"policy_blueprint"`
```
croissanne, line 72: I wonder why we're storing the toml, and not the customizations directly (achieved by going through `lintOpenscap`, it also returns customizations.
```
	var cust Customizations
	_, err := h.lintOpenscap(ctx, &cust, true, distro, policy.String())
```

That way if the output we get from the compliance service ever changes, we won't need to keep support for converting all blueprint versions.

croissanne (changes requested):
Thank you, I think this already a much more manageable change. My main concern is storing the toml instead of the customizations (after the conversion from toml to the customizations).

Regarding the commit message:
- generally we use `v1` instead of `api` as a topic;
- I thinkg splitting out the `db` changes and the `v1` changes would help;
- why does the commit message keep specifying that it's `atomic`?
- "comprehensive validation" is a bit of an overstatement, I think it just checks if it's null or not ;)
- I'm not sure if it needs to have the 'implements x data structures' in there;
- I think it could use some more information about why exactly we're doing this (so we can diff policies and inform users when customizations that were previous added by a policy but are no longer required by the policy, might be superfluous, and that they might want to remove those (but they could also have added them manually so only inform)).

internal/db/db.go
mgold1234: good point @croissanne! This is definitely worth discussing. You're right that storing `Customizations` directly would have several technical advantages:
No TOML parsing overhead, 
Changes in compliance service output format won't break old snapshots
storing TOML:
Clear record of what was actually produced from each policy,
 Matches the blueprint API format users expect
 he original requirement specifically mentions "store the blueprint toml generated from a policy" - do we need the exact TOML output for: 
Compliance auditing?
Reproducing historical blueprint states?
WDYT? should we storing `Customizations` instead of TOML?

internal/v1/handler_blueprints.go
```diff
@@ -237,7 +309,12 @@ func (h *Handlers) CreateBlueprint(ctx echo.Context) error {
 
 	id := uuid.New()
 	versionId := uuid.New()
-	ctx.Logger().Infof("Inserting blueprint: %s (%s), for orgID: %s and account: %s", blueprintRequest.Name, id, userID.OrgID(), userID.AccountNumber())
+	slog.DebugContext(ctx.Request().Context(), "inserting blueprint",
```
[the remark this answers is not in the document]
mgold1234: @croissanne I keep the slog log, do you want me to change it?

lzap (dismissed):
Great work, I was wondering why TOML was used and I thought it is the only way because OpenSCAP tool generates TOML. I am glad this is possible with JSON as what Sanne said is very important for UBP.

internal/v1/handler_blueprints_test.go
```diff
@@ -1696,3 +1696,238 @@ func TestFixupBlueprint(t *testing.T) {
...
+		blueprintCreated := false
+		for _, bp := range blueprints {
+			if bp.Name == "BlueprintShouldRollback" {
+				blueprintCreated = true
+				break
+			}
+		}
+
+		if blueprintCreated {
+			t.Errorf("Blueprint was created despite error - this suggests rollback failed")
```
croissanne, line 1797: This can be removed entirely with just `require.Empty(t, blueprints)`

internal/v1/handler_blueprints_test.go
```diff
@@ -1696,3 +1696,238 @@ func TestFixupBlueprint(t *testing.T) {
...
+		} else {
+			t.Logf("Blueprint was not created with status %d - this is actually fine for our rollback test", statusCode)
+		}
+	}
+
+	blueprints, _, err := dbase.GetBlueprints(ctx, "000000", 100, 0)
+	require.NoError(t, err)
+
+	for _, bp := range blueprints {
```
croissanne, line 1806: This is testing the same thing twice.

internal/v1/handler_blueprints_test.go
```diff
@@ -1696,3 +1696,238 @@ func TestFixupBlueprint(t *testing.T) {
...
+
+	require.NoError(t, dbase.DeleteBlueprint(ctx, result.Id, "000000", "000000"))
+}
+
+func TestBlueprintCreationRollbackOnPolicyFailure(t *testing.T) {
+	ctx := context.Background()
+	dbase, srvURL, shutdownFn := makeTestServer(t, nil)
+	defer shutdownFn(t)
+
+	body := map[string]interface{}{
```
croissanne, line 1759: Could you add a comment here that this is because there are no tailorings available for `rhel-9` in the compliance mock? Isn't immediately obvious when reading the test.

internal/v1/handler_blueprints_test.go
```diff
@@ -1696,3 +1696,238 @@ func TestFixupBlueprint(t *testing.T) {
...
+	}
+
+	ctx := context.Background()
+	dbase, srvURL, shutdownFn := makeTestServer(t, nil)
+	defer shutdownFn(t)
+
+	body := map[string]interface{}{
+		"name":        "BlueprintWithCompliance",
+		"description": "Blueprint with compliance policy",
+		"customizations": map[string]interface{}{
```
croissanne, line 1712: Let's keep the blueprint minimal here, with only the openscap customization set.

internal/v1/handler_blueprints_test.go
```diff
@@ -1696,3 +1696,238 @@ func TestFixupBlueprint(t *testing.T) {
...
+func TestBlueprintCreationRollbackOnPolicyFailure(t *testing.T) {
+	ctx := context.Background()
+	dbase, srvURL, shutdownFn := makeTestServer(t, nil)
+	defer shutdownFn(t)
+
+	body := map[string]interface{}{
+		"name":        "BlueprintShouldRollback",
+		"description": "Blueprint that should be rolled back",
+		"customizations": map[string]interface{}{
+			"packages": []string{"nginx"},
```
croissanne, line 1763: Let's keep the blueprint minimal here, with only the openscap customization set.

croissanne (changes requested):
Thank you, I think the handler looks good now. But the tests need some work.

internal/v1/handler_blueprints_test.go
```diff
@@ -1640,14 +1640,23 @@ func TestFixupBlueprint(t *testing.T) {
 	}))
 
 	cases := []struct {
-		blueprint  v1.BlueprintBody
+		createBody map[string]any
```
croissanne, line 1643: With `map[string]any` I guess the above customizations aren't needed. But I'd prefer `v1.BlueprintBody` because that way we get type checking. To generate a valid upload request you can just use 
```
	var uo v1.UploadRequest_Options
	require.NoError(t, uo.FromAWSS3UploadRequestOptions(v1.AWSS3UploadRequestOptions{}))
```
and create it like
```
						UploadRequest: v1.UploadRequest{
							Type:    v1.UploadTypesAwsS3,
							Options: uo,
						},
```

also not sure if the rename from blueprint -> createBody makes it clearer

croissanne (changes requested):
Thanks, 1 more thing :)

internal/v1/handler_blueprints.go
```diff
@@ -144,6 +145,51 @@ func (u *User) MergeForUpdate(userData []User) error {
...
+
+	var compl OpenSCAPCompliance
+	compl, err := customizations.Openscap.AsOpenSCAPCompliance()
+	if err != nil {
+		slog.ErrorContext(ctx.Request().Context(), "error in AsOpenSCAPCompliance", "error", err.Error())
+		return nil, nil
+	}
+	if compl.PolicyId == uuid.Nil {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
mgold1234: we have this log ```
'slog.DebugContext(ctx.Request().Context(), "built compliance snapshot",
		"policy_id", compl.PolicyId,
		"distribution", distribution)'
```
		few rows later, I think its enough and cover the policyID for that function

File: internal/v1/handler_blueprints_test.go [hunk 17]
```diff
@@ -1639,16 +1639,33 @@ func TestFixupBlueprint(t *testing.T) {
 		PolicyId: uuid.MustParse(mocks.PolicyID2),
 	}))
 
+	var uo v1.UploadRequest_Options
+	require.NoError(t, uo.FromAWSS3UploadRequestOptions(v1.AWSS3UploadRequestOptions{}))
+
 	cases := []struct {
-		blueprint  v1.BlueprintBody
-		lintErrors []v1.BlueprintLintItem
+		blueprint   v1.BlueprintBody
```
croissanne, line 1646: Just change the type from `v1.BlueprintBody` to `v1.CreateBlueprintRequest`, then there's no need to define name and description separately.

croissanne:
@lzap @avitova need 1 more ack, could you give it another look?

internal/v1/handler_blueprints.go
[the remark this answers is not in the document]
lzap: I will say @avitova 's thinking is good if the PolicyId being Nil is something special we should know about. I honestly do not know if this is important enough and like an error state coming from compliance. So I am not against this, it is up for consideration.

Generally, I agree that successful path should do one debug logging per function, typically at the very end where as much as important data can be added via structured fields. But this is sort of a non-standard path so I would consider it.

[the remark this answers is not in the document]
avitova: Okay, I agree, thanks for considering it and exploring!

PR #1675, opened 2025-09-01
Add distro definitions for RHEL 9.8 and 10.2 nightly (HMS-9230)

thozza:
SSIA, plus remove definitions for 9.6 and 10.0 nightly, because these are already GA.


kingsleyzissou (commented):
The failing test seems relevant but that looks like a small fix.
Otherwise lgtm

thozza:
> The failing test seems relevant but that looks like a small fix. Otherwise lgtm

Yeah, thanks. I have a hard time running the tests locally, so this is the unfortunate result of it...

PR #1678, opened 2025-09-03
oscap: request tailoring for policies if none

tkoscieln:


internal/v1/mocks/compliance.go
```diff
@@ -77,6 +77,20 @@ func tailorings(w http.ResponseWriter, r *http.Request) {
...
+	if tutils.AuthString0 != r.Header.Get("x-rh-identity") {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.Header().Set("Content-Type", "application/json")
+	_, err := w.Write([]byte("{}"))
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
```
lzap, line 90: You can ignore both number of bytes written and error because response writer is an in-memory write.

tkoscieln: Changed to the error being ignored as well

internal/clients/compliance/client.go
```diff
@@ -126,6 +127,41 @@ func (cc *ComplianceClient) PolicyDataForMinorVersion(ctx context.Context, major
 }
 
 func (cc *ComplianceClient) PolicyCustomizations(ctx context.Context, majorVersion, minorVersion int, policyID string) (*blueprint.Blueprint, error) {
+	// First check if the tailorings exist for the policy
+	tailoringsGetResp, err := cc.request(ctx, "GET", fmt.Sprintf("%s/policies/%s/tailorings", cc.url, policyID))
```
tkoscieln, line 131: One thing I am not sure about is whether we should send the minor version in this request as well. Because it may happen that the policy has a tailoring, but for a different minor version than we want to use, so the code wouldn't create the tailorings even though we don't have them.

croissanne: Yes, definitely. This can be accomplished by adding the filter= query param
```
{
            "name": "filter",
            "in": "query",
            "required": false,
            "description": "Query string to filter items by their attributes. Compliant with <a href=\"https://github.com/wvanbergen/scoped_search/wiki/Query-language\" target=\"_blank\" title=\"github.com/wvanbergen/scoped_search\">scoped_search query language</a>. However, only `=` or `!=` (resp. `<>`) operators are supported.<br><br>Tailorings are searchable using attributes `os_minor_version`<br><br>(e.g.: `(field_1=something AND field_2!=\"something else\") OR field_3>40`)",
            "schema": {
              "type": "string"
            }
          }
```
or simply going through the response and see if there's a tailoring matching the requested minor version.

internal/clients/compliance/client.go
```diff
@@ -126,6 +131,48 @@ func (cc *ComplianceClient) PolicyDataForMinorVersion(ctx context.Context, major
...
+
+	var tailoringsResponse v2TailoringsResponse
+	err = json.NewDecoder(tailoringsGetResp.Body).Decode(&tailoringsResponse)
+	if err != nil {
+		return nil, err
+	}
+
+	if tailoringsGetResp.StatusCode == http.StatusOK && len(tailoringsResponse.Data) == 0 {
+		// If there are no tailorings for the policy, we need to create one, otherwise the request for tailoring file will return 404
+		// see https://issues.redhat.com/browse/RHINENG-16877
```
croissanne, line 149: Hm generally we don't add issues in comments / code / commits unless they're public.

internal/clients/compliance/client.go
```diff
@@ -126,6 +131,48 @@ func (cc *ComplianceClient) PolicyDataForMinorVersion(ctx context.Context, major
...
+		var jsonStr = []byte(fmt.Sprintf(`{"os_minor_version": %d}`, minorVersion))
+		tailoringsReq, err := http.NewRequest("POST", fmt.Sprintf("%s/policies/%s/tailorings", cc.url, policyID), bytes.NewBuffer(jsonStr))
```
croissanne, line 151: why not let this request go via the `request` helper? Should be easy enough to add POST support (accepting a `body io.Reader`, see the composer client for some inspiration).

internal/clients/compliance/client.go
```diff
@@ -126,6 +131,48 @@ func (cc *ComplianceClient) PolicyDataForMinorVersion(ctx context.Context, major
...
+		}
+		id, ok := identity.GetIdentityHeader(ctx)
+		if !ok {
+			return nil, fmt.Errorf("unable to get identity from context")
+		}
+		tailoringsReq.Header.Add("x-rh-identity", id)
+		tailoringsReq.Header.Add("content-type", "application/json")
+
+		tailoringPostResp, err := strc.NewTracingDoer(cc.client).Do(tailoringsReq)
+		fmt.Println("tailoringPostResp", tailoringPostResp)
```
croissanne, line 163: leftover debugging?

croissanne (changes requested):
Thank you! This is looking very nice. You're correct about the minor version. And the client could be simplified a little.

internal/clients/compliance/client.go
tkoscieln: Removed

internal/clients/compliance/client.go
```diff
@@ -126,6 +131,48 @@ func (cc *ComplianceClient) PolicyDataForMinorVersion(ctx context.Context, major
...
+	var tailoringsResponse v2TailoringsResponse
+	err = json.NewDecoder(tailoringsGetResp.Body).Decode(&tailoringsResponse)
+	if err != nil {
+		return nil, err
+	}
+
+	if tailoringsGetResp.StatusCode == http.StatusOK && len(tailoringsResponse.Data) == 0 {
+		// If there are no tailorings for the policy, we need to create one, otherwise the request for tailoring file will return 404
+		// see https://issues.redhat.com/browse/RHINENG-16877
+		var jsonStr = []byte(fmt.Sprintf(`{"os_minor_version": %d}`, minorVersion))
```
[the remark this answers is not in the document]
tkoscieln: Do you mean for the following POST request? :thinking:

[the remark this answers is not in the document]
tkoscieln: Oh got it, added for the body in the POST request

internal/clients/compliance/client.go
```diff
@@ -14,6 +15,15 @@ import (
 	"github.com/redhatinsights/identity"
 )
 
+// Use custom types because OpenAPI has version 3.1 and oapi-codegen doesn't support it yet
+type v2TailoringsResponse struct {
```
croissanne, line 19: Nitpick, could you move the type definitions under the error and compliance client stuff? Under `v2PolicyResponse` and `v2PolicyData` ?

croissanne (changes requested):
one nitpick, otherwise lgtm! thank you

internal/clients/compliance/client.go
tkoscieln: Should be moved now :+1:

croissanne (approved):
Perfect, thank you!

internal/clients/compliance/client.go
```diff
@@ -44,8 +45,8 @@ func NewClient(conf ComplianceClientConfig) *ComplianceClient {
 	}
 }
 
-func (cc *ComplianceClient) request(ctx context.Context, method, url string) (*http.Response, error) {
-	req, err := http.NewRequestWithContext(ctx, method, url, nil)
+func (cc *ComplianceClient) request(ctx context.Context, method, url string, body io.ReadSeeker) (*http.Response, error) {
```
croissanne, line 48: This could just be an `io.Reader`, as we're not seeking. But it doesn't matter really.

PR #1679, opened 2025-09-04
ci: enable changelog for gobump

lzap:
I added a new feature to gobump - every update will now come with a changelog. Due to size, these will be links to gists for each PR. Example: https://gist.github.com/lzap/b46a70514be80685d9e1b9a041dfdb5c


croissanne (approved):
neat

PR #1685, opened 2025-09-07
Linting

mgold1234:


internal/v1/handler_blueprints.go
```diff
@@ -405,13 +405,13 @@ func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID, params
 	return ctx.JSON(http.StatusOK, blueprintResponse)
 }
 
-func (h *Handlers) lintBlueprint(ctx echo.Context, blueprint *BlueprintBody, fixup bool) ([]BlueprintLintItem, error) {
+func (h *Handlers) lintBlueprint(ctx echo.Context, blueprint *BlueprintBody, fixup bool, blueprintEntry *db.BlueprintEntry) ([]BlueprintLintItem, error) {
```
lzap, line 408: Please do not add a parameter that carries the same information: blueprintEntry already contains `Body` so there is no reason to pass `blueprint *BlueprintBody` anymore.

I highly suggest to this refactoring in completely isolated commit - this makes it so much easier not only to review but it also helps with testing and git history. Once this is done, you can make use of the newly introduced parameter in the next commit immediately.

internal/v1/handler_oscap.go
```diff
@@ -193,7 +195,7 @@ func (h *Handlers) GetOscapCustomizationsForPolicy(ctx echo.Context, policy uuid
 	return ctx.JSON(http.StatusOK, cust)
 }
 
-func (h *Handlers) lintOpenscap(ctx echo.Context, cust *Customizations, fixup bool, distro Distributions, policy string) ([]BlueprintLintItem, error) {
+func (h *Handlers) lintOpenscap(ctx echo.Context, cust *Customizations, fixup bool, distro Distributions, policy string, blueprintEntry *db.BlueprintEntry) ([]BlueprintLintItem, error) {
```
lzap, line 198:
```suggestion
func (h *Handlers) lintOpenscap(ctx echo.Context, cust *Customizations, fixup bool, distro Distributions, policy string, blueprint *db.BlueprintEntry) ([]BlueprintLintItem, error) {
```

I would go even further and use `bpe`, `bp` or just `b` - this is a practice found in Effective Go and review guide: https://go.dev/wiki/CodeReviewComments#variable-names

internal/v1/handler_oscap.go
```diff
@@ -205,126 +207,548 @@ func (h *Handlers) lintOpenscap(ctx echo.Context, cust *Customizations, fixup bo
...
-			if fixup {
-				cust.Packages = common.ToPtr(append(common.FromPtr(cust.Packages), pkg))
+	// Get saved policy blueprint from service snapshots
+	savedPolicyBp, err := h.getSavedPolicyBlueprint(blueprintEntry, policy)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check for policy additions (items added to current policy vs saved policy)
+	lintErrors = append(lintErrors, h.checkPolicyAdditions(currentPolicyBp, savedPolicyBp, cust, fixup)...)
```
lzap, line 226: It is more straitforward to simply pass a pointer to slice so the method can add it.

internal/v1/handler_oscap.go
```diff
@@ -205,126 +207,548 @@ func (h *Handlers) lintOpenscap(ctx echo.Context, cust *Customizations, fixup bo
...
+
+	// Check for policy removals (items removed from current policy vs saved policy)
+	lintErrors = append(lintErrors, h.checkPolicyRemovals(currentPolicyBp, savedPolicyBp, cust, fixup)...)
+
+	return lintErrors, nil
+}
+
+// checkPolicyAdditions checks for items that were added to current policy vs saved policy
+func (h *Handlers) checkPolicyAdditions(currentPolicyBp *blueprint.Blueprint, savedPolicyBp *blueprint.Blueprint, cust *Customizations, fixup bool) []BlueprintLintItem {
```
lzap, line 235: Okay it is really hard to review. Can you please describe, in your own words, what this method is supposed to do? Also the other one.

The original was a very straightforward solution and should have been refactored first, this is not very readable after the changes applied.

lzap (changes requested):
I don't understand too much. Maybe @croissanne can put some light on this.

internal/v1/handler_oscap.go
```diff
@@ -205,126 +207,548 @@ func (h *Handlers) lintOpenscap(ctx echo.Context, cust *Customizations, fixup bo
...
+	}
+
+	if len(serviceSnapshots.Compliance.PolicyCustomizations) == 0 {
+		return nil, nil
+	}
+
+	var savedCustomizations Customizations
+	if err := json.Unmarshal(serviceSnapshots.Compliance.PolicyCustomizations, &savedCustomizations); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
mgold1234: I changed the implementation, so it is not necussary to do all of that

internal/v1/handler_blueprints.go
```diff
@@ -405,13 +405,28 @@ func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID, params
...
+				var snapshots db.ServiceSnapshots
+				if err := json.Unmarshal(blueprintEntry.ServiceSnapshots, &snapshots); err == nil && snapshots.Compliance != nil {
+					if len(snapshots.Compliance.PolicyCustomizations) != 0 {
+						var savedCust Customizations
+						if err := json.Unmarshal(snapshots.Compliance.PolicyCustomizations, &savedCust); err == nil {
+							savedPolicy = &savedCust
+						}
+					}
+				}
+			}
```
avitova, line 428: This makes the code very complex and less readable. In general, having lots of "levels" of code and lots of inner conditions and loops are not a good practice. Someone else might know if we have some kind of "happy path" concept, but if not, you can also try to avoid such big nesting by doing early returns or creating new helper functions.

internal/v1/handler_blueprints.go
```diff
@@ -405,13 +405,28 @@ func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID, params
...
 }
 
-func (h *Handlers) lintBlueprint(ctx echo.Context, blueprint *BlueprintBody, fixup bool) ([]BlueprintLintItem, error) {
+func (h *Handlers) lintBlueprint(ctx echo.Context, blueprintEntry *db.BlueprintEntry, fixup bool) ([]BlueprintLintItem, error) {
 	lintErrors := []BlueprintLintItem{}
-	if blueprint.Customizations.Openscap != nil {
+	bpBody, err := BlueprintFromEntry(blueprintEntry)
+	if err != nil {
+		return nil, err
+	}
```
avitova, line 413: 2/2 And here, we are calling it again. I kinda do not like this, BlueprintFromEntry calls json.Unmarshal. I think I would even prefer passing both BlueprintEntry and BlueprintBody to the function rather than calling it twice, but that's maybe a subject to more discussion. I think we might aim to figure this out so that we do not have to do this in the first place.

mgold1234: @avitova You're absolutely right about the duplicate unmarshalling being inefficient. I appreciate your suggestion to pass blueprintEntry.ServiceSnapshots separately.
After considering both approaches, I decided to align with @lzap's suggestion to pass only `blueprintEntry` for these reasons:

1. Passing a single parameter that contains all needed data follows Go conventions and keeps the function signature simple.
2. while the duplicate unmarshalling does add overhead, the cost is relatively small compared to the compliance service calls. The cleaner API makes the code easier to understand and maintain long-term.

internal/v1/handler_blueprints.go
```diff
@@ -841,9 +879,15 @@ func (h *Handlers) FixupBlueprint(ctx echo.Context, id openapi_types.UUID) error
...
 	}
 
-	_, err = h.lintBlueprint(ctx, &blueprint, true)
-	if err != nil {
-		return err
+	if blueprint.Customizations.Openscap != nil {
+		compl, err := blueprint.Customizations.Openscap.AsOpenSCAPCompliance()
+		if err == nil && compl.PolicyId != uuid.Nil {
+			snapshotCustomizations := h.extractSnapshotCustomizations(blueprintEntry)
+			_, _, err = h.lintOpenscap(ctx, &blueprint.Customizations, true, blueprint.Distribution, compl.PolicyId.String(), snapshotCustomizations)
```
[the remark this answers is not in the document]
mgold1234: @croissanne the reason `FixupBlueprint` calls `lintOpenscap` directly with fixup rather than using `lintBlueprint` is to ensure the fixups are applied in-place on the same blueprint instance that will be persisted to the DB.

lintBlueprint is primarily an orchestration layer for read-only linting, aggregating results from multiple lint functions. It operates on temporary copies of the blueprint, so using it for fixups would require copying results back, increasing risk of state drift between memory and the persisted object.

By calling lintOpenscap directly:

We make the intent explicit: only Compliance/OpenSCAP fixups are applied.
Fixups happen on the exact Customizations instance that will be saved, ensuring idempotency and avoiding intermediate copies.
It preserves clear separation of responsibilities: lintOpenscap performs side-effects when fixup==true, while lintBlueprint aggregates linting feedback without mutating state.

internal/v1/handler_oscap.go
```diff
@@ -181,150 +182,348 @@ func (h *Handlers) GetOscapCustomizations(ctx echo.Context, distribution Distrib
...
+	if err != nil {
+		return nil, nil, echo.NewHTTPError(http.StatusInternalServerError, err)
+	}
+
+	var allErrors []BlueprintLintItem
+	var allWarnings []BlueprintLintItem
+
+	// Collect errors and warnings from all lint functions (they now return instead of mutating)
+	if errs, warns := lintPackages(policyBP, prevCust, bpBody, fixup); len(errs) > 0 || len(warns) > 0 {
```
croissanne, line 259: Are the len() checks needed here?

internal/v1/handler_oscap.go
```diff
@@ -181,150 +182,348 @@ func (h *Handlers) GetOscapCustomizations(ctx echo.Context, distribution Distrib
...
+				currentCust.Packages = common.ToPtr(append(common.FromPtr(currentCust.Packages), pkg))
 			}
 		}
 	}
 
-	// some policies (ansi minimal) only require some extra packages
-	if bp.Customizations == nil {
-		return lintErrors, nil
+	// Warnings: packages from saved policy no longer required
+	if prevCust != nil && prevCust.Packages != nil {
```
croissanne, line 303: I think this can be simplified further
```
if prevCust != nil {
for _, pkg := range common.FromPtr(prevCust.Packages) {
 …
}
}
```

`common.FromPtr` will return the default value in case the pointer is nil. So if prevCust.Packages is nil, it'll return an empty slice. I think quite a few things below here can be cimplified that way.

internal/v1/handler_blueprints.go
```diff
@@ -398,33 +400,63 @@ func (h *Handlers) GetBlueprint(ctx echo.Context, id openapi_types.UUID, params
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var allErrors []BlueprintLintItem
+	var allWarnings []BlueprintLintItem
+	snapshotCustomizations := extractSnapshotCustomizations(bpe)
+
+	if errors, warnings, err := h.lintOpenscap(ctx, &bpBody.Customizations, fixup, bpBody.Distribution, snapshotCustomizations); err != nil {
+		if err == compliance.ErrorTailoringNotFound {
```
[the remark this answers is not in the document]
croissanne: resolved in the 2nd commit, didn't see that sorry

internal/v1/handler_oscap.go
```diff
@@ -181,150 +182,393 @@ func (h *Handlers) GetOscapCustomizations(ctx echo.Context, distribution Distrib
...
+	}
+	return ""
+}
+
+func lintKernel(policyBP *blueprint.Blueprint, snapshotCust, currentCust *Customizations, fixup bool) ([]BlueprintLintItem, []BlueprintLintItem) {
+	errs := []BlueprintLintItem{}
+	warns := []BlueprintLintItem{}
+	var policyKernel *blueprint.KernelCustomization
+
+	if policyBP != nil && policyBP.Customizations != nil {
```
croissanne: The policyBP nil checks are inconsistent across the helpers.

internal/v1/handler_oscap.go
```diff
@@ -181,150 +182,393 @@ func (h *Handlers) GetOscapCustomizations(ctx echo.Context, distribution Distrib
...
+	if policyKernel != nil {
+		currentName := getKernelName(currentCust)
+		currentAppend := getKernelAppend(currentCust)
+
+		// Check required kernel name
+		if policyKernel.Name != "" && currentName != policyKernel.Name {
+			if !fixup {
+				errs = append(errs, newComplianceError("kernel name %s required by policy not set", policyKernel.Name))
+			} else {
+				ensureKernel(currentCust).Name = &policyKernel.Name
```
croissanne: Again I think the helper isn't adding much here, this works fine
```
				cust.Kernel = common.ToPtr(common.FromPtr(cust.Kernel))
				cust.Kernel.Name = common.ToPtr(kernel.Name)
```

internal/v1/handler_oscap.go
```diff
@@ -181,150 +182,439 @@ func (h *Handlers) GetOscapCustomizations(ctx echo.Context, distribution Distrib
...
+	warns = lintKernelW(policyBP, snapshotCust)
+	allWarnings = append(allWarnings, warns...)
+	errs = lintFIPSE(policyBP, bpBody, fixup)
+	allErrors = append(allErrors, errs...)
+	warns = lintFIPSW(policyBP, snapshotCust)
+	allWarnings = append(allWarnings, warns...)
+
+	// If we couldn't find the policy and fixup was requested, return error after processing warnings
+	if policyBP == nil && fixup {
+		return nil, nil, echo.NewHTTPError(http.StatusNotFound, compliance.ErrorTailoringNotFound)
```
ondrejbudai: This shouldn't be 404.

internal/v1/handler_oscap.go
```diff
@@ -181,150 +182,439 @@ func (h *Handlers) GetOscapCustomizations(ctx echo.Context, distribution Distrib
...
+
+	_, checkWarns = checkObsoleteServices(snapshotCust.Services.Disabled, "disabled", func(service string) bool {
+		return policyServices != nil && slices.Contains(policyServices.Disabled, service)
+	})
+	warns = append(warns, checkWarns...)
+
+	return warns
+}
+
+func checkObsoleteServices(services *[]string, serviceType string, isStillRequired func(string) bool) ([]BlueprintLintItem, []BlueprintLintItem) {
```
ondrejbudai: the first return val is useless.

internal/v1/handler_oscap.go
```diff
@@ -181,150 +182,439 @@ func (h *Handlers) GetOscapCustomizations(ctx echo.Context, distribution Distrib
...
+
+	if policyBP == nil {
+		return errs
+	}
+
+	// Only check if policy actually defines filesystem requirements
+	if policyBP.Customizations != nil {
+		for _, fsc := range policyBP.Customizations.GetFilesystems() {
+			if currentCust == nil || currentCust.Filesystem == nil || !slices.ContainsFunc(*currentCust.Filesystem, func(fs Filesystem) bool { return fs.Mountpoint == fsc.Mountpoint }) {
+				if !fixup {
```
croissanne: Why don't we always add the error as a blueprint lint item? Not sure why this is conditional on fixup.

internal/v1/handler_oscap.go
croissanne: Welll if the policy got deleted maybe? Though it would have failed before this point in that case.

internal/v1/handler_oscap.go
```diff
@@ -181,150 +182,406 @@ func (h *Handlers) GetOscapCustomizations(ctx echo.Context, distribution Distrib
...
-	bp, err := h.server.complianceClient.PolicyCustomizations(ctx.Request().Context(), major, minor, policy)
+
+	var allErrors []BlueprintLintItem
+	var allWarnings []BlueprintLintItem
+
+	policyBP, err := h.server.complianceClient.PolicyCustomizations(ctx.Request().Context(), major, minor, compl.PolicyId.String())
 	if err == compliance.ErrorTailoringNotFound {
-		return nil, err
+		if fixup {
+			return nil, nil, echo.NewHTTPError(http.StatusNotFound, err)
```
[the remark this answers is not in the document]
ondrejbudai: ugh... what shall we do in such a case? I wonder if the linting should just fail with an hard error returned.

internal/v1/handler_oscap.go
```diff
@@ -180,8 +181,25 @@ func (h *Handlers) GetOscapCustomizations(ctx echo.Context, distribution Distrib
...
+	}
+
+	// create empty customizations with only openscap because
+	// we need to store only that in the db.ServiceSnapshot that
+	// we return (and that gets mutated by fixup=true)
+	cust := Customizations{
+		Openscap: &openscap,
+	}
+	withFixup := true
+	_, _, err = h.lintAndFixupOpenscap(ctx, &cust, withFixup, distro, nil)
```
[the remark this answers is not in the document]
mgold1234: @croissanne @mvo5 I decided to add this change in that pr :)

lzap:
I saw a force push before this was put on the queue, what were the changes? Can you perhaps paste the diff from your git reflog? I understand some reviews are painfully slow, however, we should not be merging code that was not reviewed.

PR #1686, opened 2025-09-10
internal: exclude EPEL repos from blueprint export

xbhouse:
This adjusts the logic in the export API to exclude EPEL repos from being exported. 

Jira: [HMS-9214](https://issues.redhat.com/browse/HMS-9214)

### DIFF
--- a/internal/v1/handler_blueprints.go
+++ b/internal/v1/handler_blueprints.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 	"slices"
 	"strconv"
+	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


croissanne (commented):
I'm not sure why they shouldn't be included. Could you update the commit message to include why?

xbhouse:
@croissanne sure! as far as i understand, bulk export is only for exporting "custom" repositories. since we're moving to shared epel repos it wouldn't make sense to import these across orgs since they should already be available.

@jlsherrill can you confirm this is correct? :smiley:

PR #1687, opened 2025-09-10
db: only match on org id when deleting blueprints (HMS-9329)

croissanne:
There are certain scenarios where the account number isn't set in the identity header. For instance when using service accounts and working via the api, the account number will be empty. Deleting the same blueprint via the UI will fail because there the account number will be set.


kingsleyzissou (approved):
Thanks!

lzap (approved):
Wasn't AN deprecated years ago? Shall we straight-up remove it from the DB? (As a follow-up)?

PR #1691, opened 2025-09-18
floorplan: Dump clones table (HMS-9078)

ochosi:
In order to assess how much the `AWS clone to region` feature is used, let's take a look at the numbers.
(This feature was a pre-requisite for the - now decommissioned - 'Launch in AWS' feature. It's not useless now, but also just a convenience function.)


lzap (approved):
Consider aggregating the data right in the query.

kingsleyzissou (approved):

PR #1692, opened 2025-09-19
internal/v1: remove azure source id (HMS-9185)

avitova:
Deleting things after Launch EOF. :100: TIL that the only use of the Sources integration we had was to resolve the tenant id and subscription id. That means nothing else needs to be done, we do not have this anywhere in db.

After decommissioning of Launch, Sources will only support AWS integration with Image Builder. Therefore, we have no use of the azure source id on the backend side.


lzap (approved):
Perfect. I am assuming the UI is ready and this can be merged?

There are two more calls of the `GetUploadInfo` in the codebase until we can remove provisioning client completely.

Is a braking change technically, do we care @croissanne ?

PR #1697, opened 2025-09-26
internal/v1: do not secretly enable rhcd service (HMS-8756)

lucasgarfield:
The service rhcd is no longer in RHEL 10 and enabling it causes RHEL 10 builds with a security profile selected to fail.

I considered adding a conditional to only enable it for RHEL < 10, but instead landed on removing this behavior entirely.

It feels like an anti-pattern to have image-builder-crc "inject" an enabled service into a compose request to me.

Things on my mind that led me to this conclusion:
What if a user explicitly disables or masks `rhcd`? Do we also enable `rhcd` for requests that cockpit-image-builder makes? Shouldn't this kind of knowledge live in osbuild/images so there is only one source of truth?

After this, we will need to determine why RHEL 8 and 9 images don't register via rhc when using some (all?) security profiles.

### DIFF
--- a/internal/v1/handler_compose_image.go
+++ b/internal/v1/handler_compose_image.go
@@ -1168,21 +1168,6 @@ func (h *Handlers) buildCustomizations(ctx echo.Context, cr *ComposeRequest, d *
 		}
 	}
 
-	// we need to explicitly add 'rhcd' to the enabled services
-	// if openscap and subscription customizations are set, otherwise
-	// the insights-client doesn't register properly
-	if cust.Subscription != nil && cust.Subscription.Insights && cust.Openscap != nil {
-		if res.Services == nil {
-			res.Services = &composer.Services{}
-		}
-		if res.Services.Enabled == nil {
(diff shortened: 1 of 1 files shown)


lzap:
Wait, won't this break RHEL 9 and older? Shall we wrap it in an if-statement instead?

kingsleyzissou:
> Wait, won't this break RHEL 9 and older? Shall we wrap it in an if-statement instead?

It won't break those builds, the images just won't automatically be registered.

lucasgarfield:
> @lucasgarfield While I agree about this being an antipattern, how about we do the uncontroversial thing first and amend the conditional to allow RHEL10 builds to pass without affecting RHEL8 and 9?
> 
> It seems to me that this PR will anyway not determine why/how `rhc connect` doesn't work without `rhcd` in these specific cases. I'd rather get to the bottom of this with the OpenSCAP team since we'll also need to figure it out for RHEL 10. WDYT?

Yes, I slept on this and came to the same conclusion myself. Was happy to come here and this. I'll update the PR.

PR #1700, opened 2025-10-01
deps: update to osbuild/logging 0.0.8

lzap:
Although `gobump` would pick this up, I want to do this in an isolated commit so I can go and verify stage as some refactoring was done. It should be fine, many more tests were added to improve coverage in the process.


kingsleyzissou (approved):

lzap:
@croissanne I checked the stage, logging seems to be working fine, but in case something is wrong revert the change.

image

i noticed one small issue but I will file a small PR, not important

PR #1706, opened 2025-10-06
Add pxe image type

lucasgarfield:


croissanne:
> I’ve never needed to update `/distributions` before. I’ve added the new image type to the latest RHEL 9 and 10 releases manually as there doesn’t appear to be a way to programmatically generate these files… is that correct, @croissanne @ondrejbudai? As a followup question, _could_ we programmatically generate these? If not, why?

It's correct. We maybe could by grabbing all the info from images. Not all image types are exposed here on purpose though.

> 
> I wasn’t sure whether to add the new type to the nightlys, either. Looking at wsl as an example, it appears in `rhel-9.6.json` but NOT in `rhel-9-nightly.json`. How do we know whether an image type should be available in the nightlys or not?

The nightlies are purely for our own ends (ci images). We only use aws, gcp and qcows there.

> 
> It might be nice if we add a README to the `/distributions` directory so other newcomers in the future know how to work with these. I’d be happy to put a PR together for that if someone can explain everything to me.

sure

PR #1771, opened 2026-03-03
internal: add support for enabling additional RH repos on first boot (HMS-8857)

xbhouse:
- Looks up any non-base RH repo labels from content-sources and adds these to the subscription options, which are used to enable these repos after registration
- Additional non-base RH repos are requested in the custom and payload customizations, but we need these to be sent in the image request with the base RH repos, so if any are found in the customizations this moves them to the image request 
  - This is a big ugly change -- any input on how it could be improved or done differently is appreciated!

Follow up to https://github.com/osbuild/osbuild-composer/pull/4981 

Jira: https://issues.redhat.com/browse/HMS-8857

### DIFF
--- a/internal/v1/handler_compose_image.go
+++ b/internal/v1/handler_compose_image.go
@@ -94,7 +94,9 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
 		if err != nil {
 			return ComposeResponse{}, err
 		}
-		customizations, err = h.buildCustomizations(ctx, &composeRequest, d)
+
+		var extraRHRepos []content_sources.ApiRepositoryResponse
+		customizations, extraRHRepos, err = h.buildCustomizations(ctx, &composeRequest, d)
 		if err != nil {
 			ctx.Logger().Errorf("Failed building customizations: %v", err)
 			if _, ok := err.(*echo.HTTPError); ok {
@@ -104,7 +106,7 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
 		}
 
 		var detectedOsVersion *string
-		repositories, detectedOsVersion, err = h.buildRepositories(ctx, composeRequest, arch)
+		repositories, detectedOsVersion, err = h.buildRepositories(ctx, composeRequest, arch, extraRHRepos)
 		if err != nil {
 			return ComposeResponse{}, err
 		}
@@ -217,7 +219,7 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
 	}, nil
 }
 
-func (h *Handlers) buildRepositories(ctx echo.Context, composeRequest ComposeRequest, arch *distribution.Architecture) ([]composer.Repository, *string, error) {
+func (h *Handlers) buildRepositories(ctx echo.Context, composeRequest ComposeRequest, arch *distribution.Architecture, extraRHRepos []content_sources.ApiRepositoryResponse) ([]composer.Repository, *string, error) {
 	var repositories []composer.Repository
 	var detectedOsVersion *string
 	var err error
@@ -248,6 +250,13 @@ func (h *Handlers) buildRepositories(ctx echo.Context, composeRequest ComposeReq
 			}
 		}
 
+		// Add extra RH repos from the customizations to the URLs we send to buildRepositorySnapshots
+		for _, r := range extraRHRepos {
+			if r.Url != nil {
+				repoURLs = append(repoURLs, *r.Url)
+			}
+		}
+
 		snapshotRepos, _, snapshotOsVersion, err := h.buildRepositorySnapshots(ctx, repoURLs, nil, false, *composeRequest.ImageRequests[0].SnapshotDate)
 		if err != nil {
 			return nil, nil, err
@@ -262,6 +271,7 @@ func (h *Handlers) buildRepositories(ctx echo.Context, composeRequest ComposeReq
 				expected++
 			}
 		}
+		expected += len(extraRHRepos)
 		if len(repositories) != expected {
 			return nil, nil, fmt.Errorf("no snapshots found for all repositories (found %d, expected %d)", len(repositories), expected)
 		}
@@ -273,7 +283,7 @@ func (h *Handlers) buildRepositories(ctx echo.Context, composeRequest ComposeReq
 		}
 		detectedOsVersion = templateOsVersion
 	} else {
-		repositories, err = h.buildRepositoriesWithLatestSnapshots(ctx, arch, composeRequest.ImageRequests[0].ImageType)
+		repositories, err = h.buildRepositoriesWithLatestSnapshots(ctx, arch, composeRequest.ImageRequests[0].ImageType, extraRHRepos)
 		if err != nil {
 			return nil, nil, err
 		}
@@ -283,7 +293,7 @@ func (h *Handlers) buildRepositories(ctx echo.Context, composeRequest ComposeReq
 
 // buildRepositoriesWithLatestSnapshots transforms the CDN repository URLs to the latest snapshot URL
 // or falls back to CDN if snapshot is not available
-func (h *Handlers) buildRepositoriesWithLatestSnapshots(ctx echo.Context, arch *distribution.Architecture, imageType ImageTypes) ([]composer.Repository, error) {
+func (h *Handlers) buildRepositoriesWithLatestSnapshots(ctx echo.Context, arch *distribution.Architecture, imageType ImageTypes, extraRHRepos []content_sources.ApiRepositoryResponse) ([]composer.Repository, error) {
 	var repositories []composer.Repository
 	var cdnRepoURLs []string
 	var errs []error
@@ -299,7 +309,7 @@ func (h *Handlers) buildRepositoriesWithLatestSnapshots(ctx echo.Context, arch *
 		var err error
 		repoMap, err = h.server.csClient.GetRepositories(ctx.Request().Context(), cdnRepoURLs, nil, false)
 		if err != nil {
-			return nil, fmt.Errorf("unable to retrieve get repositories: %w", err)
+			return nil, fmt.Errorf("unable to retrieve Red Hat repositories: %w", err)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


internal/v1/handler_compose_image.go
```diff
@@ -943,6 +969,26 @@ func (h *Handlers) buildCustomizations(ctx echo.Context, cr *ComposeRequest, d *
...
+		if cust.CustomRepositories != nil {
+			var repoIDs []string
+
+			for _, repo := range *cust.CustomRepositories {
+				if repo.Id != "" {
+					repoIDs = append(repoIDs, repo.Id)
+				}
+			}
+
+			if len(repoIDs) > 0 {
```
lzap: This is already checked in the function and it should work without this check but keep it, extra safety is not bad.

lzap (dismissed):
Refreshing OpenAPI client in a separate commit is generally a good idea (someone else can pull it into their PR if needed) but this will do.

I really cannot tell when this is gonna be green again, maybe we can start merging PRs in CRC without green tests @croissanne ?

jlsherrill:
@xbhouse the reason is that subscription-manager actually manages these within the redhat.repo file, which is why DO want to run  subscription-manager repos --enable REPO_LABEL   on them after registration

internal/v1/handler_compose_image.go
```diff
@@ -130,6 +138,9 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
 				expected++
 			}
 		}
+		for range extraRHRepos {
+			expected++
+		}
```
lzap: I wish I knew how this happened :-)

```go
expected += len(extraRHRepos)
```

internal/clients/composer/openapi.v2.yml
```diff
@@ -2538,7 +2538,7 @@ components:
             type: string
             example: 'content-label-1'
           description: |
-            Optional list of repository IDs to enable with subscription-manager on first boot.
+            Optional list of repository IDs to enable with subscription-manager when registering the system with Insights.
```
lzap: This file is being updated by bot regularly from the upstream, your change will get overrwritten in a week:

https://github.com/osbuild/osbuild-composer/blob/main/internal/cloudapi/v2/openapi.v2.yml#L2543

If you want to fix this comment, do it there.

internal/v1/handler_compose_image.go
```diff
@@ -526,6 +552,137 @@ func (h *Handlers) buildPayloadRepositories(ctx echo.Context, payloadRepos []Rep
...
+				if strings.Contains(*repo.Label, "baseos") || strings.Contains(*repo.Label, "appstream") {
+					continue
+				}
+				labels = append(labels, *repo.Label)
+			}
+		}
+	}
+
+	return labels, nil
+}
```
lzap: Some of these handler methods do actually belong to `client.go`. Not only that, they form a good candidates for methods on the `content_sources.RepositoryByID` (or rather `content_sources.ApiRepositoryResponse` - see my comments about maps). This will allow you to write lower-level unit tests that will be better suited for these methods as well.

internal/v1/handler_compose_image.go
```diff
@@ -526,6 +552,137 @@ func (h *Handlers) buildPayloadRepositories(ctx echo.Context, payloadRepos []Rep
...
+	rhReposByURL := map[string]content_sources.ApiRepositoryResponse{}
+
+	for _, rhRepo := range rhRepoMap {
+		if rhRepo.Url != nil {
+			rhReposByURL[*rhRepo.Url] = rhRepo
+		}
+	}
+
+	return rhReposByURL
+}
```
lzap: I have to say that for several dozens of entries, maps are generally less efficient and IMHO less readable than making a simple lookup function that will iterate an unsorted slice. With `slices` packages, this can be done very nicely:

```go
import "slices"

func FindRepoByURL(repos []content_sources.ApiRepositoryResponse, targetURL string) (content_sources.ApiRepositoryResponse, bool) {
    for repo := range slices.Values(repos) {
	idx := slices.IndexFunc(repos, func(r content_sources.ApiRepositoryResponse) bool {
		return r.Url != nil && *r.Url == targetURL
	})

	if idx == -1 {
		return content_sources.ApiRepositoryResponse{}, false
	}
	return repos[idx], true
}
```

Since we just merged Go 1.24 upgrade today, it could have even be one-liner if we put this into our utils package:

```go
func First[T any](seq iter.Seq[T], check func(T) bool) (v T, ok bool) {
	for v = range seq {
		if check(v) { return v, true }
	}
	return v, false
}
```

The main problem with this is that repository is a big struct with many fields, are you going to build a map builder for each and every field that you need to lookup against? There already is `GetRepositories` in `client.go` which returns repos by UUID, now you are adding one that returns a map by URL. I think if the API function just returned slice of these repos as a plain slice and provided helper functions to lookup them via simple iteration, it would be cleaner and easily extensible.

Of course I am assuming CS returns just few dozens of repos, not hundreds or more. This is nitpicking feel free to ignore if you prefer building maps. At least let's put everything into `client.go` so it can be refactored later to plain slices.

internal/v1/handler_compose_image.go
```diff
@@ -526,6 +552,137 @@ func (h *Handlers) buildPayloadRepositories(ctx echo.Context, payloadRepos []Rep
...
+		if customRepository.Id != "" {
+			if _, isRedHat := rhRepoMap[customRepository.Id]; !isRedHat {
+				filteredCustomRepos = append(filteredCustomRepos, customRepository)
+				filteredCustomIDs = append(filteredCustomIDs, customRepository.Id)
+			}
+		}
+	}
+
+	return filteredCustomRepos, filteredCustomURLs, filteredCustomIDs
+}
```
lzap, line 621: Not all methods are good candidates for `client.go` tho, these `filterXXX` ones require types from other package and the API client should not depend on anything.

internal/v1/handler_compose_image.go
```diff
@@ -526,6 +552,137 @@ func (h *Handlers) buildPayloadRepositories(ctx echo.Context, payloadRepos []Rep
...
+		}
+	}
+	rhRepoMap, err := h.server.csClient.GetRepositories(ctx.Request().Context(), repoURLs, repoIDs, false)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("unable to retrieve RH repositories: %v", err)
+	}
+
+	return repoURLs, repoIDs, rhRepoMap, nil
+}
+
```
lzap, line 683: The same, these should stay here.

Maybe the repository processing could be extracted and moved somewhere else? Perhaps something for a follow up, not in this PR I guess. But this handler is already huge.

lzap (changes requested):
Blocker: change of an external API file.

Recommendation: move some functions into the CS client.

Nitpick: I think maps are not actually needed for processing repositories (by uuid, by url) and if this was simply just a slice with search helpers it would have been cleaner (and also faster as a bonus).

internal/v1/handler_compose_image.go
xbhouse: oh man, i'm going to blame this on PTO brain :sweat_smile:

internal/v1/handler_compose_image.go
xbhouse: i agree, i'm hesitant to change `GetRepositories` though in this PR since that's used in a lot of different places. i removed the repo URL map helper and added a couple helpers to client.go that use slices instead. i started seeing transient test failures when using maps too (related to ordering). good call :+1:

internal/v1/handler_compose_image.go
xbhouse: thanks, moved this to client.go

internal/v1/handler_compose_image.go
xbhouse: this could be separated from the handler, maybe to `/internal/common/`? i'm happy to change it however you all prefer

lzap: Anywhere, just not into `common` please! :-) Packages called `utils` or `common` should have been banned by the compiler IMHO :-)

Feel free to do a follow-up if you want I do not want to hold you back.

lzap (dismissed):

lzap:
Looks great. Sorry for the late review, anyone else?

internal/v1/handler_compose_image.go
```diff
@@ -548,6 +570,117 @@ func (h *Handlers) buildPayloadRepositories(ctx echo.Context, payloadRepos []Rep
...
+	for _, repo := range customRepos {
+		if repo.Id != "" {
+			if rhRepo, ok := rhRepoMap[repo.Id]; ok {
+				rhRepos = append(rhRepos, rhRepo)
+			}
+		}
+	}
+
+	return repoURLs, repoIDs, rhRepos, nil
+}
```
achilleas-k, line 688: I might be missing something here but I find this function a bit confusing.  Correct me if I'm wrong but the way I read it, it does the following:

1. Collects repo URLs and repo IDs from the list of custom repositories.  IDs are only collected if a repo doesn't have a base URL.
2. Passes both URLs and IDs to `GetRepositories()` to get the full information for all the repos (collected in `rhRepoMap`).
3. Scans through the list of custom repositories and collects an repo from `customRepos` that has a repo ID that's in the `rhRepoMap`.

My confusion:
Isn't `rhRepos` always identical to `repoIDs`?  It seems to me that `repoIDs` is the list of repositories found in `customRepos` that have an ID set.  And then we iterate over `customRepos` again and for every repo that has an ID (`repo.Id != ""`) we check if it exists in `rhRepoMap`.  Which is equivalent to checking if any ID in `repoIDs` exists in `rhRepoMap`.  But since `rhRepoMap` is constructed from combining `repoURLs` and `repoIDs`, wont it always have all `repoIDs` in it?  Or does `GetRepositories()` do filtering for RH repos?

I'm also curious why some repositories are identified by base URL, and why base URL is preferred over ID, but that's probably a wider question that depends on how repositories are defined in different places.

achilleas-k (commented):
The fact that this is all one commit made it a bit hard to review and get a good sense of everything that's going on.  I think this generally looks okay, but I do have one question below.  It seems like an obvious thing but I'm sure I'm missing something.

internal/v1/handler_compose_image.go
xbhouse: thanks for the review @achilleas-k! 

yep, when `GetRepositories` is called with `external=false`, this fetches only RH repos, so repo IDs that point to any non-RH repos won't be returned. the second loop checks if the repo ID exists in the `RepositoryByID` map returned from `GetRepositories` and builds the `ApiRepositoryResponse` objects we need to add to the image request repositories.

on using both repo URL and ID: both are checked here because upload repos don't have a URL and are only identifiable by an ID. the API also allows defining a repo by base URL only (no ID), which is what a lot of the compose tests use. happy to switch to preferring ID over URL when both are present if we want that?

let me know if any of that doesn't make sense :)

xbhouse: i'll also split this into multiple commits for easier review, sorry about that :sweat_smile:

lzap (approved):

achilleas-k (approved):

internal/v1/handler_compose_image.go
achilleas-k: Thanks for the explanation and the split

xbhouse:
thanks all!! @achilleas-k or @lzap could you add this to the merge queue when you get a chance? or maybe i need to rebase?

PR #1789, opened 2026-04-08
Check content-sources for snapshot os version

jlsherrill:
Fixes HMS-8659

When using snapshots or content templates, check the snapshots for a detected_os_version.  If one is found, use that to determine the image builder distribution to use. 

Note that this was largely built with cursor using the bug report as the main prompt, but i have reviewed the code and the test.  If there's a way to test this locally, let me know, otherwise it will be easy to test in stage.

[HMS-8659]: https://redhat.atlassian.net/browse/HMS-8659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/v1/handler_compose_image.go
+++ b/internal/v1/handler_compose_image.go
@@ -95,6 +95,7 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
 	}
 
 	var repositories []composer.Repository
+	var detectedOsVersion *string
 	// We should only build repository snapshots in the case where there is a snapshot date and no content template
 	if composeRequest.ImageRequests[0].SnapshotDate != nil && composeRequest.ImageRequests[0].ContentTemplate == nil {
 		repoURLs := []string{}
@@ -117,11 +118,12 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
 			}
 		}
 
-		snapshotRepos, _, err := h.buildRepositorySnapshots(ctx, repoURLs, nil, false, *composeRequest.ImageRequests[0].SnapshotDate)
+		snapshotRepos, _, snapshotOsVersion, err := h.buildRepositorySnapshots(ctx, repoURLs, nil, false, *composeRequest.ImageRequests[0].SnapshotDate)
 		if err != nil {
 			return ComposeResponse{}, err
 		}
 		repositories = append(repositories, snapshotRepos...)
+		detectedOsVersion = snapshotOsVersion
 
 		// A sanity check to make sure there's a snapshot for each repo
 		expected := 0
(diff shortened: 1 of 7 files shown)


internal/v1/handler_compose_image.go
```diff
@@ -161,6 +166,27 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
...
+				if det := h.server.distroRegistry(ctx).FindByMajorMinor(major, minor); det != nil {
+					if det.Distribution.ComposerName != nil {
+						distro = *det.Distribution.ComposerName
+					} else {
+						distro = det.Distribution.Name
+					}
+				}
+			}
+		}
+	}
```
lzap, line 175: Could you create `FindByMajorMinorStr(majorMinor string)` helper method in distro registry so more code can be moved from the handler? We need to at least try to keep the handlers tidy.

internal/v1/mocks/content_sources.go
```diff
@@ -377,77 +379,91 @@ func templateByID(uuid string) (res content_sources.ApiTemplateResponse) {
 }
 
 func ContentSources(w http.ResponseWriter, r *http.Request) {
-	if tutils.AuthString0 != r.Header.Get("x-rh-identity") {
-		w.WriteHeader(http.StatusInternalServerError)
-		return
-	}
+	contentSourcesHandler(nil)(w, r)
+}
```
lzap, line 383: This refactoring would be nice to have in a separate commit for easier review, but what is done is done.

internal/v1/handler_compose_image.go
```diff
@@ -444,14 +471,19 @@ func (h *Handlers) buildRepositorySnapshots(ctx echo.Context, repoURLs []string,
 		customRepo.ModuleHotfixes = repo.ModuleHotfixes
 		customRepo.CheckRepoGpg = repo.MetadataVerification
 		customRepositories = append(customRepositories, customRepo)
+
+		// Capture the first detected OS version seen across all snapshots.
+		if detectedOsVersion == nil && snap.Match.DetectedOsVersion != nil {
+			detectedOsVersion = snap.Match.DetectedOsVersion
+		}
```
lzap, line 470: Wait will this work for a snapshot chain that started as 9.6 and carried over to 9.7 where incompatibility occurs? I am bit confused, just resolve if you know this is allright.

lzap (changes requested):
One question and one small request to move more code out of the handler.

internal/v1/mocks/content_sources.go
jlsherrill: if you hide whitespace changes, it makes it easier to review

internal/v1/handler_compose_image.go
jlsherrill: yes! content-sources scans each snapshot to see the version of redhat-release and records which os version was detected.  So based on the date (or whats in the template), it will check the snapshots and get the recorded os version.  This way if you build with 9.6 snapshots, it will use the 9.6 distro config.  If you build with 9.7, it'll use the 9.7 distro config.  Let me konw if that doesn't make sense!

internal/v1/handler_compose_image.go
```diff
@@ -340,14 +366,14 @@ func (h *Handlers) buildComposerRepositoryFromSnapshot(snapshotPath string, useP
 	return composerRepo, nil
 }
 
-func (h *Handlers) buildRepositorySnapshots(ctx echo.Context, repoURLs []string, repoIDs []string, external bool, snapshotDate string) ([]composer.Repository, []composer.CustomRepository, error) {
+func (h *Handlers) buildRepositorySnapshots(ctx echo.Context, repoURLs []string, repoIDs []string, external bool, snapshotDate string) ([]composer.Repository, []composer.CustomRepository, *string, error) {
```
avitova, line 359: Should we maybe put a comment somewhere before the function, suggesting what the return value of *string is?

internal/v1/handler_compose_image.go
jlsherrill: updated, and added a test for the new function

internal/v1/handler_compose_image.go
jlsherrill: added!

avitova (approved):
I am good here.

jlsherrill:
@lzap i believe i addressed all comments

lzap (approved):
Apologies for missing this one, all good of course. Thanks!

PR #1792, opened 2026-04-14
Handle empty string for detectedOsVersion

jlsherrill:
The previous change https://github.com/osbuild/image-builder-crc/pull/1789 anticipated that content sources would return nil for detectedOSVersion, but it actually returns empty string.  This can cause the detection not to work, as it would be set to ""  which would cause it to default to rhel-9/rhel-10


avitova (approved):

lzap (approved):

jlsherrill:
it worked!   (i think) now to solve pqc issues on the content side

PR #1793, opened 2026-04-15
many: support bootc compose requests (HMS-10356)

lzap:
We rushed in an implementation based on "image ID", these turned useless since definitions are now part of regular (RPM) distros. It can be dropped now, we agreed on the following plan:

* [x] Drop ID from the distro definition, only image type and reference will be used
* [x] Make sure `reference` name is used to align with the composer api
* [x] Document the correct `reference` name and limitations
* [x] Make sure tags are part of the reference (currently only `latest`)
* [x] Lookup function will only use reference and will just validate
* [x] End to end that will query + use the reference

---

_Initially this PR had also some irrelevant optimalizations this was split into: https://github.com/osbuild/image-builder-crc/pull/1795_

https://redhat.atlassian.net/browse/HMS-10356

### DIFF
--- a/distributions/rhel-10.1/rhel-10.1.json
+++ b/distributions/rhel-10.1/rhel-10.1.json
@@ -9,10 +9,10 @@
   },
   "x86_64": {
     "bootc": [
-      { "id": "rhel-10.1-ec2", "type": "ec2", "reference": "rhel/10.1-ec2" },
-      { "id": "rhel-10.1-gcp", "type": "gcp", "reference": "rhel/10.1-gcp" },
-      { "id": "rhel-10.1-azure", "type": "azure", "reference": "rhel/10.1-azure" },
-      { "id": "rhel-10.1-qcow2", "type": "qcow2", "reference": "rhel/10.1-qcow2" }
+      { "type": "ec2", "reference": "rhel/10.1-ec2" },
+      { "type": "gcp", "reference": "rhel/10.1-gcp" },
+      { "type": "azure", "reference": "rhel/10.1-azure" },
+      { "type": "qcow2", "reference": "rhel/10.1-qcow2" }
     ],
     "image_types": ["aws", "azure", "gcp", "guest-image", "image-installer", "network-installer", "oci", "pxe-tar-xz", "vsphere", "vsphere-ova", "wsl"],
     "repositories": [{
--- a/distributions/rhel-10.1/rhel-10.1.json
+++ b/distributions/rhel-10.1/rhel-10.1.json
@@ -9,10 +9,10 @@
   },
   "x86_64": {
     "bootc": [
-      { "id": "rhel-10.1-ec2", "type": "ec2", "image_name": "rhel/10.1-ec2" },
-      { "id": "rhel-10.1-gcp", "type": "gcp", "image_name": "rhel/10.1-gcp" },
-      { "id": "rhel-10.1-azure", "type": "azure", "image_name": "rhel/10.1-azure" },
-      { "id": "rhel-10.1-qcow2", "type": "qcow2", "image_name": "rhel/10.1-qcow2" }
+      { "id": "rhel-10.1-ec2", "type": "ec2", "reference": "rhel/10.1-ec2" },
+      { "id": "rhel-10.1-gcp", "type": "gcp", "reference": "rhel/10.1-gcp" },
+      { "id": "rhel-10.1-azure", "type": "azure", "reference": "rhel/10.1-azure" },
+      { "id": "rhel-10.1-qcow2", "type": "qcow2", "reference": "rhel/10.1-qcow2" }
     ],
     "image_types": ["aws", "azure", "gcp", "guest-image", "image-installer", "network-installer", "oci", "pxe-tar-xz", "vsphere", "vsphere-ova", "wsl"],
     "repositories": [{
--- a/distributions/rhel-10.1/rhel-10.1.json
+++ b/distributions/rhel-10.1/rhel-10.1.json
@@ -9,10 +9,10 @@
   },
   "x86_64": {
     "bootc": [
-      { "type": "ec2", "reference": "rhel/10.1-ec2" },
-      { "type": "gcp", "reference": "rhel/10.1-gcp" },
-      { "type": "azure", "reference": "rhel/10.1-azure" },
-      { "type": "qcow2", "reference": "rhel/10.1-qcow2" }
+      { "type": "aws", "reference": "quay.io/redhat-services-prod/insights-management-tenant/image-builder-bootc-foundry/rhel-10.1-ec2:latest" },
+      { "type": "gcp", "reference": "quay.io/redhat-services-prod/insights-management-tenant/image-builder-bootc-foundry/rhel-10.1-gcp:latest" },
+      { "type": "azure", "reference": "quay.io/redhat-services-prod/insights-management-tenant/image-builder-bootc-foundry/rhel-10.1-azure:latest" },
+      { "type": "guest-image", "reference": "quay.io/redhat-services-prod/insights-management-tenant/image-builder-bootc-foundry/rhel-10.1-qcow2:latest" }
     ],
     "image_types": ["aws", "azure", "gcp", "guest-image", "image-installer", "network-installer", "oci", "pxe-tar-xz", "vsphere", "vsphere-ova", "wsl"],
     "repositories": [{
--- a/internal/distribution/bootc_distributions.go
+++ b/internal/distribution/bootc_distributions.go
@@ -1,7 +1,6 @@
 package distribution
 
 type BootcDistributionEntry struct {
-	ID        string `json:"id"`
 	Distro    string `json:"distro"`
 	Name      string `json:"name"`
 	Type      string `json:"type"`
--- a/internal/distribution/bootc_distributions.go
+++ b/internal/distribution/bootc_distributions.go
@@ -6,5 +6,5 @@ type BootcDistributionEntry struct {
 	Name      string `json:"name"`
 	Type      string `json:"type"`
 	Arch      string `json:"arch"`
-	ImageName string `json:"image_name"`
+	Reference string `json:"reference"`
 }
--- a/internal/distribution/bootc_distributions.go
+++ b/internal/distribution/bootc_distributions.go
@@ -1,9 +1,21 @@
 package distribution
 
+import "fmt"
+
 type BootcDistributionEntry struct {
 	Distro    string `json:"distro"`
 	Name      string `json:"name"`
 	Type      string `json:"type"`
 	Arch      string `json:"arch"`
 	Reference string `json:"reference"`
+}
+
+func (arch Architecture) ValidateBootcReference(reference string) error {
+	for _, image := range arch.Bootc {
+		if image.Reference == reference {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("bootc reference '%s' not found", reference)
 }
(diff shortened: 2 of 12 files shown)


internal/distribution/bootc_lookup.go
```diff
@@ -0,0 +1,24 @@
+package distribution
+
+import (
+	"fmt"
+)
+
+// LookupBootcImage finds a bootc image by ID in the given architecture.
+// Returns the BootcImage if found, or an error if not found.
+func LookupBootcImage(arch *Architecture, bootcID string) (*BootcImage, error) {
```
croissanne: This feels like it needs to be a member function of Architecture. Similar to FindPackages, but thenLookupBootCImage or something.

croissanne (changes requested):
There's a lot of commits here which refactor tests, which don't have that much to do with bootc support. Could you split those out? I think all we need is:
- distributions: update rhel-10.1 bootc image name
- distribution: rename bootc image_name to image_ref
- distribution: add bootc image lookup function
- v1: integrate bootc lookup into compose handler

.github/workflows/tests.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           version: v1.64.6
 
       - name: Run unit tests
-        run: make unit-tests
+        run: make unit-tests-coverage
```
thozza: This change is unrelated to https://github.com/osbuild/image-builder-crc/pull/1793/changes/6854a04f2afc7471fddc91cab3903cf13738e4e9 commit

internal/v1/api.yaml
```diff
@@ -879,9 +879,10 @@ components:
         reference:
           type: string
           description: |
-            Image name from the bootc distributions list. Must match an image_name
-            returned by GET /distributions?kind=bootc.
-          example: 'rhel-10.1-qcow2'
+            Bootc image ID from the bootc distributions list. Must match an 'id'
+            returned by GET /distributions?kind=bootc (e.g., 'rhel-10.1-ec2').
+            This will be resolved to the corresponding container image reference.
+          example: 'rhel-10.1-ec2'
```
thozza: I think that the example for the `image_ref` does not match the description of `BootcBody.reference` property. So will it be the `rhel-10.1-ec2` or a full reference without a tag? 🤔

internal/v1/handler_post_compose_test.go
```diff
@@ -3585,3 +3586,107 @@ func TestComposeWithSnapshotDetectedOsVersion(t *testing.T) {
...
+		capturedRequest = nil
+		payload := createBootcComposePayload("centos-9", "", "x86_64", v1.ImageTypesGuestImage, v1.UploadTypesAwsS3)
+
+		respStatusCode, body := tutils.PostResponseBody(t, srv.URL+"/api/image-builder/v1/compose", payload)
+		require.Equal(t, http.StatusCreated, respStatusCode, body)
+
+		// Verify composer received nil bootc field
+		require.NotNil(t, capturedRequest)
+		require.Nil(t, capturedRequest.Bootc)
+	})
```
thozza: It would be much better to have the cases be defined in a slice of test case structures and have a single common test logic instead of having 4 very similar copies of the test logic.

thozza (changes requested):
I'm with Sanne on the test changes being split out into a separate PR, let's keep this one focused.

However, there is a bigger API architecture mess from my PoV...

The current design has `bootc.reference` in the compose request accept a logical ID (e.g., `rhel-10.1-ec2`) which the server then resolves to the actual container image reference. This creates a few problems:

1. **Confusing naming**: The field is called `reference` but it accepts an `id`, not an image reference. Meanwhile the list endpoint returns both `id` and `image_ref`, making it unclear which one to use.
2. **Forward-compatibility**: If we ever want to support 3rd-party container references, we'd need to rework the API schema since clients are currently trained to send short IDs.
3. **Misleading examples**: The `image_ref` example in `BootcDistributionItem` is `quay.io/.../rhel-10.0-ec2` which looks like an ID suffix, not like the real ref (`quay.io/redhat-services-prod/insights-management-tenant/image-builder-bootc-foundry/rhel-10.1-qcow2`).

### Suggestion

Make `bootc.reference` accept the **full container image reference** (the same string returned as `image_ref` from `GET /distributions?kind=bootc`). The compose handler would then **validate** that the submitted ref is in the allowed set for the selected distro+arch, rather than resolving an ID to a ref.

This way:
- The client flow becomes straightforward: copy `image_ref` from the list response into `bootc.reference` in the compose request.
- The field name `reference` actually means what it says — a container image reference.
- Later, to support 3rd-party container images, we just relax the validation — no API schema change needed.

### Concrete changes

**`api.yaml`** — update `BootcBody.reference` description and example:

```yaml
reference:
  type: string
  description: |
    Container image reference for the bootc base image.
    Must match an 'image_ref' from GET /distributions?kind=bootc
    for the selected distribution and architecture.
  example: 'quay.io/redhat-services-prod/insights-management-tenant/image-builder-bootc-foundry/rhel-10.1-qcow2'
```

Also fix the `BootcDistributionItem.image_ref` example to use a realistic full reference instead of `quay.io/.../rhel-10.0-ec2`.

**`LookupBootcImage`** — match on `ImageRef` instead of `ID`:

```go
func LookupBootcImage(arch *Architecture, imageRef string) (*BootcImage, error) {
    // ...
    for i := range arch.Bootc {
        if arch.Bootc[i].ImageRef == imageRef {
            return &arch.Bootc[i], nil
        }
    }
    return nil, fmt.Errorf("bootc image ref '%s' is not allowed", imageRef)
}
```

(Also consider making this a method on `Architecture` per @croissanne's other comment.)

**Compose handler** — the logic barely changes, but the semantics flip from "resolve" to "validate":

```go
if composeRequest.Bootc != nil {
    validatedBootc, err := distribution.LookupBootcImage(arch, composeRequest.Bootc.Reference)
    if err != nil {
        return ComposeResponse{}, echo.NewHTTPError(http.StatusBadRequest, err.Error())
    }
    bootc = &composer.Bootc{
        Reference: validatedBootc.ImageRef,
    }
}
```

### Note on tags

The current distro JSON refs have no tag (e.g., no `:latest` or `:v1.0`). Worth clarifying whether the API should expect/require tags, and whether the server should pin specific tags in the distro JSON for reproducibility.

### end2end test

I would propose that the end to end unit test should also query the distributions endpoint and use the value returned by it for the compose endpoint, instead of directly calling the compose endpoint with know to be good value. This way we validate the actual client API call flow.

lzap:
I split the tests optimalizations into a separate PR, see the description, I rebased. This is now ready for re-review. Please note the end to end test is missing I plan on doing a follow up, or add it after the first round of reviews.

internal/v1/handler_post_compose_test.go
lzap: You mean tabular test? Sure.

internal/distribution/bootc_distributions.go
```diff
@@ -1,9 +1,31 @@
...
+
+func (df DistributionFile) ValidateBootcReference(reference string) error {
+	if df.ArchX86 != nil {
+		return df.ArchX86.ValidateBootcReference(reference)
+	}
+	if df.Aarch64 != nil {
+		return df.Aarch64.ValidateBootcReference(reference)
+	}
+	return fmt.Errorf("no architecture found in distribution file: %s", df.Distribution.Name)
+}
```
lzap: This one is unused, it is only used on architecture. Can drop it.

distributions/rhel-10.1/rhel-10.1.json
```diff
@@ -9,10 +9,10 @@
...
     "bootc": [
-      { "type": "ec2", "reference": "rhel/10.1-ec2" },
-      { "type": "gcp", "reference": "rhel/10.1-gcp" },
-      { "type": "azure", "reference": "rhel/10.1-azure" },
-      { "type": "qcow2", "reference": "rhel/10.1-qcow2" }
+      { "type": "aws", "reference": "quay.io/redhat-services-prod/insights-management-tenant/image-builder-bootc-foundry/rhel-10.1-ec2:latest" },
+      { "type": "gcp", "reference": "quay.io/redhat-services-prod/insights-management-tenant/image-builder-bootc-foundry/rhel-10.1-gcp:latest" },
+      { "type": "azure", "reference": "quay.io/redhat-services-prod/insights-management-tenant/image-builder-bootc-foundry/rhel-10.1-azure:latest" },
+      { "type": "guest-image", "reference": "quay.io/redhat-services-prod/insights-management-tenant/image-builder-bootc-foundry/rhel-10.1-qcow2:latest" }
     ],
```
lzap, line 16: The type change belongs to the previous commit.

croissanne (dismissed):

kingsleyzissou (dismissed):

internal/distribution/bootc_distributions.go
```diff
@@ -1,9 +1,31 @@
...
+
+func (df DistributionFile) ValidateBootcReference(reference string) error {
+	if df.ArchX86 != nil {
+		return df.ArchX86.ValidateBootcReference(reference)
+	}
+	if df.Aarch64 != nil {
+		return df.Aarch64.ValidateBootcReference(reference)
+	}
+	return fmt.Errorf("no architecture found in distribution file: %s", df.Distribution.Name)
```
thozza: This is wrong, the check always checks only the first non-nil architecture. So if the distro config has both, `x86_64` and `aarch64` and one provides `aarch64` reference, it will be considered invalid, because it won't be found in the `x86_64` config.

also, the error returned when not found in any architecture should be really
```go
return fmt.Errorf("bootc reference '%s' not found in distribution: %s", reference, df.Distribution.Name)
```

Please, also extend the test coverage to catch the case that would incorrectly fail now due to the bug.

Thinking about this, why is this method even needed? It does not seem to be used anywhere and the validation can be done correctly only if you know also the architecture that the reference is going to be used for. Doing just YOLO search to find a match in any architecture (in this case really only in the first non-nil arch) makes no sense.

internal/distribution/bootc_distributions_test.go
```diff
@@ -6,6 +6,71 @@ import (
...
+		df := DistributionFile{
+			Distribution: DistributionItem{Name: "test-distro"},
+			ArchX86:      &Architecture{Bootc: bootcEC2},
+			Aarch64:      &Architecture{Bootc: []BootcImage{{Type: "arm-only", Reference: "r2"}}},
+		}
+		require.NoError(t, df.ValidateBootcReference("r1"))
+		err := df.ValidateBootcReference("r2")
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "bootc reference 'r2' not found")
+	})
```
thozza: This makes NO sense. This test case tests basically for a buggy behavior and makes it the desired state. Getting `r2` should still work.

thozza (changes requested):
Thanks, this looks much better.

The `func (df DistributionFile) ValidateBootcReference(reference string) error` method is incorrect and it is not used anywhere. Let's just remove it. The validation as part of the header is done by using the `Architecture.ValidateBootcReference()` method, which does what I would expect to happen. So it should work fine.

I won't block on this, but I hate seeing the separate subtests in each unit test function. Ideally, the cases would be defined in a table and then a single test logic implementation would run them. The same goes for the compose handler unit tests - the two test functions should be really a single parametrized test with three scenarios.

lzap:
Thanks all.

* Dropped the unused (and buggy) function.
* Added a new commit that converts both new and already existing tests to tabular-style.

croissanne (approved):

thozza (approved):
Thanks. LGTM.

I have to note that `internal/v1/handler_post_compose_bootc_test.go` newly added tests could be also table-driven, but let's get this PR in as is...

lzap:
> I have to note that `internal/v1/handler_post_compose_bootc_test.go` newly added tests could be also table-driven, but let's get this PR in as is...

Thanks I have big plans with CRC tests, once we merge the performance uplift, I want to break the monster test into individual files. I can throw some table driven test refactoring too.

PR #1795, opened 2026-04-16
many: speed up unit/db tests from

lzap:
Several changes that speeds up unit tests significantly. Some patches are easy wins, others are debatable but I really like to have fast tests that just helps me to run them more often when changing things.

Major contributors:

* Race detector (can be now turned off locally via `UNIT_TEST_ARGS` ENV variable)
* Coverage (ditto)
* Creating new database via `psql` inside `podman` (is now native pgx call)
* Reloading distro registry in the same test package over and over again
* Waiting for postgres via iterating each second (this is now 250ms, also fixed an infinite loop bug)
* Optimizations to postgres config (it already runs in `tmpfs` so minor improvement)


internal/tutils/psql.go
```diff
@@ -95,18 +112,69 @@ func (p *PSQLContainer) execCommand(args ...string) (string, error) {
...
 
-func (p *PSQLContainer) execQuery(dbase, cmd string) (string, error) {
-	args := []string{
-		"exec", p.name, "psql", "-U", user,
+func (p *PSQLContainer) pgConnString(database string) string {
+	host := net.JoinHostPort("127.0.0.1", strconv.Itoa(p.port))
+	return fmt.Sprintf("postgres://%s@%s/%s?sslmode=disable", user, host, database)
+}
+
+func (p *PSQLContainer) adminPGConn(ctx context.Context) (*pgx.Conn, error) {
```
croissanne, line 120: something here is leaking connections

```

tern command error: exit status 1, output: Unable to connect to PostgreSQL:
        	            	  failed to connect to `host=localhost user=postgres database=test2faec46a6d4149cfb88bbc0e0a9f207b`: server error (FATAL: sorry, too many clients already (SQLSTATE 53300))
```

internal/tutils/psql.go
```diff
@@ -17,13 +21,23 @@ type PSQLContainer struct {
...
 const (
 	image string = "quay.io/osbuild/postgres:13-alpine"
 	user  string = "postgres"
 )
 
+var fastPgArgs = []string{
+	"postgres",
+	"-c", "fsync=off",
+	"-c", "full_page_writes=off",
+	"-c", "synchronous_commit=off",
```
croissanne, line 38: The amount of data we write / deal with is so small that these options don't actually have any effect on the speed of the test. The DB related speedup just comes from not having to start a process / calling podman etc.

internal/tutils/psql.go
lzap: I do not see this how do I reproduce this?

internal/tutils/psql.go
lzap: Yes I know, this was a left over from my experiments I thought it will not hurt, do you want me to remove it?

lzap:
Fixed the leaking connections, this was not caused by this PR it was there before as well. The pool was never closed after tests were finished.

Also noted some warnings in the logs which were caused by the container not using IPv6 which is now the default for `localhost` on modern systems.

lzap:
Also appended two more commits that enables `t.Parallel()` for all non-db tests (the first one) and then also for the db tests (the second one). The latter needed some concurrency changes in the psql.go but I tested everything 50 times with concurrency of 32.

The total test run on my system is now 3.3s :-)

internal/tutils/psql.go
lzap: Do you want me to remove this? Getting conflicts so only if you insist.

lzap:
@croissanne deps are broken due to OpenAPI, I had to add an extra commit that configures gobump to run unit tests after every single bump. since tests are now fast this is very nice and also it should improve gobump success rate with its PRs: `gobump -exec ./tools/prepare-source.sh -exec "make unit-tests"`

I just re-executed tests 100x times to ensure this is stable enough, I know timing is not great but in the worst case we can turn off `t.Parallel()` everywhere if we bump into issues, I have been rebasing this for quite some time.

internal/distribution/distroregistry_test.go
```diff
@@ -27,8 +30,7 @@ func TestDistroRegistry_List(t *testing.T) {
 		"with-bootc",
 	}
 
-	dr, err := LoadDistroRegistry("./testdata/distributions")
-	require.NoError(t, err)
+	dr := distroReg
```
croissanne: Why copy it to a local variable at here? Might as well use the global one now, no?

internal/distribution/distroregistry_test.go
```diff
@@ -44,8 +46,7 @@ func TestDistroRegistry_List(t *testing.T) {
 }
 
 func TestDistroRegistry_Get(t *testing.T) {
-	dr, err := LoadDistroRegistry("./testdata/distributions")
-	require.NoError(t, err)
+	dr := distroReg
```
croissanne: ditto

internal/tutils/psql.go
croissanne: Why not, if it doesn't add anything might as well not have it?

croissanne (changes requested):
This is a big PR and I'm not sure if all commits are actually helping with the speedup.

I think it might be nice to collect some stats per 'speedup' commit, in a little table or something, to see what is actually worth having  the extra complexity.

The following I think are definitely worth it:
- [tutils: improve postgres container startup](https://github.com/osbuild/image-builder-crc/pull/1795/changes/340d5668085a5db47a96bad9e99e5d041fb50bca)
- [distribution: share distro registry between tests](https://github.com/osbuild/image-builder-crc/pull/1795/changes/36c17dc5627ba54337cd96370c85e4a78fc5d6fb)
- [cicd: fix race conditions in tests](https://github.com/osbuild/image-builder-crc/pull/1795/changes/7b79767b339c9a82894b9701155be23cd62bcc9a)
- [db: close connection pool in tests](https://github.com/osbuild/image-builder-crc/pull/1795/changes/80afd6769a71b52b4f587194f7b39af21d9cf93e)
- [tutils: further speed up unit tests](https://github.com/osbuild/image-builder-crc/pull/1795/changes/1b0e41a891330b13c3ce33f57e3a8f06f3e17ba2) (but needs a better commit message title).
- and the .github workflow changes etc. which make use of the faster options

lzap:
I rebased, resolved conflicts, reordered commits so tests are all passing, dropped the "localhost" commit that was not contributing any value and then reexecuted tests 3 times. In the PR description, I present cumulative number. As you can see, each commit contributes to better performance except the two that are technical things that are relevant to this:

* closing connection pool (this was a bug in `main`, without it, when running tests multiple times it fails)
* reconfigure gobump to run tests during bumps (this is possible now that tests are fast)

.github/workflows/tests.yml
```diff
@@ -47,8 +47,8 @@ jobs:
         with:
           version: v1.64.6
 
-      - name: Run unit tests
-        run: make unit-tests
+      - name: Run unit tests with coverage
+        run: make coverage-dump
```
avitova, line 51: `make unit-tests` still works, but do we want to maybe change this in README to suggest running tests with coverage?
https://github.com/osbuild/image-builder-crc/blob/fa6f93701de60a8978d982312cb05d468090113e/README.md?plain=1#L40

internal/common/pagination_test.go
```diff
@@ -72,7 +73,9 @@ func TestPaginatorBasic(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
+		tc := tc
```
avitova: So I found this interesting and tried to google why you might be doing this.
I think we [might not need](https://go.dev/doc/go1.22#language) this anymore.

> Previously, the variables declared by a “for” loop were created once and updated by each iteration. In Go 1.22, each iteration of the loop creates new variables, to avoid accidental sharing bugs. The [transition support tooling](https://go.dev/wiki/LoopvarExperiment#my-test-fails-with-the-change-how-can-i-debug-it) described in the proposal continues to work in the same way it did in Go 1.21.

internal/tutils/psql.go
```diff
@@ -154,11 +136,23 @@ func (p *PSQLContainer) execQuery(ctx context.Context, dbase, cmd string) (strin
...
+		}
+		if p.pgConn == nil {
+			conn, err := pgx.Connect(ctx, p.pgConnString("postgres"))
+			if err != nil {
+				return "", err
+			}
+			p.pgConn = conn
 		}
-		tag, err := conn.Exec(ctx, cmd)
+		tag, err := p.pgConn.Exec(ctx, cmd)
```
avitova, line 155: Right here, there is no check that `p.pgConn` is not nil, but I assume that after checking that the connection did not end up with an error, we do not need to check that?

lzap: Not sure what you mean here, the nil check is few lines above and then the result from pgx.Connect does not return (nil, nil).

lzap:
Thank you, fixed both problems. Also fixed the embarrassing Schutzbot git authorship (I accidentally set that while I was testing gobump a week before).

internal/tutils/psql.go
avitova: > Not sure what you mean here, the nil check is few lines above and then the result from pgx.Connect does not return (nil, nil).
That's exactly what I was checking, thanks, I should have looked it up:)

avitova:
There is more of the `tc := tc` or `tt := tt`. I personally do not mind, so I think I am good here. You need only one approval, so I will let Sanne judge if she's okay with it now.

lzap:
Dropped it, I missed it, thanks. Rebased. Here is the summary of all these improvements:

Commit | Time (s) | Comment
-- | -- | --
db: close connection pool in tests | 491.33 | A bugfix that only manifests when parallel execution is enabled later on
cicd: allow overriding unit flags locally | 503.36 | Improvement: Allows for running tests without coverage report locally.
tutils: improve postgres container startup | 64.52 | Improvement: This is the biggest leap in this patch series.
tutils: run tern directly instead of through podman | 60.96 | Improvement: Looks like a small improvement, but this is actually a lot when tests are executed in parallel later.
cicd: run tests and prepare-source for each gobump | 59.16 | Since tests are now fast, let's make use of it in gobump.
distribution: share distro registry between tests | 29.17 | Improvement: Excessive loading of distros.
Use parallel tests for some unit tests | 12.36 | Improvement: Codebase is now thread-safe, so we can do this.

@croissanne

ochosi (dismissed):
This looks good to me! ⚡  I have a slight concern about the test behaving differently and returning different results locally and in CI, but I get that the coverage may not something you always want to execute.

lzap:
Added a warning message: `WARNING: To run tests in CICD configuration, run: make unit-tests-coverage`

lzap:
@croissanne did I drop all of the unnecessary commits?

avitova:
@croissanne Any comments?

PR #1798, opened 2026-04-17
handler_blueprints: add Bootc to ComposeRequest

ksiekl:
When a user triggers a compose from an image-mode blueprint, the bootc field was being dropped — the blueprint stored it correctly, but ComposeBlueprint never included it in the ComposeRequest passed to handleCommonCompose. This meant osbuild-composer received no bootc reference and built a non-bootc image.

### DIFF
--- a/internal/v1/api.yaml
+++ b/internal/v1/api.yaml
@@ -1284,7 +1284,6 @@ components:
         - fedora-42
         - fedora-43
         - fedora-44
-        - image-mode
     ImageRequest:
       type: object
       additionalProperties: false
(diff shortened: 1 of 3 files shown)


croissanne (approved):
yup

internal/v1/api.yaml
```diff
@@ -1284,7 +1284,6 @@ components:
         - fedora-42
         - fedora-43
         - fedora-44
-        - image-mode
```
croissanne, line 1287: lovely, thank you :)

lzap (approved):

PR #1799, opened 2026-04-17
v1: drop distribution from compose request if bootc is set

croissanne:
Composer expects either bootc or distribution, but not both. Ideally this would be enforced, and distribution would no longer be required in the API. However, this might have unforseen consequences in the frontend and considering the tight deadline for this, add the hack for now.


lzap (approved):

kingsleyzissou (approved):

lzap:
The followup is: https://github.com/osbuild/image-builder-crc/pull/1802

PR #1802, opened 2026-04-20
v1: make distribution not required (HMS-10557)

croissanne:
With bootc composes being supported, either distribution or the bootc
    reference can be set.
    
    However, in order to avoid breaking frontend, do not enforce the
    exclusive or for now. Once the frontend has been adapted, the mutally
    exclusive logic can be enforced.


internal/distribution/bootc_distributions.go
```diff
@@ -10,7 +10,10 @@ type BootcDistributionEntry struct {
 	Reference string `json:"reference"`
 }
 
-func (arch Architecture) ValidateBootcReference(reference string) error {
+func (arch *Architecture) ValidateBootcReference(reference string) error {
```
lzap: Nitpick: Does not have to be a pointer receiver. In fact, distro registry below isn't.

internal/v1/api.yaml
```diff
@@ -1161,7 +1158,6 @@ components:
       required:
         - name
         - description
-        - distribution
         - customizations
         - metadata
       properties:
```
lzap, line 1163: I actually do not mind keeping it like this, these anyOf make those weird union-like JSONs which are ugly...

lzap (commented):
Nice, let's not merge until we hold THE demo :-)

Kinda do not understand the last commit description and why it is not part of the previous commit.

internal/v1/api.yaml
croissanne: Hm wait wdym here, there's no anyOf

internal/distribution/bootc_distributions.go
croissanne: Needs it for the nil guard

internal/v1/api.yaml
lzap: Ah sorry I just appreciate that you do not use `anyOf` here and trying to say I do not mind it. I just heard you saying on scrum that you plan on doing that.

croissanne: It would be for a followup anyway; and yea  it's easier to just enforce it in code rather than the union jsons.

croissanne:
> Nice, let's not merge until we hold THE demo :-)
> 

Ye we can wait. That said it shouldn't break anything, the actual api from the user's POV hasn't changed, he objects that are actually posted are the same.

> Kinda do not understand the last commit description and why it is not part of the previous commit.

Was mostly for ease of review, can easily squash them if you feel it would be more logical. I mostly separated it to highlight the changes to fixup (short circuiting those handlers / sections for blueprints with bootc).

lzap (approved):
I'd probably use HTTP 4xx for the error but 5xx will work too.

kingsleyzissou (approved):

PR #1807, opened 2026-04-27
unleash: add context support (HMS-10567)

lzap:
This extends our unleash support with context with custom field `osDistro` that is used by the new `image-builder.bootc-distro.disabled` feature flag to filter out (hide) particular bootc distros and/or image types. This can be useful during the insights freeze.

With this change, `orgId` field in Unleash can be used to enable/disable features. This can be useful for gradual releases or kill switch per-organization, it is a bonus feature.

I also added tests, this is kind of hard to test as I do not have credentials for stage and I do not want to deploy my own instance just for this. @croissanne

### DIFF
--- a/internal/unleash/unleash.go
+++ b/internal/unleash/unleash.go
@@ -1,11 +1,15 @@
 package unleash
 
 import (
+	"context"
 	"fmt"
+	"maps"
 	"net/http"
 
 	"github.com/Unleash/unleash-client-go/v4"
 	ucontext "github.com/Unleash/unleash-client-go/v4/context"
+	fedora_identity "github.com/osbuild/community-gateway/oidc-authorizer/pkg/identity"
+	rh_identity "github.com/redhatinsights/identity"
 )
 
 type FeatureFlag string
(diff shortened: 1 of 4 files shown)


internal/unleash/unleash.go
```diff
@@ -37,7 +42,103 @@ func Initialize(conf Config) error {
...
+	if distroName != "" {
+		if !IsEnabledWithFallback(withOsDistro(uc, distroName), string(flag), false) {
+			return true
+		}
+	}
+	if imageType == "" {
+		return false
+	}
+	return !IsEnabledWithFallback(withOsDistro(uc, distroName+"-"+imageType), string(flag), false)
+}
```
thozza, line 74: `EnabledBootcCtx()` is a bit hard to follow because it negates `IsEnabledWithFallback()`, while the flag itself is a `*.disabled flag`. Even if the logic is correct, the double-negative makes the intent difficult to parse. Could we rename or wrap this so the call site reads in domain terms, e.g. `isBootcDisabled(...)` / `isBootcAllowed(...)`, instead of negating a generic `IsEnabled...` helper?

Alternatively, would it be simpler to model this as `BootcDistroEnabled` with fallback `true`? That seems easier to reason about than a `*.disabled` flag evaluated via `IsEnabledWithFallback(...)` and then negated. If changing the flag semantics in Unleash is undesirable, at least the wrapper would make things easier to follow.

thozza (commented):
Thanks for the PR.

Logically, the implementation seems OK to me, but I find it hard to follow. Please see my inline comment.

internal/unleash/unleash.go
lzap: Great idea I haven't realized that the fallback actually supports values, that is indeed much cleaner I did not like this at all.

thozza:
needs rebase

lzap:
Closing this, I mean, would be nice but we do not need it. Less code, less bugs.

PR #1813, opened 2026-04-30
v1: let bootc take priority over package mode

croissanne:
In case both distribution and bootc are set (the current case for requests coming from the frontend), let bootc take priority.

Otherwise the bootc part will not be forwarded, essentially giving users a package mode image.


lzap (approved):
Makes sense.

achilleas-k (approved):
Ah I see.  yes, lgtm

kingsleyzissou (commented):
Cheeky, late review

kingsleyzissou (approved):

PR #1822, opened 2026-05-12
v1: require either bootc or distribution in common compose

croissanne:
`/compose`: the direct compose handler will now require one or the
    other, and never both.
    
    `/blueprints/{id}/compose`: composing images via the blueprint will
    result in the distribution being dropped from the compose request in
    case bootc is set. The ability to keep setting distribution in the
    blueprint itself is kept, to allow the frontend to easily present bootc
    images categorized by distribution.

---

- [x] #1821

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -169,3 +169,23 @@ file to generate a package list.
 
 If the repository requires a client tls key/cert you can supply them with
 `--key` and `--cert`.
+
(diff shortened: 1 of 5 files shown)


kingsleyzissou (approved):

lzap (approved):
Goodness, try again

PR #1829, opened 2026-05-20
v1: fix parsing of compose error statuses (HMS-10693)

croissanne:
In select error cases the compose error is presumed to contain error
    details which are formatted as a list of
    `composer.ComposeStatusError`s. However these details could also be a
    string or just a single, top-level `composer.ComposeStatusError`
    object. This fixes a panic in case the error details are formatted
    differently.

### DIFF
--- a/internal/v1/export_test.go
+++ b/internal/v1/export_test.go
@@ -5,8 +5,9 @@ import (
 )
 
 var (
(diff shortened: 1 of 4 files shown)


kingsleyzissou (approved):

ochosi (approved):
LGTM!

PR #1833, opened 2026-05-27
distributions: remove the google-cloud-sdk repo (HMS-10712)

ondrejbudai:
This is actually the same thing as we already did on-prem, see: https://github.com/osbuild/images/commit/0d56c60e54575c4ff23fb4c53b9b8e7e05f89f53

Keeping it in crc actually bit us - google-cloud-cli-spanner-emulator from this repo (EL10) started bundling a copy of libstdc++. Image Builder started preferring this package as a provider of this library instead of the real libstdc++ coming from EL10 baseos. This broke the images because the bundled one is obviously not available in standard library lookup paths (/usr/lib64/)

By removing this repository, we will fix the images - image-builder will revert back to installing the standard libstdc++ that its deps seek.


croissanne (approved):
thanks!

lzap (approved):

PR #1839, opened 2026-06-16
floorist: remove sensitive aap_registration data (HMS-10828)

lzap:
The feature has been deployed in prod for quite some time and we are not tracking its use yet. Let's fix that.

---

The first commit drops storing of the whole AAP registration JSON, I think we should not collect that. The new solution only tracks for presence of the customization instead. Note: AAP is stored under a different column now that it is a different data type to make it easier during analysis.

Edit: Update I was made aware we are tracking Satellite registration but not AAP registration. I wonder how and why we have the opposite in the SQL. I can probably turn this PR into fixing this (maybe we can just drop AAP from there)?

@jlsherrill @ochosi

### DIFF
--- a/templates/image-builder.yml
+++ b/templates/image-builder.yml
@@ -219,8 +219,7 @@ objects:
                else c.request->'customizations'->'kernel' end as kernel,
           case when c.request->'customizations'->'firewall'='{}' then 'null'
                else c.request->'customizations'->'firewall' end as firewall,
-          jsonb_array_length(c.request->'customizations'->'users') as num_users,
-          c.request->'customizations'->'aap_registration' as aap_registration
+          jsonb_array_length(c.request->'customizations'->'users') as num_users
         from
           composes c
           left outer join blueprint_versions v on c.blueprint_version_id = v.id


templates/image-builder.yml
```diff
@@ -220,7 +220,8 @@ objects:
           case when c.request->'customizations'->'firewall'='{}' then 'null'
                else c.request->'customizations'->'firewall' end as firewall,
           jsonb_array_length(c.request->'customizations'->'users') as num_users,
-          c.request->'customizations'->'aap_registration' as aap_registration
+          (c.request->'customizations'->'aap_registration' IS NOT NULL AND c.request->'customizations'->'aap_registration' != 'null'::jsonb) as registration_aap,
+          (c.request->'customizations'->'subscription' IS NOT NULL AND c.request->'customizations'->'subscription' != 'null'::jsonb) as registration_sat
```
ochosi: How is this different from what we already dump in line 202?

lzap: Line 202 is `  c.request->'customizations'->'subscription' as subscription,` I am not following?

lzap: I will take a look on how we generate the current dashboard, then I will probably close this.

lzap:
Ah this is how our dashboard works: If `/etc/systemd/system/register-satellite.service` exists then the feature is detected as in use. For AAP, we do not do anything. Let me add this to the dashboard. Also, I am dropping the second commit, the AAP customization is not needed for this at all.

lzap:
https://github.com/osbuild/image-builder-crc/pull/1841

PR #1841, opened 2026-06-18
clowdapp: fix aap floorist selection

jlsherrill:
parquet seems to interpet {} as 'false', and thus categorizes the entire filed as boolean.  This appears to be counting everything as 'false' even though data is there

### DIFF
--- a/templates/image-builder.yml
+++ b/templates/image-builder.yml
@@ -220,7 +220,8 @@ objects:
           case when c.request->'customizations'->'firewall'='{}' then 'null'
                else c.request->'customizations'->'firewall' end as firewall,
           jsonb_array_length(c.request->'customizations'->'users') as num_users,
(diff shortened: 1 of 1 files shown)


ochosi (dismissed):

jlsherrill:
updated to just include true/false if its customized, tested in stage:
```
select aap_registration, count(*) from ( select   (c.request->'customizations'->'aap_registration' is not null
   and c.request->'customizations'->'aap_registration' != '{}') as aap_registration from composes c ) group by aap_registration;

 AAP_REGISTRATION | COUNT 
------------------+-------
 false            | 52680 
 true             | 2177
```

ochosi (approved):

achilleas-k (approved):

lzap:
Good catch!

PR #1844, opened 2026-06-25
maintenance: add failed builds Sentry reporting

lzap:
Adds failed-build reporting to the image-builder-maintenance cron job. When DB maintenance runs, it queries composes from the last 24 hours, checks each build’s status via the composer API, and reports failures to Sentry/GlitchTip with context and JSON payloads.

This builds on the infrastructure we already have in place for the web app. There is a simple fallback mechanism for when Sentry rejects the event with attachments, which is what it does today but I talked to AppSRE and they can do this for us if we ask them to.

Workflow

- Query local composes table for builds created in the last 24 hours
- For each compose, call composer GET /composes/{id} to check status
- On failure, build a Sentry event with attachments
- If attachments are rejected, retry with trimmed inline JSON in event Extra fields (this can be dropped once file upload feature is enabled)

---

This is just an idea how we could leverage GlitchTip for this. If there are no objections, I would like to test this on stage, present the results on a demo and then we can decide if this is a good idea or not. If the latter, removing this from the codebase will be easy as it is literally one Go source file and three new configuration options.

### DIFF
--- a/cmd/image-builder-maintenance/failed_builds.go
+++ b/cmd/image-builder-maintenance/failed_builds.go
@@ -1,0 +1,408 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/getsentry/sentry-go"
+	"github.com/google/uuid"
+	"github.com/osbuild/image-builder-crc/internal/clients/composer"
+	"github.com/osbuild/image-builder-crc/internal/oauth2"
+	v1 "github.com/osbuild/image-builder-crc/internal/v1"
+)
+
+const (
+	sentryMaxExtraFieldBytes = 4096
+	truncatedMiddle          = " [... truncated ...] "
+	window                   = 24 * time.Hour
+
+	sqlListRecentComposes = `
+		SELECT job_id, request, org_id, image_name, created_at
+		FROM composes
+		WHERE created_at >= $1 AND deleted = FALSE
+		ORDER BY created_at DESC`
+)
+
+type recentCompose struct {
+	JobID     uuid.UUID
+	Request   json.RawMessage
+	OrgID     string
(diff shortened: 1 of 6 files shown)


croissanne (dismissed):
let's try it

cmd/image-builder-maintenance/failed_builds.go
```diff
@@ -0,0 +1,406 @@
...
+}
+
+func trimForSentryField(ctx context.Context, s string, maxBytes int) string {
+	if len(s) <= maxBytes {
+		return s
+	}
+
+	keep := max(maxBytes-len(truncatedSuffix), 0)
+
+	trimmed := s[:keep] + truncatedSuffix
```
schuellerf: could we change that to have half of the keep-bytes before and the half at the end?
"long text [… truncated …] end of long text"
as often either the beginning or the end contains a hint

lzap: Sure, but I am going to ask for attachments feature immediately if this turns out to be turned off (we do not know yet until we try). Let me fix this.

lzap:
I rebased the requested change.

lzap:
I have reviewed Clawder YAML and `COMPOSER_TOKEN_URL` was missing, the other values are correct. For DSN I will raise a trivial MR to app-interface.

schuellerf (approved):

croissanne (approved):

lzap:
FTR I am adding automated-actions script support to the project, this is documented in `app-interface` internally. To trigger cron job manually:

```
 Trigger a CronJob:
  automated-actions openshift-trigger-cronjob \
    --cluster app-sre-stage-01 \
    --namespace composer-stage \
    --cronjob <cronjob-name>

  Delete a Job:
  automated-actions openshift-workload-delete \
    --cluster app-sre-stage-01 \
    --namespace composer-stage \
    --kind Job \
    --name <job-name>
```

PR #1857, opened 2026-07-27
internal: fix error handling of 401 from content sources (HMS-10083)

matej-lohnicky:
### Context
Content Sources require specific roles for operations during blueprint compose and blueprint export, which can result in a `401 Unauthorized` response. 

Previously, errors from `content_sources/client.go/request` were wrapped in plain `fmt.Errorf`, causing them to lose their specific error status and ultimately surface to the user as generic `500 Internal Server Error`.

### Changes
- adds explicit handling for 401 errors so they are preserved and properly communicated to the user instead of being swallowed.
-  adds a dedicated unit test covering the 401 error path.
- corrects inverted logic when handling `body` on read errors (drive-by fix):
  ```go
  body, err := io.ReadAll(resp.Body)
  if err != nil {
     return nil, fmt.Errorf("Some text: %s", body)
  }

### DIFF
--- a/internal/clients/content_sources/client.go
+++ b/internal/clients/content_sources/client.go
@@ -158,7 +158,7 @@ func (csc *ContentSourcesClient) FindRepoByID(repos []ApiRepositoryResponse, tar
 func (csc *ContentSourcesClient) FetchLabelsFromNonBaseRHRepos(ctx context.Context, repoIDs []string) ([]string, error) {
 	rhRepoMap, err := csc.GetRepositories(ctx, nil, repoIDs, false)
 	if err != nil {
-		return nil, fmt.Errorf("unable to retrieve RH repositories: %v", err)
+		return nil, fmt.Errorf("unable to retrieve RH repositories: %w", err)
 	}
 
 	var labels []string
--- a/internal/clients/content_sources/client.go
+++ b/internal/clients/content_sources/client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -14,6 +15,8 @@ import (
 	"github.com/osbuild/logging/pkg/strc"
 	"github.com/redhatinsights/identity"
 )
+
+var ErrorAuth = errors.New("user is not authorized - please check your 'Repositories viewer' or 'Content Template viewer' permissions")
 
 type RepositoryByID map[string]ApiRepositoryResponse
 
@@ -83,13 +86,14 @@ func (csc *ContentSourcesClient) fetchRepositories(ctx context.Context, repoURLs
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		if resp.StatusCode != http.StatusUnauthorized {
-			body, err := io.ReadAll(resp.Body)
-			if err != nil {
-				return nil, fmt.Errorf("unable to fetch repositories from %s, got %v response, body: %s", csReposURL.String(), resp.StatusCode, body)
-			}
+		if resp.StatusCode == http.StatusUnauthorized {
+			return nil, ErrorAuth
 		}
-		return nil, fmt.Errorf("unable to fetch repositories from %s, got %v response", csReposURL.String(), resp.StatusCode)
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, fmt.Errorf("unable to fetch repositories from %s, got %v response", csReposURL.String(), resp.StatusCode)
+		}
+		return nil, fmt.Errorf("unable to fetch repositories from %s, got %v response, body: %s", csReposURL.String(), resp.StatusCode, body)
 	}
 
 	var repos *ApiRepositoryCollectionResponse
@@ -223,13 +227,14 @@ func (csc *ContentSourcesClient) GetTemplateByID(ctx context.Context, uuid strin
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		if resp.StatusCode != http.StatusUnauthorized {
-			body, err := io.ReadAll(resp.Body)
-			if err != nil {
-				return nil, fmt.Errorf("unable to fetch template, got %v response, body: %s", resp.StatusCode, body)
-			}
+		if resp.StatusCode == http.StatusUnauthorized {
+			return nil, ErrorAuth
 		}
-		return nil, fmt.Errorf("unable to fetch template, got %v response", resp.StatusCode)
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, fmt.Errorf("unable to fetch template, got %v response", resp.StatusCode)
+		}
+		return nil, fmt.Errorf("unable to fetch template, got %v response, body: %s", resp.StatusCode, body)
 	}
 
 	var template *ApiTemplateResponse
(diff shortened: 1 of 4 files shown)


internal/clients/content_sources/client.go
```diff
@@ -15,6 +16,10 @@ import (
 	"github.com/redhatinsights/identity"
 )
 
+var (
+	ErrorAuth = errors.New("user is not authorized for content sources")
```
schuellerf: I was wondering if we could use the _exact_ wording of the RBAC permission needed, so users (and agents) can map this better? Is this possible, @croissanne ?

internal/v1/handler_compose_image.go
```diff
@@ -492,13 +492,14 @@ func (h *Handlers) buildRepositorySnapshots(ctx echo.Context, repoURLs []string,
...
 
 	if snapResp.StatusCode != http.StatusOK {
-		if snapResp.StatusCode != http.StatusUnauthorized {
-			body, err := io.ReadAll(snapResp.Body)
-			if err != nil {
-				return nil, nil, nil, err
-			}
-			ctx.Logger().Warnf("Unable to resolve snapshots: %s", body)
+		if snapResp.StatusCode == http.StatusUnauthorized {
+			return nil, nil, nil, content_sources.ErrorAuth
```
croissanne, line 505: Nitpick, but to make the whole client a bit more consistent maybe it would be nice to return content_sources.ErrorAuth rather than callers having to compare the status code. This would mean changing the functions in the content sources client a bit, but I don't think the function signatures would change.

Not a blocker just wondering if it would make it nicer / more consistent. It's what we do for compliance as well. We don't do it for composer because anything unexpected there should be a 500 anyways.

internal/clients/content_sources/client.go
```diff
@@ -223,13 +229,14 @@ func (csc *ContentSourcesClient) GetTemplateByID(ctx context.Context, uuid strin
...
 
 	if resp.StatusCode != http.StatusOK {
-		if resp.StatusCode != http.StatusUnauthorized {
-			body, err := io.ReadAll(resp.Body)
-			if err != nil {
-				return nil, fmt.Errorf("Unable to fetch template, got %v response, body: %s", resp.StatusCode, body)
-			}
+		if resp.StatusCode == http.StatusUnauthorized {
+			return nil, ErrorAuth
+		}
```
croissanne, line 232: Basically we could do this for all functions in the client.

croissanne (dismissed):
Very nice, thank you!

Some thoughts about consistency& commit messages. 

The only other client that interacts with another insights service is the compliance client, might make sense to make any client that interacts with insights services as similar as possible. Meaning always returning the sentinel error in case of 401, that way users of the client don't need to do different handling depending on the function called (and just always do an errors.Is check).

One note about commit messages though, in general we don't use the `fix()` `feature()` etc. thing, Take a look at https://osbuild.org/docs/developer-guide/general/workflow/#preferred-commit-message-format. so in this case it would be something like `clients/content_sources: return sentinel error in case...` and `v1: check for sentinel error...` etc.

internal/clients/content_sources/client.go
croissanne: I think the user needs repositories read access or something, lemme check

croissanne: Ok so at the very least `Repositories viewer` and/or `content template viewer`. It's worth it to refer to those roles directly in the error ig, makes it easier.

lzap: Conventions in Go are:

* Use `Error` suffix for error types.
* Use `Err` prefix for error values.

This should have been `ErrAuthentication` because 401 errors address authentication issues, and 403 errors point towards authorization problems. Just "auth" could mean anything.

matej-lohnicky: Hmm, I basically took this from the compliance client, which has `ErrorAuth = errors.New("user is not authorized")` as I tried to stick with the style.

internal/v1/handler_compose_image.go
```diff
@@ -103,7 +103,9 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
 		customizations, _, err = h.buildCustomizations(ctx, &composeRequest, nil)
 		if err != nil {
 			ctx.Logger().Errorf("Failed building customizations: %v", err)
-			if _, ok := err.(*echo.HTTPError); ok {
+			if errors.Is(err, content_sources.ErrorAuth) {
+				return ComposeResponse{}, echo.NewHTTPError(http.StatusForbidden, err)
+			} else if _, ok := err.(*echo.HTTPError); ok {
 				return ComposeResponse{}, err
```
avitova, line 110:
```suggestion
			if errors.Is(err, content_sources.ErrorAuth) {
				return ComposeResponse{}, echo.NewHTTPError(http.StatusForbidden, err)
			}
			if _, ok := err.(*echo.HTTPError); ok {
return ComposeResponse{}, err
```
No need for "else", the previous "if" branch has a return.

internal/v1/handler_compose_image.go
```diff
@@ -133,7 +135,9 @@ func (h *Handlers) handleCommonCompose(ctx echo.Context, composeRequest ComposeR
 		customizations, extraRHRepos, err = h.buildCustomizations(ctx, &composeRequest, d)
 		if err != nil {
 			ctx.Logger().Errorf("Failed building customizations: %v", err)
-			if _, ok := err.(*echo.HTTPError); ok {
+			if errors.Is(err, content_sources.ErrorAuth) {
+				return ComposeResponse{}, echo.NewHTTPError(http.StatusForbidden, err)
+			} else if _, ok := err.(*echo.HTTPError); ok {
```
avitova:
```suggestion
			if errors.Is(err, content_sources.ErrorAuth) {
				return ComposeResponse{}, echo.NewHTTPError(http.StatusForbidden, err)
			}
			if _, ok := err.(*echo.HTTPError); ok {
```

avitova (commented):
Good job! Just a small nitpick that I did not see anyone else commenting on.

internal/clients/content_sources/client.go
croissanne: Let's maybe keep it as `ErrorAuth` for now, and then in a followup we could switch them all over? Consistency within a codebase is more important imo.

matej-lohnicky: The whole error message the user gets is now `unable to retrieve repositories: user is not authorized - please check your 'Repositories viewer' or 'Content Template viewer' permissions`  (the prefix is preexisting)

internal/v1/handler_compose_image.go
matej-lohnicky: Skipping due to significantly higher complexity than the existing PR. (partial fix or signature changes)

internal/clients/content_sources/client.go
```diff
@@ -15,6 +16,10 @@ import (
 	"github.com/redhatinsights/identity"
 )
 
+var (
+	ErrorAuth = errors.New("user is not authorized - please check your 'Repositories viewer' or 'Content Template viewer' permissions")
+)
```
avitova:
```suggestion
var ErrorAuth = errors.New("user is not authorized - please check your 'Repositories viewer' or 'Content Template viewer' permissions")
```
We do not need a grouped declaration, since this is only one variable.

internal/v1/handler_compose_image.go
avitova: @croissanne Since this is out of scope, I suggest we create a ticket for this, so that we do not forget, and maybe some day we fix this? But not in this specific PR.

avitova (dismissed):

avitova:
Sorry, we have to wait before #1871 gets admin merged, and then you will have to rebase in order for the Trigger GitLab CI to pass.

avitova (approved):
Looks good! Thank you for your patience with rebases and solving conflicts.

croissanne (approved):
thanks!

PR #1860, opened 2026-07-29
chore: upgrade to Go 1.26 and openapi deps

lzap:
I was expecting build failures but it was straightforward. Gobump would pick this up next week but since I have the branch...

Edit: Okay was not that straightforward as I thought but it is done. Also UBI9 is now on Go 1.26 so we can do this now.


go.mod
```diff
@@ -1,19 +1,19 @@
 module github.com/osbuild/image-builder-crc
 
-go 1.24.12
+go 1.25.0
```
lzap: Wait I missed this.

avitova (approved):
LGTM! Thanks!

ondrejbudai (approved):

PR #1887, opened 2026-08-31
image-builder-maintenance: single target migration (HMS-11188)

ksiekl:
This PR:

- Adds a maintenance job to split multi-target blueprints into one blueprint per image target
- Naming is `{name} - {image_type}`, if that is taken, `{name} - {image_type}-{arch}`
- After split, the old multi-target blueprint is soft-deleted
- Composes are moved to the matching new blueprint (by image type + arch), not deleted
- Version history is not copied
- Runs in one DB transaction; each blueprint has its own savepoint so one failure doesn’t stop the rest
- If a blueprint can’t be split (e.g. no free name), it’s skipped and left unchanged
- Controlled by `ENABLE_BLUEPRINT_SPLIT` (off by default), works with `DRY_RUN`
- Deleted blueprints and latest versions that are already single-target are skipped


cmd/image-builder-maintenance/split.go
```diff
@@ -0,0 +1,411 @@
...
+			created++
+			continue
+		}
+
+		body, err := bodyWithSingleImageRequest(src.Body, req)
+		if err != nil {
+			return created, skipped, fmt.Errorf("blueprint %s: build body: %w", src.ID, err)
+		}
+
+		for {
```
ondrejbudai: Please extract this to a helper, or at least make a comment, this is a tad long to read and understand.

internal/db/db.go
```diff
@@ -127,7 +127,7 @@ const (
         `
 )
 
-func InitDBConnectionPool(ctx context.Context, connStr string) (DB, error) {
+func InitDBConnectionPool(ctx context.Context, connStr string) (*dB, error) {
```
croissanne, line 130: Instead of changing the type here, could we just add the functions to the `DB` interface?

cmd/image-builder-maintenance/split.go
```diff
@@ -0,0 +1,415 @@
...
+			"new_name", name,
+			"image_type", meta.ImageType,
+			"architecture", meta.Architecture)
+		return name, nil
+	}
+}
+
+// logSkip records why one image request was not turned into a new blueprint.
+func logSkip(ctx context.Context, src multiTargetBlueprint, meta imageRequestMeta, reason string) {
+	slog.InfoContext(ctx, "skipping image request",
```
croissanne, line 270: this feels more than just an info statement. Essentially both of the cases here (name collision and already split) are error cases. Just to make sure we pay attention to these.

croissanne: Also, nitpick, not entirely sure this needed to be a separate function here, since the caller has a bunch of other log statements in its body.

cmd/image-builder-maintenance/split.go
```diff
@@ -0,0 +1,415 @@
...
+// one image_requests entry and creates one new blueprint per target. After a
+// source is fully split, it is soft-deleted.
+func SplitMultiTargetBlueprints(ctx context.Context, dbURL string, dryRun bool) error {
+	d, err := db.InitDBConnectionPool(ctx, dbURL)
+	if err != nil {
+		return err
+	}
+	defer d.Pool.Close()
+
+	tx, err := d.Pool.Begin(ctx)
```
croissanne, line 71: Would it make sense to have one transaction per blueprint? Rather than batching them by 500?

Like query the batch in the main tx; and then run all the operations for 1 blueprint (including the name choices) in a child tx?

cmd/image-builder-maintenance/split.go
```diff
@@ -0,0 +1,415 @@
...
+	slog.InfoContext(ctx, "soft-deleted original blueprint after split",
+		"source_id", src.ID,
+		"org_id", src.OrgID,
+		"source_name", src.Name)
+	return created, skipped, true, nil
+}
+
+// splitOneRequest creates a single-target child for one image request, or
+// proposes a name in dry-run. claimed tracks names already used in this split.
+func splitOneRequest(ctx context.Context, tx pgx.Tx, src multiTargetBlueprint, req json.RawMessage, meta imageRequestMeta, dryRun bool, claimed map[string]struct{}) (string, error) {
```
croissanne, line 218: This name is confusing imo, `splitOneRequest` doesn't actually split anything; it maps 1 image request to 1 new blueprint. `mapOneRequest` would be better imo.

internal/db/db.go
ksiekl: Oh yeah, sure!

cmd/image-builder-maintenance/split.go
ksiekl: Good point, thanks. I made it an Error

cmd/image-builder-maintenance/split.go
ksiekl: I've added nested transactions, so that they create savepoints in case some bp has a problem splitting. 
So now it's:
One outer transaction wraps the entire thing. Inside it, each blueprint gets its own savepoint. If it cannot be split and the original bp soft-deleted, the bp is skipped. Inside the splitting blueprint function, each insert can get a nested savepoint for name-collision retries.

cmd/image-builder-maintenance/split.go
```diff
@@ -0,0 +1,410 @@
...
+	if err != nil {
+		return err
+	}
+	defer d.Close()
+
+	tx, err := d.Begin(ctx)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = tx.Rollback(ctx) }()
```
ondrejbudai: Please log the error here. If it happened, we would be blind.

I'm a big believer in handling all errors, even if just logging them, because anytime an error is ignored, we never see it.

achilleas-k (dismissed):
LGTM but I'm not as familiar with the this side of things as the others so take it for what it is: A surface level approval of the general idea.

ksiekl:
Thank you @achilleas-k for the review, I thought of one more thing so I added one more commit, sorry!

brlane-rht:
I too think this looks ok, despite not being familiar with this codebase. I had a mild concern about the sorting by UUID missing a newly inserted blueprint, but since it starts the batch gathering with UUID.Nil it would get picked up on the next run.

ondrejbudai (approved):

kingsleyzissou (approved):
Thanks for looking into this!

PR #1892, opened 2026-09-17
v1: restricting the API to allow only one target per request (HMS-11323)

ksiekl:
The upcoming change of having blueprints with only one target env means that API has to be restricted to allow only one item in the array.

I suppose the *proper* change would be to remove the array completely and keep only the target object, but that would break frontend, and change the type for database so I went with this approach.


kingsleyzissou (approved):

avitova (approved):
Nice!

